PR #1, opened 2023-02-20
Added some typing throughout api

Lash-L:
I am trying to get the custom component into a state that would be approved by Home Assistant as a core integration. Part of that process includes adding typing inside the component, it becomes a bit easier if the underlying api has typing as well. there were some parts I wasn't sure about such as the queue items, and what all data types can be returned from the api, so they are just "Any" for  now.

As far as preparing the integration to get it to a good enough state for Home Assistant, I'd love to talk to you about that if you're interested. I've gotten a decent bit of work done. and with your help, we could definitely get it up for review in a week or two.

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -16,6 +16,7 @@ import threading
 import time
 from asyncio import Lock
 from asyncio.exceptions import TimeoutError, CancelledError
+from typing import Any
 from urllib.parse import urlparse
 
 import aiohttp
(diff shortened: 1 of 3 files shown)


humbertogontijo (approved):
The changes in api.py are welcome, but I don't think it was needed since it would auto type.
The other files are looking good.
Thanks for the help
You can email me so we can get in touch

PR #2, opened 2023-02-21
Bump pycryptodome to 3.17

Lash-L:
version 3.16 causes a version discrepancy as version 3.17 is required by another integration in HA core.


Lash-L:
Hey @humbertogontijo this is the one of the PRs that is completely needed. The checks for a review won't pass with this dependency error. Other than that, I can pretty much submit a home assistant review whenever. I sent you an email if you have any concerns! 

No real rush, just wanted to let you know!

PR #3, opened 2023-02-22
Moved code mappings to api

Lash-L:
Getting an integration approved in HA is complicated. The larger the PR, the less likely it will be to be reviewed at any time soon. 

So I moved the code mappings to the api to decrease the file size of vacuum.py.

We don't have to do this if you are against this. 

This is a breaking change for the hacs integration. unless I make the same changes over there.

Let me know your thoughts.

### DIFF
--- a/roborock/containers.py
+++ b/roborock/containers.py
@@ -1,5 +1,8 @@
 from enum import Enum
 
+from roborock.code_mappings import STATE_CODE_TO_STATUS, ERROR_CODE_TO_TEXT, FAN_SPEED_CODES, MOP_MODE_CODES, \
+    MOP_INTENSITY_CODES
+
 
 class UserDataRRiotReferenceField(str, Enum):
     REGION = "r"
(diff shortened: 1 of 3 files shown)


roborock/containers.py
```diff
@@ -629,6 +632,10 @@ def msg_ver(self) -> int:
     def msg_seq(self) -> int:
         return self.get(StatusField.MSG_SEQ)
 
+    @property
+    def status(self) -> str:
+        return STATE_CODE_TO_STATUS.get(self.state, f"Unknown Status {self.state}")
```
humbertogontijo: I think this should be handled in another way. Returning a free text instead of the status itself looks bad. There are already attribute like state, error_code and the new ones you added handling the raw value

humbertogontijo (changes requested):

roborock/containers.py
Lash-L: What would you prefer? returning the state_code like before and just handling the get on the frontend like before? Perfectly fine with whatever you would like to do, just let me know.

humbertogontijo: Yes. Just returning str mapping or None as before, and let whoever is using the data deal with it

Lash-L: Okay, changed!

PR #5, opened 2023-02-24
Change to timeout from wait_for

Lash-L:
wait_for creates a task, async_timeout does the same work and avoids the task creation

This is one of the current optimizations they are making on HA core.

Sorry to be making so many PRS!

### DIFF
--- a/roborock/roborock_queue.py
+++ b/roborock/roborock_queue.py
@@ -18,5 +18,4 @@ class RoborockQueue(Queue):
 
     async def async_get(self, timeout: float | int) -> tuple[Any, RoborockException | None]:
         async with async_timeout.timeout(timeout):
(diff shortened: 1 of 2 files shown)


roborock/roborock_queue.py
```diff
@@ -12,7 +13,9 @@ def __init__(self, protocol: int, *args):
         self.protocol = protocol
 
     async def async_put(self, item: tuple[Any, RoborockException | None], timeout: float | int) -> None:
-        return await asyncio.wait_for(self.put(item), timeout=timeout)
+        async with async_timeout.timeout(timeout):
+            await self.put(item)
```
Lash-L, line 17: We can put return back here if you would like, but I think it is not needed, since put does not return anything

PR #9, opened 2023-03-25
Moved map parser to package for future cleanup

Lash-L:
In the future I'll take the time to fully convert the component to use this part of the package instead, but for now I am moving it here so that it is possible to do in the future

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ async-timeout = "*"
 pycryptodome = "~3.17.0"
 pycryptodomex = {version = "~3.17.0", markers = "sys_platform == 'darwin'"}
 paho-mqtt = "~1.6.1"
+Pillow = "*"
 
 
 [build-system]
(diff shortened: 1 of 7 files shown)


humbertogontijo:
To be honest we should use a library or include the map with @PiotrMachowski integration. No point on keeping a copy of his code

Lash-L:
Agreed that would be best case. I don't believe he has a pypi package for his map parser though - and we'll need one when this part of the codebase eventually makes it to core (who knows how long that will be). I could potentially make one if he would like.

PiotrMachowski:
I think I could make a package out of it, but it would require a significant bit of work. Before that I would prefer to use a different image rendering library - pillow seems to be a little bit slow (maybe my code is not optimized enough though). 

Recently I have been working more on the Map Card instead of Map Extractor so it has some pending maintenance.

Lash-L:
@PiotrMachowski awesome - let me know if I can help in anyway

PiotrMachowski:
@Lash-L can you make a feature request in Map Extractor to keep a record of this discussion in my repo as well?

Lash-L:
> @Lash-L can you make a feature request in Map Extractor to keep a record of this discussion in my repo as well?

Done.

PR #10, opened 2023-03-27
Add Docking Information

Lash-L:
I don't have a dock so I would wait to set the version for the custom component until I have someone test this. I just wanted to get this added here so I can give them a demo version of the custom component to try.

- added new commands
- Added dock summary, which changes based on which dock you have
- changed to local imports rather than full package imports


humbertogontijo:
I'll release 0.1.11.dev out of this PR

PR #11, opened 2023-03-28
feat: Add automatic release

Lash-L:
So - this is something I've done for a few other pypi projects, but I've never done it in the middle of the project like this. So it may take some testing/ tuning.

You'll have to get an access token for pypi https://test.pypi.org/help/#apitoken

and a github token. And allow actions permissions to write on the repository.

This will when you merge into main, automatically create a changelog, bump the version, put out a release on github, and then put out a pypi release.

We need a changelog for HA core.

It means we should tag our commits with chore: (something small, doesn't need a version bump)

fix: Fixes an existing bump, needs a small version bump

feat: A new feature

If you don't want to do this, that is fine, but we do need to create a changelog at some point for HA if nothing else.

I'd add this one before the other PR as I manually set the version to catch up with the last one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,5 +68,5 @@ jobs:
             -   name: Python Semantic Release
                 uses: relekang/python-semantic-release@v7.33.2
                 with:
-                    github_token: ${{ secrets.GITHUB_TOKEN }}
+                    github_token: ${{ secrets.GH_TOKEN }}
                     pypi_token: ${{ secrets.PYPI_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,17 @@ concurrency:
     cancel-in-progress: true
 
 jobs:
+    # Make sure commit messages follow the conventional commits convention:
+    # https://www.conventionalcommits.org
+    commitlint:
+        name: Lint Commit Messages
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: actions/checkout@v3
+                with:
+                    fetch-depth: 0
+            -   uses: wagoid/commitlint-github-action@v5.3.0
+
     test:
         strategy:
             fail-fast: false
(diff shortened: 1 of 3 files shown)


humbertogontijo (changes requested):
I was about to reject this PR but found out that GH have scoped tokens now.

A few changes are needed: 
1. GITHUB_TOKEN is not a valid secret name. Can you change it to GH_TOKEN?
2. Can you change it to deploy to test pypi first? And we change it to main pypi if everything works

Lash-L:
> I was about to reject this PR but found out that GH have scoped tokens now.

Yep - that's what is nice about it, you're in full control and it can only affect this project.


> A few changes are needed:
> 
> 1. GITHUB_TOKEN is not a valid secret name. Can you change it to GH_TOKEN?

Done

> 2. Can you change it to deploy to test pypi first? And we change it to main pypi if everything works

To be honest with you, I'm not sure how. python-semantic-release technically has an option for repository url - so potentially that?
https://python-semantic-release.readthedocs.io/en/latest/#distributing-release-on-pypi-or-custom-repository

humbertogontijo:
F**k, lets yolo it

Lash-L:
> F**k, lets yolo it

haha, my kind of guy

I think it may be failing because there is no current github release? So it thinks it is 0.0.0 Try creating a tag and a release and see if that works.

humbertogontijo:
I added the tag and release and re ran it.
Maybe it is related to version_variable and version_toml variables in toml file. Are these supposed to be used together?

Lash-L:
They should be able to - I do something similar in another project.

I think it needs https://python-semantic-release.readthedocs.io/en/latest/configuration.html#version-source to be tag so that it is looking at tags for the release version instead ofthe commits.

                uses: relekang/python-semantic-release@v7.33.2
                with:
                    github_token: ${{ secrets.GH_TOKEN }}
                    pypi_token: ${{ secrets.PYPI_TOKEN }}
                    version_source: tag

humbertogontijo:
Lets try it. Create a PR with it and I'll approve it

PR #12, opened 2023-03-28
fix: add version source

Lash-L:


humbertogontijo:
Still got it from the commits
debug: get_current_release_version_by_commits -> 0.0.0

Lash-L:
> Still got it from the commits
> 
> `debug: get_current_release_version_by_commits -> 0.0.0`

Potentially we need to tag one of the commits? 

I can try in 30 min or so

PR #13, opened 2023-03-28
fix: version source from pyproject.toml

humbertogontijo:


Lash-L:
Go into repo settings and give actions read and write and what ever other permissions you can

humbertogontijo:
It worked now

Lash-L:
Seems like it passed but didn't work for some reason.

The ci says "No release will be made." it didn't make a changelog or release on pypi/github.

When I do it here: https://github.com/Lash-L/anova_wifi/commits/main

github actions has a commit, it makes a gh release, and pushes to pypi, for some reason this says it was v 2.1 but it didn't make any changes. perhaps because of "tag_only"? "If set to tag_only, then version_variable is ignored and no changes are made or committed to local config files. The current version from the latest tag matching vX.Y.Z. This won’t change the source defined in [version_variable](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#config-version-variable)."

humbertogontijo:
Good catch. Changed from tag_only to tag and it created a release

Lash-L:
The only thing I'm still not sure about is that it doesn't seem to actually make the changes.

If you look here: https://github.com/Lash-L/anova_wifi/actions/runs/4452317606/jobs/7819888701

On line 51 inside Semantic release it states that it is making a commit, and it updates the version in the pyproject.toml and the changelog.

This doesn't seem to do that.

Under repo settings, Actions, I have "Allow all actions" selected and Read and write permission for workflow permissions and allow to create and approve pull request. Potentially it is one of those things?

humbertogontijo:
Create and approve PR was not checked

PR #17, opened 2023-03-29
fix: bug with dock commands

Lash-L:
Added test and fix bugs with the docking commands.


humbertogontijo:
What is the return of commands RoborockCommand.GET_WASH_TOWEL_MODE and RoborockCommand.GET_DUST_COLLECTION_MOD ?

Lash-L:
A single object dict. 

{"mode": 0} or {"wash_mode": 0} Wasn't sure if it was worth creating a container for it

humbertogontijo:
I think we should, cuz dust_collection and washing_mode variables could be None

Lash-L:
Okay sounds  good - I can make that change in a bit. Let me see if the user has any more bugs or feedback and I'll group it.

PR #23, opened 2023-04-05
feat: add initial support for room mapping

Lash-L:
@humbertogontijo Want your opinion.

the command get_room_mapping returns a list of [room_name_id, segment_id]

In order to use that data I need the HomeDataRoom which has room_name_id, room_name

What's the best way to handle that? 

Do it in the coordinator?

Pass home data into RoborockClient? Then access room on the homeData object? 

Something else?

What are your thoughts?


humbertogontijo:
I will create an entity to share states between vacuum and camera entities for the next release. You can use this new entity that will be created in the __init__.py

Lash-L:
> I will create an entity to share states between vacuum and camera entities for the next release. You can use this new entity that will be created in the **init**.py

Awesome - let me know when it is up.

PR #24, opened 2023-04-05
feat: add networking function

Lash-L:
I didn't realize you figured out the networking function (if you feel like it, i'd be curious to know how)

Added a function to get networking information. 

At least in the core version, I'm going to make it so that devices will store their mac + ip.


humbertogontijo:
I usually get commands from https://github.com/marcelrv/XiaomiRobotVacuumProtocol or from sniffing the mobile app

Issue #25, opened 2023-04-06
RoborockCode does not inherit RoborockEnum functionality.


Lash-L:
@humbertogontijo Not sure about this one as I've never really used __new__ functionality before, so not sure the best approach to fix.

But compilers don't realize that RoborockCode has subscriptable functionality or values() or items()

Unless if I'm doing something wrong, VS code yells at me for attempting to do any of that (but it still works)

humbertogontijo:
We can use a function that returns RoborockEnum instead of the RoborockCode class

Lash-L:
> We can use a function that returns RoborockEnum instead of the RoborockCode class

What function would that be? 

Are you implementing it or would you like me to?

humbertogontijo:
Same thing as the __new__ function. I'm focusing on stability of local integration right now. Could you do it?

Lash-L:
> Same thing as the __new__ function. I'm focusing on stability of local integration right now. Could you do it?

I'll give it a shot in a bit

Lash-L:
@humbertogontijo speaking of stability - have you run into issues with changing the fan mode on local? Seems to take longer than 4 seconds and then the api becomes unresponsive - been testing rather extensively as I was planning to switch to offline for the core integration

humbertogontijo:
Yeah, there is something weird happening. I'm able to run all commands listed in CommandInfoMap using the api repository, but when using the integration a good amount is failing. I'm investigating it

Lash-L:
Okay - I'll keep an eye out for anything you put out. If it seems like too big of a hurdle to overcome right now, I may just put out the online version, and add the offline version in the future. The core maintainer is thinking the repository is pretty close to being ready to be merged in to core. 

Thanks!

PR #26, opened 2023-04-06
fix: make roborockCode through function

Lash-L:
I think this is what you meant. Seems to fix issues as well.

I attempted to create RoborockCode a subclass of RoborockEnum, but it gave errors being unpacked in the import, so I went this approach as well.

per #25


humbertogontijo:
Yes, this is what I meant

PR #27, opened 2023-04-07
fix: serialize nested objects

Lash-L:
So - This was part of a problem I was having, where I wasn't getting a fully serialized version of the object.

But there is a rather large secondary problem with the new dict approach. the keys don't match up. i.e. "geoName" != "geo_name" for HomeData. This is a problem that will reoccur a lot.

Not sure the best approach, I can put some thought into it tomorrow unless you have any ideas

### DIFF
--- a/roborock/containers.py
+++ b/roborock/containers.py
@@ -29,7 +29,19 @@ class RoborockBase:
         return from_dict(cls, decamelize(data), config=Config(cast=[Enum]))
 
     def as_dict(self):
-        return self.__dict__
+        dict = self.__dict__
+        for key, value in dict.items():
+            if isinstance(value, RoborockBase):
+                dict[key] = value.as_dict()
+            if isinstance(value, list):
+                new_list = []
+                for val in value:
+                    if isinstance(val, RoborockBase):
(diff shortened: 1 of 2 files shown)


Lash-L:
@humbertogontijo my thoughts for a fix: 

Anywhere the variable is different in python vs their rest calls, we can add a secondary variable to the data class. This one will also be optional, but then we use post init in data class and we check if the pythonic variable is empty, set it = the rest variable.

Will be slower - but should fix our problem

humbertogontijo:
Dataclasses have as_dict function. Shouldn't we use that?

Lash-L:
Well as_dict doesn't solve our problem because the keys are different in python vs the api, right?

the api returns "geoNames" whereas python naming is geo_names.

so because geoNames != geo_names, running as_dict or from_dict will result in losing the value.

Am I making sense?

humbertogontijo:
Hmm, thats true

humbertogontijo:
Fixed in https://github.com/humbertogontijo/python-roborock/commit/bf52b44b01e93000268c9fa274a3449ac3f82e36

PR #28, opened 2023-04-07
fix: changed prefix for set_custom_mode

Lash-L:
I connected to my phone through wireshark, and the prefix for custom mode was x00w not x00\x87.

I was timing out and the command would end up sending after 10 seconds, but with this fix, it is instant.

If you merge this and #27 in, I think I have fixed all of the bugs for the core version (outside of the dict issues, but I'm not sure if there are any direct bugs from that.


humbertogontijo:
I believe that RoborockCommand.RESUME_ZONED_CLEAN and RoborockCommand.RESUME_SEGMENT_CLEAN are also not working. Could you check it?

Lash-L:
> I believe that RoborockCommand.RESUME_ZONED_CLEAN and RoborockCommand.RESUME_SEGMENT_CLEAN are also not working. Could you check it?

Yeah - I can later today. Right now isn't a good time to run the vacuum for me, but I can test it later today

Can we merge this one in for now and I'll test the others later?

humbertogontijo:
Yeap

PR #29, opened 2023-04-07
fix: assume device prop attr can be none

Lash-L:
We already assume that it can be none whenever we update it, but we don't type it to correspond to that fact.

### DIFF
--- a/roborock/typing.py
+++ b/roborock/typing.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import typing
 from dataclasses import dataclass
 from enum import Enum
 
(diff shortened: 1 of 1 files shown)


Lash-L:
@humbertogontijo When you get a chance this one is blocking for me to continue core.

roborock/typing.py
```diff
@@ -208,8 +209,10 @@ def __init__(self, dust_collection_mode: DustCollectionMode,
 
 
 class RoborockDeviceProp:
-    def __init__(self, status: Status, dnd_timer: DNDTimer, clean_summary: CleanSummary, consumable: Consumable,
-                 last_clean_record: CleanRecord = None, dock_summary: RoborockDockSummary = None):
+    def __init__(self, status: typing.Optional[Status] = None, dnd_timer: typing.Optional[DNDTimer] = None,
```
humbertogontijo: Convert it to a dataclass instead then

humbertogontijo (changes requested):

humbertogontijo (approved):

PR #30, opened 2023-04-08
fix: changed prefixes for debugged commands

Lash-L:
I accidentally only debuged segment clean, I'll try to do zoned clean later.


Lash-L:
Went through all of the commands I have logged and double checked everything. I don't have a history of all the commands you do - but for the 55 commands I have, this is what it looks like

humbertogontijo (approved):
Looks good! Thank you
Good job getting the core PR approved btw

Lash-L:
@humbertogontijo Thanks for your help answering some of the questions. And obviously for building this to begin with. 

Once it gets merged in, it becomes a lot easier to add things and hopefully within a few months it will have the same functionality as the custom one

PR #35, opened 2023-04-13
feat: new styling

Lash-L:
Added styling using ruff/ isort/ black

Can be run by doing running pre-commit, which can be configured to run automatically every time you commit. We can manipulate styling a bit if you don't like anything

Not really a feat, but I'd like this to put out a new version when merged


humbertogontijo:
Looks good. Let's just keep an eye if this is not going to conflict with the file CHANGELOG.md on every change

Lash-L:
Ah good call - I can add an exclude for the change log tomorrow. I think I'm done for the day tonight

PR #41, opened 2023-04-25
feat: add room mapping

Lash-L:


Lash-L:
Fixed some lint but not all that is from the local client changes.

roborock/local_api.py
```diff
@@ -94,6 +94,13 @@ async def send_message(self, device_id: str, roborock_messages: list[RoborockMes
             return_exceptions=True,
         )
 
+    async def get_room_mapping(self, device_id: str) -> list[RoomMapping]:
```
humbertogontijo: Since this is doable for both local and mqtt. This should go to parent class

humbertogontijo (changes requested):

roborock/local_api.py
Lash-L: room_mapping in mqtt is unreliable from my experiences. Most of the time it gives no information back. Do you have different results?

humbertogontijo: It looks reliable to me. Smoke testing it worked all times

Lash-L: actually - it seems like it may work for mqtt now. Perhaps something has changed with the prefix changes? I know it used to not give any real results. I will move it

humbertogontijo: Could be. Refactoring for local affected mqtt client, which fixed some issues with responses

Lash-L: Okay fixed

humbertogontijo:
Do you know the meaning of the third item?

Lash-L:
> Do you know the meaning of the third item?

Third item?

humbertogontijo:
DEBUG:roborock.local_api:id=24684 Response from get_room_mapping: [[16, '2396684', 13], [17, '2396658', 7], [18, '2396663', 3], [19, '2396661', 15], [20, '2396667', 8], [21, '2396647', 2], [23, '2396669', 14], [24, '2396683', 7], [25, '2396665', 15]]

Lash-L:
Weird...
```
[[16, '2537175'], [17, '2537176'], [18, '2537178'], [19, '2537174'], [20, '3009519'], [21, '2957102'], [22, '2537149'], [23, '2537147']]
```

Lash-L:
Do both ids correspond to the segments for your rooms?

i.e. if you were to clean 15 and 19 would they clean the same area?

humbertogontijo:
I just noticed that my command app_clean_segment is not working for prefix b"\x00\x00\x00\x67" but it works with b"\x00\x00\x00\x87"

Lash-L:
> I just noticed that my command app_clean_segment is not working for prefix b"\x00\x00\x00\x67" but it works with b"\x00\x00\x00\x87"

Ah I have not actually tried it- is that one I changed? I'm in mobile right now so I can't check git history/ check if it is the same for me

Lash-L:
As a quick check it does not appear to be working with the current settings on my HA app by calling the service. So the prefix change sounds good

roborock/api.py
```diff
@@ -318,6 +319,13 @@ async def get_networking(self, device_id) -> NetworkInfo | None:
             _LOGGER.error(e)
         return None
 
+    async def get_room_mapping(self, device_id: str) -> list[RoomMapping]:
+        """Gets the mapping from segment id -> iot id. Only works on local api."""
+        mapping = await self.send_command(device_id, RoborockCommand.GET_ROOM_MAPPING)
+        if isinstance(mapping, list):
+            return [RoomMapping(room[0], room[1]) for room in mapping]
```
humbertogontijo:
```suggestion
            return [
                RoomMapping(segment_id, iot_id)
                for segment_id, iot_id in [unpack_list(room, 2) for room in mapping]
            ]
```

humbertogontijo (changes requested):

humbertogontijo (approved):

PR #42, opened 2023-04-27
feat: added error check and deviceprop functionality for core

Lash-L:
- Added time left for consumable with post init
- Added update() function back to DeviceProp
- Added extra error checking for the web_api

### DIFF
--- a/roborock/containers.py
+++ b/roborock/containers.py
@@ -285,10 +285,11 @@ class Consumable(RoborockBase):
     sensor_time_left: Optional[int] = None
 
     def __post_init__(self):
-        self.main_brush_time_left = 1080000 - self.main_brush_work_time
-        self.side_brush_time_left = 720000 - self.side_brush_work_time
-        self.filter_time_left =540000 - self.filter_work_time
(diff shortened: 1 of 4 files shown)


roborock/containers.py
```diff
@@ -279,6 +279,16 @@ class Consumable(RoborockBase):
     strainer_work_times: Optional[int] = None
     dust_collection_work_times: Optional[int] = None
     cleaning_brush_work_times: Optional[int] = None
+    main_brush_time_left: Optional[int] = None
+    side_brush_time_left: Optional[int] = None
+    filter_time_left: Optional[int] = None
+    sensor_time_left: Optional[int] = None
+
+    def __post_init__(self):
+        self.main_brush_time_left = 1080000 - self.main_brush_work_time if self.main_brush_work_time else None
```
humbertogontijo: Let's use constants for these values

humbertogontijo (changes requested):

roborock/containers.py
Lash-L: done

humbertogontijo (approved):

PR #43, opened 2023-04-27
fix: add functionality for missing enum values

Lash-L:
Added a check if it is missing, return "UNKNOWN" which is now on all enums, but does not show up in lists.

Thoughts?

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -14,7 +14,7 @@ import secrets
 import struct
 import time
 from random import randint
-from typing import Any, Callable, Coroutine, Optional
+from typing import Any, Callable, Coroutine, Mapping, Optional
 
 import aiohttp
 from Crypto.Cipher import AES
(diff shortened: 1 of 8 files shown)


Lash-L:
code_mappings is where my code changes are, the rest is just reverted code

humbertogontijo:
It's late for me, but I'll try this out as soon as I can. I still feel like if this value is 207 it still throw an error cus the vacuum will not recognize it. But I'll try it out

Lash-L:
> It's late for me, but I'll try this out as soon as I can. I still feel like if this value is 207 it still throw an error cus the vacuum will not recognize it. But I'll try it out.
> Anyway, this might be better then making the integration useless

I'm surprised we haven't run into this before. But I guess with the new enum system, winces it's just on main, no one with an s6 maxv has tried it? 

Look at my test in test container that was my proof of concept

PR #44, opened 2023-04-28
feat: add support for old mop and vacuum codes

Lash-L:
Thoughts on this approach?

We can expand it as we find device specific functionality. but the device object is always around so it is easy to access. 

Alternatively we could do an enum but that felt like overkill for right now.

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -35,7 +35,7 @@ from .containers import (
     SmartWashParams,
     Status,
     UserData,
-    WashTowelMode,
+    WashTowelMode, StatusOldModes,
 )
 from .exceptions import (
     RoborockAccountDoesNotExist,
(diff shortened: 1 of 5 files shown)


humbertogontijo:
Not really linking this approach.
Check out https://github.com/rytilahti/python-miio/blob/313707d27c6766bdc80417ab4faa4d54a609dcff/miio/integrations/roborock/vacuum/vacuum.py#L651

Lash-L:
> Not really linking this approach. Check out https://github.com/rytilahti/python-miio/blob/313707d27c6766bdc80417ab4faa4d54a609dcff/miio/integrations/roborock/vacuum/vacuum.py#L651

Will do when I get a chance

Lash-L:
@humbertogontijo Are you sold on our current StrEnum approach for our code enums?

Any objection to doing it similar to how it is done in the link above - regular enums that get turned into a dict? Esp cause we don't actually have real typing with these and we couldn't subclass them very well/ easily to describe all "FanCodeEnums" or something like that. 

What do you think?

humbertogontijo:
I'm not happy with current strenum for codes.
No objection

Lash-L:
Unfortunately I couldn't figure out how to add a missing value by default - so I just did it in the ones where they weren't device specific/ I thought we may run into a missing error in the future, which is not ideal

PR #45, opened 2023-04-29
Feat/single device api

humbertogontijo:


roborock/typing.py
```diff
@@ -17,192 +17,288 @@
...
+    RoborockCommand.APP_START: CommandInfo(prefix=b"\x00\x00\x00\x87", params=[{"use_new_map": 1}]),
+    RoborockCommand.APP_START_WASH: CommandInfo(prefix=b"\x00\x00\x00w", params=None),
+    RoborockCommand.APP_STAT: CommandInfo(
+        prefix=b"\x00\x00\x01\xa7",
+        params=[
+            {
+                "ver": "0.1",
+                "data": [
+                    {
+                        "times": [1682723478],
```
Lash-L, line 161: hard coded time?

roborock/typing.py
```diff
@@ -17,192 +17,288 @@
...
+    RoborockCommand.START_CAMERA_PREVIEW: CommandInfo(
+        prefix=b"\x00\x00\x00\xa7", params={"client_id": "443f8636", "quality": "SD"}
+    ),
+    RoborockCommand.START_EDIT_MAP: CommandInfo(prefix=b"\x00\x00\x00w", params=[]),
+    RoborockCommand.START_WASH_THEN_CHARGE: CommandInfo(prefix=b"\x00\x00\x00\x87", params=None),
+    RoborockCommand.STOP_CAMERA_PREVIEW: CommandInfo(prefix=b"\x00\x00\x00\x97", params={"client_id": "443f8636"}),
+    RoborockCommand.SWITCH_WATER_MARK: CommandInfo(prefix=b"\x00\x00\x00\x97", params={"waterMark": "OFF"}),
+    RoborockCommand.TEST_SOUND_VOLUME: CommandInfo(prefix=b"\x00\x00\x00w", params=None),
+    RoborockCommand.UPD_SERVER_TIMER: CommandInfo(prefix=b"\x00\x00\x00\x97", params=None),
 }
```
Lash-L, line 302: I like this approach  of having the params attached, but I'm confused why they are often pre-filled out? What is the desire with the params attribute? Is it just to show an example?

Also I think it would be helpful to add a description parameter that is defaulted to "", that way we can start describing some of these commands to make it easier during development. I can go through and add a lot of them if you'd like/ think thats a good idea

Lash-L (commented):
I definitely like your single device per api approach, makes things a lot easier especially as we look towards adding support for functionality for specific devices. Let me know if you want my help with anything.

humbertogontijo:
I just used raw data for now. The idea will be to add validators for each command

Lash-L:
It seems all the issues in core are solved (outside of the device specific one)

So i'm 100% good with having our main branch go ahead and make this change if you are

humbertogontijo:
@Lash-L I reversed engineered device broadcast. Are you familiar with zeroconf?

Lash-L:
@humbertogontijo i am - I didn't really Roborock broadcasted in zeroconf, I was thinking about adding dhcp at some point

humbertogontijo:
Dhcp is not sufficient since we can't map hostname to local_key. The device is broadcast it's device_id

Lash-L:
Gotcha - I'll be curious to see what you found / reverse engineered

humbertogontijo:
How do I register a service for zeroconf? Is it possible to register a service based on a parsed broadcast data?
I saw that are service names like _xiaomi._udp._local, but how that was registered?

Lash-L:
I'm out right now so I can't check - but my understanding is you add it to manifest then homeassistant auto applies it.

You also add it to config flow

humbertogontijo:
But I need to know how to configure decoding the device broadcast first. The manifest only expects the service name. How do I add that service?

Lash-L:
I can try to check when I'm home

PR #46, opened 2023-04-29
feat: add specific device functionality

Lash-L:
based this off of your single device work.

This is still very much WIP. but it is a basis and I thought I would publish it to see if you had any feedback

### DIFF
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -234,6 +234,7 @@ class RoborockDockTypeCode(RoborockEnum):
     no_dock = 0
     empty_wash_fill_dock = 3
     auto_empty_dock_pure = 5
+    s8_dock = 7
 
 
 class RoborockDockDustCollectionModeCode(RoborockEnum):
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import logging
 from enum import Enum
 from typing import Any, Type, TypeVar
 
 _StrEnumT = TypeVar("_StrEnumT", bound="RoborockEnum")
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class RoborockEnum(str, Enum):
@@ -18,11 +21,15 @@ class RoborockEnum(str, Enum):
 
     @classmethod
     def _missing_(cls: Type[_StrEnumT], code: object):
-        return cls._member_map_.get(str(code))
+        if cls._member_map_.get(str(code)):
+            return cls._member_map_.get(str(code))
+        else:
+            _LOGGER.warning(f"Unknown code {code} for {cls.__name__}")
+            return cls._member_map_.get(str(-9999))
 
     @classmethod
     def as_dict(cls: Type[_StrEnumT]):
-        return {int(i.name): i.value for i in cls}
+        return {int(i.name): i.value for i in cls if i.value != "UNKNOWN"}
 
     @classmethod
     def values(cls: Type[_StrEnumT]):
@@ -42,6 +49,7 @@ class RoborockEnum(str, Enum):
 
 
 def create_code_enum(name: str, data: dict) -> RoborockEnum:
+    data[-9999] = "UNKNOWN"
     return RoborockEnum(name, {str(key): value for key, value in data.items()})
 
 
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -22,7 +22,8 @@ class RoborockEnum(IntEnum):
 
     @classmethod
     def __missing__(cls: Type[RoborockEnum], key) -> str:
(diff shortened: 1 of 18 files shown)


Lash-L:
I'm also adding a vacuummodelspecification which states which of the enum to use

roborock/code_mappings.py
```diff
@@ -1,183 +1,228 @@
...
-RoborockDockWashTowelModeCode = create_code_enum(
-    "RoborockDockWashTowelModeCode",
-    {
-        0: "light",
-        1: "balanced",
-        2: "deep",
-    },
-)
+class RoborockStateCode(RoborockEnum):
+    starting = 1
```
humbertogontijo, line 54: Shouldn't enum names be upper case? We could create a function to return it lower

humbertogontijo (changes requested):
Looks much better

roborock/code_mappings.py
Lash-L: What would that function be? We would just call it on top of each time we need it lowercase?

I was thinking about translations of the entities, it my head it made more sense to be lower case but it could be upper case

humbertogontijo: Let me test this when I have the chance. My first impression was that this is non standard. But if it helps with some scenarios I can work with it

Lash-L: I wish I pushed my changes so that you could see it/ use it. But basically I have a function that builds a vacuum mode specification based on the model str that determines which e has to use

roborock/containers.py
```diff
@@ -132,6 +144,14 @@ class HomeDataDeviceStatus(RoborockBase):
     capability: Optional[Any] = None
     category: Optional[Any] = None
     schema: Optional[Any] = None
+    model_specification: ModelSpecification | None = None
```
humbertogontijo, line 150: Ops. I'm pretty sure this class is wrong. Might have been an issue on my end when coping and pasting all these entities

Lash-L:
@humbertogontijo This still has more work - but let me know what you think conceptually whenever you have a chance - no rush

humbertogontijo:
Overall it is looking fine

Lash-L:
Code is tested and working - you can see how I tested implementing it here: https://github.com/Lash-L/core/tree/roborock_device_specific/homeassistant/components/roborock

roborock/roborock_typing.py
```diff
@@ -275,7 +275,7 @@ class CommandInfo:
     RoborockCommand.SET_CUSTOMIZE_CLEAN_MODE: CommandInfo(
         prefix=b"\x00\x00\x00\xa7", params={"data": [], "need_retry": 1}
     ),
-    RoborockCommand.SET_CUSTOM_MODE: CommandInfo(prefix=b"\x00\x00\x00\x87", params=[108]),
+    RoborockCommand.SET_CUSTOM_MODE: CommandInfo(prefix=b"\x00\x00\x00w", params=[108]),
```
Lash-L, line 278: I had to change this to get it to work - was this change made on purpose?

humbertogontijo: I sniffed commands and merged them into CommandInfoMap. It is strange that the app sends invalid prefixes haha

Lash-L: Idk maybe there's some magic resting to prefixes we don't know yet haha. 

Does the prefix appear to be wrong for you as well?

humbertogontijo:
After this gets merged, should we use dicts to parse the enum value?
We need the translation key for state within extra_attributes (for lovelace xiaomi vacuum card) and for error sensor

roborock/roborock_typing.py
humbertogontijo: Probably. I still don't know much about these 4 heading bytes
Yes, I had to apply your change for it to work.

Lash-L:
What do you mean by that?

humbertogontijo:
Oh, I tried it just by logging the return of a command. But didn't verified that it was actually returning the enum for state and error_code

Lash-L:
I think this in its current form would work for that - this https://github.com/Lash-L/core/blob/34f175e59a722ce9a1c12bf0f64c2d079e446f4b/homeassistant/components/roborock/vacuum.py#L27 correctly maps, so I assume the errors would work as well. And whenever we need the string, we just call .name like I do herehttps://github.com/Lash-L/core/blob/34f175e59a722ce9a1c12bf0f64c2d079e446f4b/homeassistant/components/roborock/select.py#L44

humbertogontijo (approved):

humbertogontijo:
Had to make a small change for translations to work
https://github.com/humbertogontijo/python-roborock/commit/c559d40183e47ef8698651281ae8946a99cb897e

Lash-L:
You were able to get fan speed translations to work? How do? It wasn't translating for me when I tried

humbertogontijo:
I'm not there yet. Still making necessary changes
I'm stuck with mop_mode schema for service vacuum_set_mop_mode

PR #49, opened 2023-05-04
feat: add error check for invalid user agreement

Lash-L:


Lash-L:
@humbertogontijo sorry this commitlint is wrong, should be fine with a squish merge though as it should take the title I think

Lash-L:
This also fixes a dependency issue:
```
    The user requested construct==2.10.56
    nibe 2.1.4 depends on construct<3.0.0 and >=2.10.0
    pyps4-2ndscreen 1.3.1 depends on construct>=2.9.45
    python-miio 0.5.12 depends on construct<3.0.0 and >=2.10.56
    python-roborock 0.10.3 depends on construct<3.0.0 and >=2.10.68
    ```
    
    we don't need > 2.10.56 for anything specific do we?

File: roborock/api.py [hunk 1]
```diff
@@ -474,6 +476,12 @@ async def code_login(self, code) -> UserData:
         if response_code != 200:
             if response_code == 2018:
                 raise RoborockInvalidCode("Invalid code - check your code and try again.")
+            if response_code == 3009:
+                raise RoborockNoUserAgreement("You must accept the user agreement in the Roborock app to continue.")
+            if response_code == 3006:
+                raise RoborockInvalidUserAgreement(
+                    "User agreement must be accepted again - or you are attempting to use the Mi Home app."
```
humbertogontijo: "User agreement must be accepted again - or you are attempting to use the Mi Home app account." maybe?

Lash-L: 👍

PR #51, opened 2023-05-05
feat: extending device status by device model

humbertogontijo:


humbertogontijo:
@Lash-L I invited you as a colaborator so I can add you as reviewer in future changes. Could you review this one?

roborock/containers.py
```diff
@@ -261,12 +249,60 @@ class Status(RoborockBase):
...
+
+
+@dataclass
+class S8ProUltraStatus(Status):
+    fan_power: Optional[RoborockFanSpeedS7MaxV] = None
+    water_box_mode: Optional[RoborockMopIntensityS7] = None
+    mop_mode: Optional[RoborockMopModeS8ProUltra] = None
+
+
+ModelStatus: dict[str, Type[Status]] = {
```
Lash-L, line 297: I think keeping this in normal dict style would be nice to make it clear.I was confused when I saw it above,, I thought it might have been some kind of generic. 
But that's just a styling thing and not really important

roborock/containers.py
```diff
@@ -261,12 +249,60 @@ class Status(RoborockBase):
...
+    fan_power: Optional[RoborockFanSpeedS7] = None
+    water_box_mode: Optional[RoborockMopIntensityS7] = None
+    mop_mode: Optional[RoborockMopModeS7] = None
+
+
+@dataclass
+class S8ProUltraStatus(Status):
+    fan_power: Optional[RoborockFanSpeedS7MaxV] = None
+    water_box_mode: Optional[RoborockMopIntensityS7] = None
+    mop_mode: Optional[RoborockMopModeS8ProUltra] = None
```
Lash-L, line 294: I think these shouldn't be optional. We know they exist for that model so they should just be empty right?

But then I think we need what they don't have to be None, so that we don't have to do a hasattr check, we can just do a none check

roborock/containers.py
humbertogontijo: What you mean by normal dict style? The variable name?

roborock/containers.py
humbertogontijo: Oh, I haven't thought about that. You are right

roborock/containers.py
Lash-L: yeah model_status vs ModelStatus. not a big deal though

roborock/containers.py
humbertogontijo: I'm annoyed by mypy sorting consts in imports different. I can't make PyCharm work with it 😅
The easy way was to make this const look like a class (This is bad tho)

Lash-L: Oh yeah I'm sure theres some mypy setting we could change, but I feel like theres so many mypy settings it feels like information overload.

humbertogontijo:
> I think this all looks good except my one typing comment.
> 
> Perhaps it makes sense to make Another dataclass called ModelStatus or something that the others inherit from, and it has all of the possibilities set to None. So you don't have to super repeat all of the non existent attributes for each one.

What would this ModelStatus looks like?

Lash-L:
> What would this ModelStatus looks like?

I realized I was wrong.

Status already has these enums as None, so they don't need to be redeclared as None as they are inherited as None from Status.

humbertogontijo:
> > What would this ModelStatus looks like?
> 
> I realized I was wrong.
> 
> Status already has these enums as None, so they don't need to be redeclared as None as they are inherited as None from Status.

The only way I can think of is creating a mixing for each model status

Lash-L:
Sorry I think I wasn't very clear. I think how you have it is actually fine.

The base Status class already has 'fan_power' for instance, and it is set to None.

The specific S6PureStatus for example, just has to set the type of 'fan_power'

Then since the base Status class already has 'mop_mode' and it is None, S6PureStatus will also make it None.

I think the approach you have now is completely fine unless I am missing something

humbertogontijo:
I wanted to make these fields required. But I think I'll just let it be optional for now

PR #53, opened 2023-05-05
fix: changed incorrect s8 pro ultra string

### LINKED ISSUE #52: Vacuum status reported as an int rather than being mapped to a string
Until yesterday, I had string based vacuum statuses in Home Assistant, now I have int statuses.  I believe this is because the model string for my S8 Pro Ultra is incorrect in the library.

this line https://github.com/humbertogontijo/python-roborock/blob/main/roborock/const.py#L33 should be `roborock.vacuum.a70` for the new S8 Pro Ultra

cryptk:
This fixes #52 , verified the model string with my own S8 Pro Ultra.


Lash-L (approved):

Lash-L:
That's okay I'll squash it and it will use the title

humbertogontijo:
> @humbertogontijo do you mind if I merge in PRs like this? Small changes? Didn't want to do it before checking it is okay with you

go ahead

PR #58, opened 2023-05-09
feat: add docs

Lash-L:
basic skeleton for docs so I can start adding and testing with readthedocs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@ roborock/__pycache__
 poetry.lock
 *.pyc
 .coverage
+
+# Sphinx documentation
+docs/_build/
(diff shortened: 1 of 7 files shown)


humbertogontijo (approved):

Lash-L:
ah I'm unable to do it with my permissions - can you make a readthedocs account @humbertogontijo ?

Alternatively you could follow these steps:

https://docs.readthedocs.io/en/latest/guides/git-integrations.html

with this webhook: https://readthedocs.org/api/v2/webhook/python-roborock/238448/

humbertogontijo:
Should be working
image

PR #59, opened 2023-05-10
chore: fix readthedocs

Lash-L:


Lash-L:
Turns out there is a recent bug with readthedocs and it needs to specify the build version to correctly work

I haven't figured out when it allows me to merge myself and when it doesnt

Lash-L:
oh jk - I see it lets me do it if you approve it, and I can merge someone else's in if I approve it

Lash-L:
Using this as it's a location where it won't spam other people.

@humbertogontijo theres no way for us to be able to debug log all messages that come in - not just the ones that relate to the commands we see? I.e see data coming in from the app? 

Secondly good call with the 1.0.0 was probably time for a major release!

humbertogontijo:
To see all messages the phone would need to be connected to the HA instance. The only way to debug messages is using that new command added to cli.py. There is no way to intercept it from the network without a "men in the middle" approach

PR #62, opened 2023-05-27
fix: change e2 fan codes

Lash-L:
per #327

### DIFF
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -131,12 +131,11 @@ class RoborockFanSpeedV3(RoborockFanPowerCode):
 
 
 class RoborockFanSpeedE2(RoborockFanPowerCode):
-    gentle = 41
-    silent = 50
-    standard = 68
-    medium = 79
-    turbo = 100
-
+    quiet = 101
+    balanced = 203
+    turbo = 103
+    max = 104
+    gentle = 105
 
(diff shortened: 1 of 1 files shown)


roborock/code_mappings.py
```diff
@@ -131,12 +131,11 @@ class RoborockFanSpeedV3(RoborockFanPowerCode):
...
 
 class RoborockFanSpeedE2(RoborockFanPowerCode):
-    gentle = 41
-    silent = 50
-    standard = 68
-    medium = 79
-    turbo = 100
-
+    quiet = 101
+    balanced = 203
```
humbertogontijo: Probably 102, no?

humbertogontijo (approved):

Lash-L:
@humbertogontijo I was out of town for a bit but now I'm trying to make some progress on some things. 

Any idea what is going on with the disconnects people are having with the map?

Also are there any bugs with the current version of the api? I'm on 17.0 on core. Particularly just any bugs with local connection. Want to know before I bump it

humbertogontijo:
I did some work with the local api adding a ping method that runs in the background every 10 seconds to avoid disconnection on actual commands(it could be the reason of lots of reconnections, I think we can just change the logs to ignore it for the ping command)
For the mqtt api I haven't digged into it to be honest

PR #63, opened 2023-05-30
feat: moving clean area to api

Lash-L:
Core doesn't like calculations being done in core, so I moved the division to the api.


humbertogontijo:
Can you fix mypy for this PR?

Lash-L:
Yes will do in the morning

humbertogontijo:
I dont think we should change the existing field types to Optional[float]

Lash-L:
That was left behind from before I added the new attributes. nice catch, thanks

PR #65, opened 2023-06-03
feat: add time datetime for valley

Lash-L:
Adds Unknown Method error

Time for the dnd and valley.

I need to create a setter for those, but I need to do some reverse engineering as I can't get the set command to actually change the timer.

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -36,6 +36,7 @@ from .containers import (
     Status,
     UserData,
     WashTowelMode,
+    ValleyElectricityTimer,
 )
 from .exceptions import (
     RoborockAccountDoesNotExist,
@@ -47,6 +48,7 @@ from .exceptions import (
     RoborockTimeout,
     RoborockUrlException,
     VacuumError,
+    UnknownMethodError,
 )
 from .protocol import Utils
 from .roborock_future import RoborockFuture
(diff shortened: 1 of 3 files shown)


roborock/api.py
```diff
@@ -189,6 +191,8 @@ async def _async_response(self, request_id: int, protocol_id: int = 0) -> tuple[
             queue = RoborockFuture(protocol_id)
             self._waiting_queue[request_id] = queue
             (response, err) = await queue.async_get(QUEUE_TIMEOUT)
+            if response == "unknown_method":
```
humbertogontijo, line 194: I probably got this a feel times but I cant remember if the response is just this code unknown_method. Is it?

Lash-L: At least sometimes it is. That's the response I got for valley timer as example

Lash-L: I'm not sure why the response is that instead of the err

Lash-L: I would assume it always returns in English and never changes but I'm not 100% on that

humbertogontijo: Valley timer does work for you?

Lash-L: It does not. It's not an option for my device

humbertogontijo (approved):

PR #67, opened 2023-06-05
feat: add diagnostic data and extra containers

Lash-L:
Adding diagnostic data that will allow users to get diagnostic data rather than having to use debug. Also adds stuff for flow led

### DIFF
--- a/roborock/roborock_typing.py
+++ b/roborock/roborock_typing.py
@@ -10,6 +10,7 @@ from .containers import (
     Consumable,
     DnDTimer,
     DustCollectionMode,
+    RoborockBase,
     SmartWashParams,
     Status,
     WashTowelMode,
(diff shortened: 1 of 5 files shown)


roborock/roborock_typing.py
```diff
@@ -306,6 +306,19 @@ class DockSummary:
     wash_towel_mode: Optional[WashTowelMode] = None
     smart_wash_params: Optional[SmartWashParams] = None
 
+    def as_dict(self) -> dict:
```
humbertogontijo: Why not extend RoborockBase?

Lash-L: changed

roborock/roborock_typing.py
```diff
@@ -329,3 +330,13 @@ def update(self, device_prop: DeviceProp) -> None:
             self.last_clean_record = device_prop.last_clean_record
         if device_prop.dock_summary:
             self.dock_summary = device_prop.dock_summary
+
```
humbertogontijo: Same here. We can use RoborockBase

Lash-L: Changed.

humbertogontijo (approved):

PR #70, opened 2023-06-15
chore: add dependabot

Lash-L:
Will help us notice when we are outdated on things like semantic release


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+    schedule:
+      interval: "weekly"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      # Dependabot should not update Home Assistant as that should match the homeassistant key in hacs.json
+      - dependency-name: "homeassistant"
```
humbertogontijo: But there is no homeassistant dependency in this repo

PR #74, opened 2023-06-15
chore: commitlint ignore dependabot

Lash-L:


Lash-L:
semantic release should be fixed now. The manual releases were missing a v before the tag. I have not been able to figure out the issue with 1.0.7 but I will maybe have more time later tonight if you don't have a chance to get to it first

Lash-L:
Ok figured it out, the issue is with 24.0 of python-roborock.

Something with the valley timer and dnd timer being in the get_props.

Which in my opinion, we shouldn't be doing. If you look at how I do it in core, for entities outside of status, I have non-coordinated entities, these entities make the api call directly, that way if they aren't supported or if the user disables that entity, that call wont get made. Or If I update that entity, i.e. a switch on, I can just request the new status of that entity, rather than doing an entire coordinator refresh.

The issue stems with the valley timer, not sure why but that is causing all of the issues.

PR #75, opened 2023-06-15
fix: remove dnd timer and valley electricity from props

Lash-L:
Ok figured it out, the issue is with 24.0 of python-roborock.

Something with the valley timer and dnd timer being in the get_props.

Which in my opinion, we shouldn't be doing. If you look at how I do it in core, for entities outside of status, I have non-coordinated entities, these entities make the api call directly, that way if they aren't supported or if the user disables that entity, that call wont get made. Or If I update that entity, i.e. a switch on, I can just request the new status of that entity, rather than doing an entire coordinator refresh.

The issue stems with the valley timer, not sure why but that is causing all of the issues.

But for now I'm removing these from props as long as that is fine with you. And if I have time, I will try to rework the switches on homeassistant-roborock to match core as long as this is okay with you

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -323,13 +323,11 @@ class RoborockClient:
     @fallback_cache
     async def get_prop(self) -> DeviceProp | None:
         """Gets device general properties."""
-        [status, clean_summary, consumable, dnd_timer, valley_electricity_timer] = await asyncio.gather(
+        [status, clean_summary, consumable] = await asyncio.gather(
             *[
                 self.get_status(),
                 self.get_clean_summary(),
                 self.get_consumable(),
-                self.get_dnd_timer(),
-                self.get_valley_electricity_timer(),
             ]
         )
         last_clean_record = None
@@ -338,13 +336,11 @@ class RoborockClient:
         dock_summary = None
         if status and status.dock_type is not None and status.dock_type != RoborockDockTypeCode.no_dock:
             dock_summary = await self.get_dock_summary(status.dock_type)
-        if any([status, dnd_timer, clean_summary, consumable]):
+        if any([status, clean_summary, consumable]):
             return DeviceProp(
                 status,
                 clean_summary,
                 consumable,
-                dnd_timer,
-                valley_electricity_timer,
                 last_clean_record,
                 dock_summary,
             )
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -103,6 +103,8 @@ class RoborockClient:
         self._last_disconnection = self.time_func()
         self.keep_alive = KEEPALIVE
         self._diagnostic_data: dict[str, dict[str, Any]] = {}
+        self.dnd_timer: DnDTimer | None = None
+        self.valley_timer: ValleyElectricityTimer | None = None
 
     def __del__(self) -> None:
         self.sync_disconnect()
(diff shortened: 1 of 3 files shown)


Lash-L:
The cause of the error is that UnknownMethodException is being raised, but it is being treated as if it is a timeout, so it attempts to connect again

Lash-L:
also seems like we were scheduling a new ping everytime we reconnected, so the amount of pings grew exponential as we reconnected.

humbertogontijo:
What issues you are mentioning here?

Lash-L:
https://github.com/humbertogontijo/homeassistant-roborock/issues/379

Any user who doesn't have valley timer - the version wasn't working

roborock/local_api.py
```diff
@@ -51,6 +51,8 @@ async def keep_alive_func(self, _=None):
             await self.ping()
         except RoborockException:
             pass
+        if self.keep_alive_task is not None:
+            self.keep_alive_task.cancel()
```
humbertogontijo: Is this needed? The task will end after next line

humbertogontijo:
Yeah, using a non coordinated entity is the way to go

roborock/local_api.py
Lash-L: My understanding is that call later returns a obj that can cancel the original callback, but when you call call_later it will run the action regardless of if the response gets overwritten.

I could be wrong though, I Didn’t extensively test

humbertogontijo: Yes. But this is cancelling it during execution. Right after ping was already called and next one is being scheduled
I think there is a cancel already when disconnected

Lash-L: That's the case if the keep_alive_func is called by the call_later in keep alive func, but if keep_alive_func is called in async connect, then the existing keep alive task still exist and will continue to populate itself, but the new one will also be populating itself.

I just removed it for now though because I am not confident.

roborock/api.py
```diff
@@ -103,6 +103,8 @@ def __init__(self, endpoint: str, device_info: DeviceData) -> None:
         self._last_disconnection = self.time_func()
         self.keep_alive = KEEPALIVE
         self._diagnostic_data: dict[str, dict[str, Any]] = {}
+        self.dnd_timer: DnDTimer | None = None
+        self.valley_timer: ValleyElectricityTimer | None = None
```
Lash-L, line 107: Not sure if this is the best way, but I'm trying to avoid doing the coordinator, and rather than place it in device info, cause then I'd have to make a new function call on top of get_dnd_timer to update it, rather than just setting it here.

humbertogontijo: I thought about something similar. Having an object to hold all executed queries with evicting and caching based on GET / (SET / CHANGE) commands.
You can check them out in roborock_typing CacheableCommands

humbertogontijo (approved):

humbertogontijo:
I'll accept this since 1.0.7 is broken due to props not being compatible with all devices

humbertogontijo (approved):

roborock/api.py
Lash-L: Oh that's nice and probably much better long term than this

Lash-L:
Ugh not sure why semantic release is still failing

humbertogontijo:
Yeah. Maybe we need a commit message with current version?

PR #76, opened 2023-06-16
fix: semantic release

Lash-L:
Tested locally and this should work

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "python-roborock"
-version = "0.24.1"
+version = "0.24.2"
 description = "A package to control Roborock vacuums."
 authors = ["humbertogontijo <humbertogontijo@users.noreply.github.com>"]
 license = "GPL-3.0-only"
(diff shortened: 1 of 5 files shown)


humbertogontijo (approved):

Lash-L:
I'm not sure why when it succeded for one of your last runs, it ended up not actually releasing, stating there was no change.

humbertogontijo:
I added a commit message with latest version and that got fixed. The issue now is construct lib failing build

Lash-L:
It doesn't look like it actually did - no release was actually made. Semantic release passed, but it thought there were no changes so it did not push anything to mypy or github

humbertogontijo:
Check the previous action. That's the one that failed

humbertogontijo:
image

Lash-L:
Yeha I'm talking about the release commit that did succeed when you removed python 3.10
image

humbertogontijo:
I don't want to spend time trying to solve this problem. I just removed python 10 tests from ci
Using only python 11 should be fine

PR #77, opened 2023-06-17
feat: adding command cache

humbertogontijo:
This is my proposal for command cache. Fetch a single time and evict if set or change is called(Only for the methods in CacheableAttribute enum)

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -18,7 +18,7 @@ from typing import Any, Callable, Coroutine, Optional, Type, TypeVar, final
 import aiohttp
 
 from .code_mappings import RoborockDockTypeCode
-from .command_cache import CacheableAttribute, CommandType, parse_method
+from .command_cache import AttributeCache, CacheableAttribute, CommandType, parse_method
 from .containers import (
     ChildLockStatus,
     CleanRecord,
@@ -104,7 +104,9 @@ class RoborockClient:
         self._last_disconnection = self.time_func()
         self.keep_alive = KEEPALIVE
         self._diagnostic_data: dict[str, dict[str, Any]] = {}
-        self.cache: dict[CacheableAttribute, Any] = {}
+        self.cache: dict[CacheableAttribute, AttributeCache] = {
+            cacheable_attribute: AttributeCache(cacheable_attribute) for cacheable_attribute in CacheableAttribute
+        }
 
     def __del__(self) -> None:
         self.sync_disconnect()
@@ -256,7 +258,7 @@ class RoborockClient:
     ) -> RT:
         parsed_method = parse_method(method)
         if parsed_method is not None and parsed_method.type == CommandType.GET:
-            cache = self.cache.get(parsed_method.attribute)
+            cache = self.cache[parsed_method.attribute].value
             if cache is not None:
                 if return_type:
                     return return_type.from_dict(cache)
@@ -266,9 +268,9 @@ class RoborockClient:
 
         if parsed_method is not None:
             if parsed_method.type == CommandType.SET:
-                self.cache[parsed_method.attribute] = None
+                self.cache[parsed_method.attribute].evict()
             elif parsed_method.type == CommandType.GET:
-                self.cache[parsed_method.attribute] = response
+                self.cache[parsed_method.attribute].load(response)
         if return_type:
             return return_type.from_dict(response)
         return response
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -13,11 +13,12 @@ import secrets
 import struct
 import time
 from random import randint
-from typing import Any, Callable, Coroutine, Optional, Type, TypeVar
+from typing import Any, Callable, Coroutine, Optional, Type, TypeVar, final
 
 import aiohttp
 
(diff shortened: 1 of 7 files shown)


humbertogontijo:
Every set command could potentially be followed by a get command too. So we don't need async calls for values

roborock/command_cache.py
```diff
@@ -0,0 +1,54 @@
...
+    type: CommandType
+    attribute: CacheableAttribute
+
+
+def parse_method(method: str):
+    if method is not None:
+        attribute = method.lower()
+        if attribute.startswith(GET_PREFIX):
+            attribute = attribute.removeprefix(GET_PREFIX)
+        elif attribute.startswith(SET_PREFIX):
```
Lash-L, line 77: I would say close is maybe also a set prefix.

Lash-L (approved):
I like this a lot. 

Just one tiny piece of feedback.

roborock/command_cache.py
humbertogontijo: Oh, that's true

Lash-L:
I also think potentially there should be a time aspect. Items in the cache should expire after a set period of time. That way if a value changes in the app or by itself, it will update without the user needing to do a set command.

Also may be nice to have a method to get from attributes and if the value doesn't exist in attributes or if it does but it's None, then it automatically calls the get.

Just two thoughts, neither are strictly needed though

humbertogontijo:
I agree with your first point.
The second one I previously though about it, but it would require the getter to be async

Lash-L:
Potentially could have an async version that calls the sync version and if the result is None it does what we need. 

That way you can have a way to always get a value if you need it. I.e when you are doing the setup, you call the async one to make sure you get a value and can determine if it is supported, then you know there is a value in it and you can call the sync version every other time you need it

humbertogontijo:
> Potentially could have an async version that calls the sync version and if the result is None it does what we need.
> 
> That way you can have a way to always get a value if you need it. I.e when you are doing the setup, you call the async one to make sure you get a value and can determine if it is supported, then you know there is a value in it and you can call the sync version every other time you need it

Couldn't think of a way without creating a circular reference between api and the cache for this. Probably there is no good way of doing it

roborock/api.py
```diff
@@ -240,38 +242,51 @@ def _get_payload(
...
+
+        response = await self._send_command(method, params)
+
+        if parsed_method is not None:
+            if parsed_method.type == CommandType.SET:
+                self.cache[parsed_method.attribute].evict()
+            elif parsed_method.type == CommandType.GET:
+                self.cache[parsed_method.attribute].load(response)
+        if return_type:
+            return return_type.from_dict(response)
```
Lash-L, line 275: Can't these be combined? We are doing the same if a few times here.

Lash-L (approved):

roborock/api.py
Lash-L: Actually looking at it again, this is probably fine, I see why you did it like this.

Lash-L:
LGTM

seidtgeist:
may I ask why there's an attribute cache in the first place? i found this super confusing:

1. LOAD_MULTI_MAP 0 → ok
2. GET_STATUS map_status → 3 (ok since 3 >> 2 = 0) 
3. LOAD_MULTI_MAP 1 → ok
4. GET_STATUS map_status → 3 (NOT ok, should be 7 since 7 >> 2 = 1) 
5. LOAD_MULTI_MAP 2 → ok
6. GET_STATUS map_status → 3 (still NOT ok, should be 11 since 11 >> 2 = 2) 

if i call `await mqtt_client.cache['status'].refresh_value()` in-between `LOAD_MULTI_MAP` and `GET_STATUS`, the cache gets refreshed and the values being returned are correct.

am i doing something wrong, or is the cache outdated after a change (which IMO it shouldn't be)?

humbertogontijo:
> may I ask why there's an attribute cache in the first place? i found this super confusing:
> 
> 1. LOAD_MULTI_MAP 0 → ok
> 2. GET_STATUS map_status → 3 (ok since 3 >> 2 = 0)
> 3. LOAD_MULTI_MAP 1 → ok
> 4. GET_STATUS map_status → 3 (NOT ok, should be 7 since 7 >> 2 = 1)
> 5. LOAD_MULTI_MAP 2 → ok
> 6. GET_STATUS map_status → 3 (still NOT ok, should be 11 since 11 >> 2 = 2)
> 
> if i call `await mqtt_client.cache['status'].refresh_value()` in-between `LOAD_MULTI_MAP` and `GET_STATUS`, the cache gets refreshed and the values being returned are correct.
> 
> am i doing something wrong, or is the cache outdated after a change (which IMO it shouldn't be)?

To avoid Roborock restrictions when frequently fetching new data from their servers

seidtgeist:
Ouff, thank you. i can’t believe MQTT is tunneled via their server…

seidtgeist:
@humbertogontijo Is a cache also required when using the local client with a `device_info.host` set?

humbertogontijo:
Randomly commands like get_status fails some times. For local client cache is there just to keep last known state

PR #79, opened 2023-06-20
chore: update pyproject

Lash-L:
a few things I saw suggested to try to fix our poetry issues


Lash-L:
I also saw suggested to poetry install 3 times haha
```poetry install -vvv --no-root || poetry install -vvv --no-root || poetry install -vvv --no-root...```

humbertogontijo (approved):

PR #83, opened 2023-06-25
feat: adding status and consumable listeners

humbertogontijo:
The mqtt client have messages for some attribute changes that are instantly. I'm creating some use for it here

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -55,7 +55,13 @@ from .exceptions import (
 )
 from .protocol import Utils
 from .roborock_future import RoborockFuture
-from .roborock_message import RoborockDataProtocol, RoborockMessage, RoborockMessageProtocol
+from .roborock_message import (
+    ROBOROCK_DATA_CONSUMABLE_PROTOCOL,
+    ROBOROCK_DATA_STATUS_PROTOCOL,
+    RoborockDataProtocol,
+    RoborockMessage,
+    RoborockMessageProtocol,
+)
 from .roborock_typing import DeviceProp, DockSummary, RoborockCommand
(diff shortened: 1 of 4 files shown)


humbertogontijo:
This also fixes https://github.com/humbertogontijo/homeassistant-roborock/issues/395

Lash-L:
Sorry I missed this and I've been out of commission lately. Been rather busy.

As an aside, in the component, instead of using self.api.cache.get, we should use

```    def get_cache(self, attribute: CacheableAttribute) -> AttributeCache:
        try:
            result = self._api.cache.get(attribute)
        except RoborockException as err:
            raise HomeAssistantError(f"Error while getting {attribute.name}") from err
        return result
```

To allow for HomeAssistant to recover gracefully on a failed error

I plan to add it to the custom side once I get done with the core side unless you have a chance to first

humbertogontijo:
> Sorry I missed this and I've been out of commission lately. Been rather busy.
> 
> As an aside, in the component, instead of using self.api.cache.get, we should use
> 
> ```
>         try:
>             result = self._api.cache.get(attribute)
>         except RoborockException as err:
>             raise HomeAssistantError(f"Error while getting {attribute.name}") from err
>         return result
> ```
> 
> To allow for HomeAssistant to recover gracefully on a failed error
> 
> I plan to add it to the custom side once I get done with the core side unless you have a chance to first

Actually there shouldn't be any issue on this get. Only when calling api.cache[attribute].async_value()

Lash-L:
That is a good point. I started working on it not realizing at first that the get was just a dict get and then I didn't think about it. Good call.

PR #87, opened 2023-07-09
fix: listeners getting protocol data before it exists.

Lash-L:
https://github.com/home-assistant/core/issues/96082 relates to this issue

Essentially, what seems to happen is that the listeners are getting data before get_status is ever called, so .value has never been set.


Lash-L:
@humbertogontijo when you have a chance can you review this? Want to get this included in a HA patch

PR #91, opened 2023-07-17
fix: remove depreciated commands

Lash-L:


Lash-L:
@humbertogontijo not sure if this will actually fix it, but I bumped semantic release to its newest version and it seems to fail. Either this needs to be merged + some changes (maybe) or the pr should be reverted

humbertogontijo:
Removing the pypi token will prevent it from releasing new versions, no? What extra changes are needed? Can't those be in this same PR?

Lash-L:
I think they need a different key now. I’m away from a computer for a week, but I can look once I’m back

I haven’t been able to dove into the docs.

Lash-L:
Perhaps it just needs to be reverted for now

PR #92, opened 2023-07-17
fix: add missing s8 pro mop code and q revo dock

Lash-L:


humbertogontijo (approved):

Lash-L:
Bumping poetry to try to fix the annoying issues with the dependency installs

humbertogontijo (approved):

PR #94, opened 2023-07-20
chore: add some initial documentation

shred86:
Initial commit to add some more information to the documentation. I'll continue to add more but wanted to make sure I'm doing this correctly as I haven't worked with Sphinx before. Is updating the .rst files like this correct? Additionally, I wanted to make sure I'm following any standard conventions you guys want to use.

Some discussion items:
- I added an extra line called "Return example:". I think it would be helpful to have an example of the response where applicable. 
- Do we want a line for "Source: Roborock S7 MaxV Ultra" so we know where the information is from? I'm guessing different vacuums will return different keys and values.
- The other thought I had was for each return key description, perhaps we can add the data type (str, int, etc.)?


humbertogontijo:
Looks good to me. But lets confirm with @Lash-L

Lash-L:
I'm currently without much service until Saturday or so probably. Give me until then and I'll give it a pass. The changes won't even load for me rn

docs/source/api_commands.rst
```diff
@@ -15,10 +15,37 @@ Parameters: None
 app_get_dryer_setting
 ---------------------
 
-Description:
+Description: Get docker dryer settings.
```
Lash-L: Docker -> dock

docs/source/api_commands.rst
```diff
@@ -15,10 +15,37 @@ Parameters: None
...
+    
+        cliff_on:
+
+        cliff_off:
+
+        count:
+
+Return example::
+
+    {'status': 1, 'on': {'cliff_on': 1, 'cliff_off': 1, 'count': 10, 'dry_time': 7200}, 'off': {'cliff_on': 2, 'cliff_off': 1, 'count': 10}}
```
Lash-L, line 46: Put this in some md code comments

docs/source/api_commands.rst
```diff
@@ -15,10 +15,37 @@ Parameters: None
...
 
+    off:
+    
+        cliff_on:
+
+        cliff_off:
+
+        count:
+
+Return example::
```
Lash-L, line 44: Extra :

docs/source/api_commands.rst
```diff
@@ -15,10 +15,37 @@ Parameters: None
...
+
+    {'status': 1, 'on': {'cliff_on': 1, 'cliff_off': 1, 'count': 10, 'dry_time': 7200}, 'off': {'cliff_on': 2, 'cliff_off': 1, 'count': 10}}
+
+Source: Roborock S7 MaxV Ultra
```
Lash-L, line 48: Rather than a source, I think a supported would be helpful. A generic table for each one where each vacuum states if it is supported/ not supported/ unknown. And then if they ever return different results, we could add that when it occurs rather than always having, just my opinion

docs/source/api_commands.rst
```diff
@@ -177,9 +204,13 @@ camera_status
 
 Get: get_camera_status
```
Lash-L, line 205: For one like this where we don't really know the why the result means, we should probably add some kind of indicator. As just putting the name of the command in the description isn't too helpful.

Would be good so other people know what to look at for helping to reverse engineer

docs/source/api_commands.rst
```diff
@@ -194,10 +225,20 @@ carpet_clean_mode
 
 Get: get_carpet_clean_mode
```
Lash-L, line 226: Would be nice if for all carpet clean modes we could say what each number means

docs/source/api_commands.rst
```diff
@@ -194,10 +225,20 @@ carpet_clean_mode
...
 
 Parameters:
 
+Returns:
+
+    carpet_clean_mode: Enumeration for carpet clean mode.
+
+Return example::
+
+    {'carpet_clean_mode': 3}
```
Lash-L, line 238: For all returns I think it would be nice to have them in code blocks

docs/source/api_commands.rst
```diff
@@ -527,9 +624,27 @@ Parameters:
 get_network_info
```
Lash-L, line 628: I think what you have for this one is perfect outside of needing code blocks on the return. But this is a good basis for the rest

docs/source/api_commands.rst
```diff
@@ -559,9 +674,19 @@ Parameters:
...
-Parameters:
+Parameters: None
+
+Returns:
+
+    serial_number: Serial number of the vacuum.
+
+Return example::
+
+    {'serial_number': 'B16EVD12345678'}
```
Lash-L, line 691: If this is yours, I would change it and make it generic

Lash-L (changes requested):
I think this is great, thanks for doing this. Just a few changes. I didn't test compiling and viewing but I assume it is good

docs/source/api_commands.rst
shred86: This is to create a [literal block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) for the actual JSON response below.

docs/source/api_commands.rst
shred86: I tried looking up how to format a code block and the only thing I could really find is using a [literal block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks). Is there an example of an md code comment? I wasn't sure if normal markdown worked for .rst

docs/source/api_commands.rst
shred86: Agreed. I just wasn't sure what the return values actually mean so I'm not quite sure how to describe it. I just added what I could find for now with the plan to update it, but if it's not useful, I can just delete it for now.

docs/source/api_commands.rst
shred86: Agreed. I'm planning on adding more as I learn what each command does. Hoping other people will be able to help out as well, but just adding as I go along.

docs/source/api_commands.rst
shred86: Yeah I did modify it, but appreciate the back up! Wouldn't be the first time I forgot to obfuscate something. 😄

shred86:
This is minor but I noticed some commands listed on the API documentation have a title, then the actual get and set commands below. For consistency, would it be better if we just separated each "entry" as the actual API command? For example, the [dnd_timer](https://python-roborock.readthedocs.io/en/latest/api_commands.html#dnd-timer) is the essentially the title but the actual API commands are `get_dnd_timer` and `set_dnd_timer`. I think it would be preferable to just separate them as two separate entries, `get_dnd_timer` and `set_dnd_timer` without nesting them under a title that isn't actually an API command.

docs/source/api_commands.rst
Lash-L: Ah - you're right that is the right way to do it. I'm still learning sphinx as well

Lash-L:
Make sure to run pre-commit before you commit to fix linting.

Lash-L:
Run pre commit then I'm good with merging this in

Lash-L:
@shred86 Can you fix the linting?

shred86:
> @shred86 Can you fix the linting?

Sorry I haven't been able to work on this more - just too busy with work lately. Just ran pre-commit and the linting issue should be fixed.

Lash-L (approved):

PR #100, opened 2023-07-31
feat: add device name to logs

Lash-L:
image
Makes logs like this so that we can more easily identify what logsare coming from what vacuum when there are multiple devices on an account.


humbertogontijo (approved):
I wonder if using the device id is worth, cause the device names could be the same.
Anyway, this will help with the existing issues

Lash-L:
I thought the same thing, but the device ids are so long that I thought it would be much easier to understand if it was with device name.

The user can always change their vacuums name if needed

PR #103, opened 2023-08-05
fix: resolve unawaited task errors on connect/disconnect

the-ress:
The mqtt client now logs exceptions like `Error doing job: Task exception was never retrieved [...] RoborockTimeout: id=0 Timeout after 4 seconds` when it's already connected.

This PR fixes that by only calling `_async_response` when we actually need to connect. (Including the same-ish fix for disconnect.)

Full log entry:
```
2023-08-05 13:32:15.078 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved:   File "/home/vscode/.local/bin/hass", line 8, in <module>
    sys.exit(main())
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/__main__.py", line 214, in main
    exit_code = runner.run(runtime_conf)
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/runner.py", line 179, in run
    return loop.run_until_complete(setup_and_run_hass(runtime_config))
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 640, in run_until_complete
    self.run_forever()
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 607, in run_forever
    self._run_once()
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 1914, in _run_once
    handle._run()
  File "/usr/local/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/workspaces/homeassistant-roborock/roborock/api.py", line 428, in get_consumable
    return Consumable.from_dict(await self.cache[CacheableAttribute.consumable].async_value())
  File "/workspaces/homeassistant-roborock/roborock/api.py", line 128, in async_value
    return await self.task.reset()
  File "/workspaces/homeassistant-roborock/roborock/util.py", line 101, in reset
    return await self._run_task()
  File "/workspaces/homeassistant-roborock/roborock/util.py", line 86, in _run_task
    response = await self.callback()
  File "/workspaces/homeassistant-roborock/roborock/api.py", line 122, in _async_value
    self._value = await self.api._send_command(self.attribute.get_command)
  File "/workspaces/homeassistant-roborock/roborock/cloud_api.py", line 195, in _send_command
    return await self.send_message(roborock_message)
  File "/workspaces/homeassistant-roborock/roborock/cloud_api.py", line 158, in send_message
    await self.validate_connection()
[...]

PR #104, opened 2023-08-06
test: fix sync_connect test

the-ress:
Fixes test warning after #103 (`coroutine 'RoborockClient._wait_response' was never awaited`)


Lash-L (approved):
LGTM

PR #105, opened 2023-08-06
fix: refactor cache and call get_status after changing mop mode

the-ress:
Updates get_status cache after some commands that change it - because the vacuum doesn't seem to send mop mode change notifications (at least S7 MaxV doesn't).


roborock/api.py
```diff
@@ -171,8 +174,7 @@ def __init__(self, endpoint: str, device_info: DeviceData) -> None:
         cache = device_cache.get(device_info.device.duid)
         if not cache:
             cache = {
-                cacheable_attribute: AttributeCache(attr, self)
-                for cacheable_attribute, attr in create_cache_map().items()
+                cacheable_attribute: AttributeCache(attr, self) for cacheable_attribute, attr in get_cache_map().items()
```
humbertogontijo, line 177: Correct me if I'm wrong, but get_cache_map is always returning the same object. This will affect multiple devices

the-ress: The object returned from `get_cache_map` contains just definitions that aren't specific to any device (unless I'm missing something). The actual per-device cache gets created from it here in `RoborockClient`.

humbertogontijo: Oh, thats true. This was initially created dynamically, getting available attributes for each device. But since this is a set of predefined attributes now, it should be fine

humbertogontijo (approved):

PR #107, opened 2023-08-07
fix: custom for mopping and custom for S6 sucking

fuatakgun:
I have setup my roborock as custom for both sucking and mopping. I see that, S6 is missing custom sucking, added it. And S6 has value of 204 for custom mopping in my case.

### DIFF
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -158,11 +158,12 @@ class RoborockFanSpeedS7MaxV(RoborockFanPowerCode):
 
 
 class RoborockFanSpeedS6Pure(RoborockFanPowerCode):
-    gentle = 105
     quiet = 101
     balanced = 102
     turbo = 103
     max = 104
+    gentle = 105
+    custom = 106
 
 
 class RoborockFanSpeedQ7Max(RoborockFanPowerCode):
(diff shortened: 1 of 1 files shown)


roborock/code_mappings.py
```diff
@@ -214,7 +215,7 @@ class RoborockMopIntensityV2(RoborockMopIntensityCode):
     low = 201
     medium = 202
     high = 203
-    custom = 207
+    custom = 204
```
Lash-L, line 218: Can you make a new enum like we do above for the s7 but for your device instead of modifying this one

Lash-L (changes requested):

humbertogontijo:
@fuatakgun is your device in latest firmware?

fuatakgun:
> @fuatakgun is your device in latest firmware?

I am on 02.16.62 and I do not have any more updates.

roborock/code_mappings.py
fuatakgun: Now, it requires a much more proper testing locally, I will do tonight :)

Lash-L:
Hi @fuatakgun is this still a task you plan to tackle?

Lash-L:
On second thought - I will just knock it out

PR #118, opened 2023-09-09
chore: add pyupgrade to ruff

Lash-L:
pyupgrade changes all of our optionals to be | None and a few other changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ build_command = "pip install poetry && poetry build"
 [tool.ruff]
 ignore = ["F403", "E741"]
 line-length = 120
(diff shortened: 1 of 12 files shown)


humbertogontijo:
Oh, is this the new preferred approach? I mean, it makes more sense, but I always had complains from mypy that made me go this way haha

Lash-L:
Yeah this is the preferred approach now, but mypy probably contained before we upgraded our python version as this wasn't supported in 3.8

humbertogontijo:
Oh, got it. Thank you

humbertogontijo (approved):
Looks good

PR #121, opened 2023-09-14
fix: status reworking

Lash-L:
Instead of getting status each time we need it, we get it once so it is always available.

As well, we Shouldn't set the cache based off of listener if nothing exist, this was causing a few issues when I was testing robustness of being able to set up in core when the device is offline and then comes online


humbertogontijo (approved):
LGTM

PR #130, opened 2023-10-18
feat(code_mappings): add error n53 cleaning tank full or blocked

Jezza34000:
Feat : Add error n°53 => Cleaning tank is full or blocked


Lash-L:
When you make a PR to bump the dependency, please add this code to all languages in english

PR #131, opened 2023-10-20
feat: add custom state for charging and will continue

Lash-L:
This way a user can tell if their vacuum is just charging or if it is charging and planning to continue cleaning

### DIFF
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -67,6 +67,7 @@ class RoborockStateCode(RoborockEnum):
     going_to_wash_the_mop = 26  # on a46
     charging_complete = 100
     device_offline = 101
+    charging_will_continue_cleaning = 9999  # Manually added
 
 
 class RoborockErrorCode(RoborockEnum):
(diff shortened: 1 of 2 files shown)


humbertogontijo:
Shouldn't this be done on automations and not in the library itself?

Lash-L:
My thought was that this is generic enough of information that it should always be available to the user rather than the user having to specify it in an automation- and I'm surprised it is not a base state to begin with. But I'm open to not doing it.

It is kind of weird though since it takes a class that previously only specified data from the api directly and makes it semi-manual.

I'm open to going either way - was curious on your thoughts

humbertogontijo:
How it goes in the app?

Lash-L:
> How it goes in the app?

iirc - it says the device is charging and then the bottom part says 'stop' and 'dock' like it typically does when you can cancel a clean

humbertogontijo:
I think it will be better not create sub statuses. Using in_cleaning as an entity might be the way to go

PR #132, opened 2023-10-27
fix: Add S5Max Mop Code 207

shbatm:
Add Mop Code 207 to eliminate error:

```
Logger: roborock.code_mappings
Source: runner.py:186
First occurred: 5:41:40 PM (1 occurrences)
Last logged: 5:41:40 PM

Missing RoborockMopIntensityS5Max code: 207 - defaulting to 200
```

### DIFF
--- a/roborock/code_mappings.py
+++ b/roborock/code_mappings.py
@@ -246,6 +246,7 @@ class RoborockMopIntensityS5Max(RoborockMopIntensityCode):
     medium = 202
     high = 203
     custom = 204
+    custom_water_flow = 207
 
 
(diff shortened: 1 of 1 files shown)


Lash-L:
Just to confirm @shbatm you have a s5 max, right? We were previously using the s5 max intensity for a few other modes and on older version of the package that could be confusing, so I just want to double check bckfore I merge. But otherwise LGTM - you can ignore commit lint.

shbatm:
Correct, this is from my S5 Max.

Lash-L (approved):

PR #134, opened 2023-10-29
fix: typing and versioning

Lash-L:
I think adding poetry lock may help with our dacite issues from a bit of what I am reading

I also made a few changes for variables that we know we will have/ we rely on having to make them not None


humbertogontijo (approved):
Lets give it a try

PR #137, opened 2023-11-01
Fix mypy complaints

Lash-L:
With py.typed enabled there are a lot of mypy complaints on the component side using the library.

This makes a few changes such as making some variables that we require/ know we will have into empty versions of themselves before we get them, i.e. Status.

Exporting some commonly accessed values like the name of the dock error

In the old system it was 

props.status.dock_error.name But status could be none and dock_error could be None and that was two additional None checks that had to be done in the component side - this simplifies things.


humbertogontijo:
I'm ok with the new field within Status class

Something else is wrong with typing. RoborockBase.from_dict can return None

Lash-L:
> I'm ok with the new fields within Status class
> 
> 
> 
> Something else is wrong with typing. RoborockBase.from_dict can return None

Hmm

Should it raise when from dict is passed a non dict and returns none? Or create an empty version of the class.

I see what you mean now. What's you opinion on best typing there?

humbertogontijo:
Raising an exception is a good way to type it. But at the same time it could have a big impact on core and hacs implementation

Lash-L:
Hows that? A few guaranteed types - others optional None

roborock/api.py
```diff
@@ -413,9 +410,12 @@ async def send_command(
         return response
 
     async def get_status(self) -> Status:
-        return self._status_type.from_dict(await self.cache[CacheableAttribute.status].async_value())
+        data = self._status_type.from_dict(await self.cache[CacheableAttribute.status].async_value())
+        if data is None:
+            return Status()
```
humbertogontijo: self._status_type() maybe?

roborock/api.py
```diff
@@ -293,10 +293,7 @@ def on_message_received(self, messages: list[RoborockMessage]) -> None:
                     payload = data.payload[0:24]
                     [endpoint, _, request_id, _] = struct.unpack("<8s8sH6s", payload)
                     if endpoint.decode().startswith(self._endpoint):
-                        try:
-                            decrypted = Utils.decrypt_cbc(data.payload[24:], self._nonce)
-                        except ValueError as err:
-                            raise RoborockException("Failed to decode %s for %s", data.payload, data.protocol) from err
+                        decrypted = Utils.decrypt_cbc(data.payload[24:], self._nonce)
```
humbertogontijo: Why raise ValueError?

Lash-L: That's what the underlying package raises

Lash-L: Although 0 i'm not sure why i deleted it. thta was a mistake i will add back

roborock/api.py
Lash-L: good call

Lash-L:
I'll fix lint commit in merge

humbertogontijo (approved):

PR #140, opened 2023-11-06
fix: update listeners

Lash-L:
My thought here was that it is global, but we would be able to just call functions that are affected by this value change. i.e. just update the error sensor when we get an error instead of updating all of our entitites.

Thoughts?


roborock/code_mappings.py
```diff
@@ -249,6 +249,17 @@ class RoborockMopIntensityS5Max(RoborockMopIntensityCode):
     custom_water_flow = 207
 
 
+class RoborockMopIntensityS6MaxV(RoborockMopIntensityCode):
```
Lash-L, line 252: Also added this from issue on github

humbertogontijo (approved):
Looking good

Lash-L:
Made a few changes to get it to work well, good to go as long as you agree with it

roborock/api.py
```diff
@@ -184,10 +193,11 @@ def __init__(self, endpoint: str, device_info: DeviceData, queue_timeout: int =
         self.cache: dict[CacheableAttribute, AttributeCache] = {
             cacheable_attribute: AttributeCache(attr, self) for cacheable_attribute, attr in get_cache_map().items()
         }
-        self._listeners: list[Callable[[str, CacheableAttribute, RoborockBase], None]] = []
         self.is_available: bool = True
         self.queue_timeout = queue_timeout
         self._status_type: type[Status] = ModelStatus.get(self.device_info.model, S7MaxVStatus)
+        if device_info.device.duid not in self._listeners:
+            self._listeners[device_info.device.duid] = ListenerModel({}, self.cache)
```
humbertogontijo, line 200: Although this is global now we could still have a variable at class level. Just for simple access to this ListenerModel

roborock/api.py
```diff
@@ -169,7 +170,15 @@ async def refresh_value(self):
         await self._async_value()
 
 
+@dataclasses.dataclass
+class ListenerModel:
+    protocol_handles: dict[RoborockDataProtocol, list[Callable[[Status | Consumable], None]]]
```
humbertogontijo: typo?

Lash-L: Good catch

roborock/api.py
Lash-L: good point - changed

humbertogontijo (approved):
Good to go

Lash-L:
Tested having a missing key - and I realized the _missing_ function of RoborockEnum was going to automatically make it not raise a valueerror. So I overroad it for this enum. Now it should be good haah

humbertogontijo (approved):

PR #145, opened 2023-11-14
Initial view of extending the documentation 

chrisn-au:
Hi,

I have just started a go at extending the documentation based on the framework you have done and a couple of updates by others. If you are ok with my changes to date - style rather than content as that may need to be modified, I will complete it for as much as i can without looking at code this week and then dive into the code for the more complex calls next week.

BTW An awesome project.

Feel free to reject the pull request if content is to early but would appreciate comments on style

Chris


humbertogontijo:
Looking good. Except for the "Comment :...". I think this should actually be a comment and not get into the final document

docs/source/api_commands.rst
```diff
@@ -357,19 +387,48 @@ Returns:
...
+
+Returns:
+
+    integer value of the current custom mode
+
+Return example::
+
+    102
 
+Comment : Not clear what a custom mode is = will explore
```
Lash-L, line 420: custom mode is used for fan speed

chrisn-au: thx

chrisn-au:
@Lash-L @humbertogontijo 

I continue to make changes can be tracked [here](https://github.com/chrisn-au/python-roborock/tree/documentation-update-s8-Ultra)

 I will not create a pull request before it is in a mode solid state

I am getting to the point of needing to understand parameters sent for a command. I am guessing the only sensible way to do this is to observe the conversation for the app. Is there a sensible way to do this. Can I subscribe to the matt queue ?

humbertogontijo:
@chrisn-au each Roborock's mqtt connection have it's own topic(App have one topic with the command responses and another connection won't see them). Meaning that the app and your connection have different messages. So, the easiest way to get information from the app is sniffing the local connection https://github.com/humbertogontijo/python-roborock/blob/59a59dbb811868541b033b4f7e17d5f098395e78/roborock/cli.py#L149

Lash-L:
If you'd like as an easier first approach. You should able to turn on debug logging and get a better idea of the commands that are used within the home assistant integration. Should give you a good understanding of most commands

chrisn-au:
Thanks but keen to have a got a Wireshark. After some fun with MAC os and rvitcl have got it monitoring. I am not python native (go and node.js) and trying to figure out the response

Am getting 

Received response: []
Received request: []

but no content ?? Does this make sense ?

chrisn-au:
Pull request closed as it is not current

PR #158, opened 2023-12-11
fix: add information from product api

Lash-L:
The product api had some good information including new status codes, some status codes for the dyad series, model numbers, etc.

### DIFF
--- a/roborock/const.py
+++ b/roborock/const.py
@@ -18,11 +18,14 @@ ROBOROCK_T7 = "roborock.vacuum.a11"  # cn s7
 ROBOROCK_T7S = "roborock.vacuum.a14"
 ROBOROCK_T7SPLUS = "roborock.vacuum.a23"
 ROBOROCK_S7_MAXV = "roborock.vacuum.a27"
+ROBOROCK_S7_MAXV_ULTRA = "roborock.vacuum.a65"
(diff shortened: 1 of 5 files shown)


roborock/const.py
```diff
@@ -18,11 +18,14 @@
 ROBOROCK_T7S = "roborock.vacuum.a14"
 ROBOROCK_T7SPLUS = "roborock.vacuum.a23"
 ROBOROCK_S7_MAXV = "roborock.vacuum.a27"
+ROBOROCK_S7_MAXV_ULTRA = "roborock.vacuum.a65"
 ROBOROCK_S7_PRO_ULTRA = "roborock.vacuum.a62"
 ROBOROCK_Q5 = "roborock.vacuum.a34"
-ROBOROCK_Q7 = "roborock.vacuum.a37"  # CHECK THIS
```
humbertogontijo, line 23: Should we remove it? Maybe just rename it to unknown

Lash-L: It was moved down below to be the correct model number. 37 isn't actually a real number - it was really just a guess.

humbertogontijo: Ok

humbertogontijo (approved):

PR #194, opened 2024-03-01
feat: add v1 api

Lash-L:
This is the next big push towards getting support for A01 devices added in. It may change a bit from here but this is the foundation. Essentially, RoborockClient, RoborockLocalClient, and RoborockMqttClient should hold generic actions that should be true for all Roborock devices. Local + Mqtt should contain their respective generic things, then V1 and A01 should hold any specific implementation that is unique to that protocol.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -930,28 +930,28 @@ files = [
 
 [[package]]
 name = "ruff"
-version = "0.2.2"
+version = "0.3.0"
 description = "An extremely fast Python linter and code formatter, written in Rust."
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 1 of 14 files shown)


humbertogontijo:
I loved the way you made this PR looks simpler than it actually is
I think it looks good, but let me just digest these changes a bit
One thing that I can see is the naming with v1 in the middle. I think it should be at the end (file name and class name)

Lash-L:
> I loved the way you made this PR looks simpler than it actually is I think it looks good, but let me just digest these changes a bit One thing that I can see is the naming with v1 in the middle. I think it should be at the end (file name and class name)

I tried to keep it as basic as possible, but it is a large change so that was a bit tricky. I also didn't want to complicate things more by adding A01 in yet.

I've gone ahead and changed the location of the V1 in the naming

humbertogontijo (approved):
Lets move on with this

PR #195, opened 2024-03-03
feat: add nonce to diagnostic data

Lash-L:
I am trying to debug an issue with some users map, but their data is useless to me without the nonce


Lash-L:
@humbertogontijo When you have a chance can you check this one out? would like to get this in sooner than later so I can try to debug these users problems

humbertogontijo (approved):

PR #198, opened 2024-04-07
chore: move more things around in version 1 api

Lash-L:
Tested it all and working - figured i'd split this up into another PR before I added the A01 stuff.


Lash-L:
Once this is in - I think I finally have A01 stuff finalized and I will make a PR for that @humbertogontijo

humbertogontijo (approved):
LGTM

PR #199, opened 2024-04-08
refactor: add A01

Lash-L:


commitlint.config.js
```diff
@@ -1,4 +1,21 @@
+import {
```
humbertogontijo: A check is failing because of this import. Everything else looks good

Lash-L: Looks like it should be passing now.

humbertogontijo (dismissed):
Looking good

Lash-L:
Okay I finally figured out how to get the major release to work, so I just need another approval and I can fix it. You can ignore commitlint as I will squash commits anyways.

humbertogontijo (approved):

Lash-L:
coudl be both things
image

Lash-L:
I removed the old release and tag but I think for semantic release to do it all you need to remove the last 1.0.0 commit. I'm unable to do that

humbertogontijo:
I deleted the last commit and its running again

Lash-L:
> I deleted the last commit and its running again

looks like it worked - thanks

PR #203, opened 2024-05-08
fix: set roommapping when it is only one room

Lash-L:
There is a bug where if only one room is on a map, it only sends a 1d array instead of a 2d array, and our code fails.


roborock/version_1_apis/roborock_client_v1.py
```diff
@@ -277,6 +277,8 @@ async def get_room_mapping(self) -> list[RoomMapping] | None:
         """Gets the mapping from segment id -> iot id. Only works on local api."""
         mapping: list = await self.send_command(RoborockCommand.GET_ROOM_MAPPING)
         if isinstance(mapping, list):
+            if not isinstance(mapping[0], list):
```
humbertogontijo: Would you mind adding some length checks? I looks like mapping variable needs to be at least 2 items long

Lash-L: done

humbertogontijo (approved):

PR #205, opened 2024-05-11
fix: add missing value "high = 203" to RoborockMopIntensityS8MaxVUltra

stefanroellin:
Solves the warning 

```Missing RoborockMopIntensityS8MaxVUltra code: 203 - defaulting to 200```

if command "get_status" is sent with return_type=S8MaxvUltraStatus


humbertogontijo (approved):
Looks good

PR #206, opened 2024-05-13
fix: handle weird clean record response

Lash-L:
Oddly enough - sometimes two records can apparently come back for one timestamp - while that makes no logical sense, this handles it.


humbertogontijo (dismissed):
This is weird, but I'm approving it anyway

Lash-L:
i've made it a bit more robust.

humbertogontijo (approved):

PR #210, opened 2024-05-26
fix: S8 maxv highest mop intensity setting

ethemcemozkan:
Newest roborock s8 maxv has one more mop intensity setting. I am open to changing its name

image


humbertogontijo:
Previously the mop intensity was jumping from high to max plus. But since you have only max in your app I can accept it

humbertogontijo (approved):

PR #213, opened 2024-05-30
fix: s8 maxv has a wash and fill dock

ethemcemozkan:
I was checking what else is missing for s8 maxv. This model has a dock that washes the mop and also fills the robot. So I think it should be included in this list.

Please let me know if I have to do anything else.


humbertogontijo (approved):
Yes, you're right

PR #220, opened 2024-06-29
feat: add q revo pro/p10 pro support

steffenweber:
I have a Q Revo Pro / P10 Pro which triggers the following warning:

`Missing RoborockDockTypeCode code: 9 - defaulting to 'unknown'`


humbertogontijo (approved):
Looks good

PR #227, opened 2024-10-28
feat: remove dacite

Lash-L:
Finally did this to end my dacite dependency suffering


humbertogontijo:
There's been a lot of version conflicts with dacite indeed

Issue #228, opened 2024-10-28
Asyncio improvements


Lash-L:
Place to keep track of all of the ideas @allenporter brought up in #222

- [ ] making breaking changes to drop all the sync APIs and focus on simpler async APIs.
- [ ] reduce the class hierarhcy so there are not multiple base classes involved, and use composition rather than inheritance.
- [ ] adding logging for each outgoing request seq id and received resopnse seq id could help show the issue happening in more detail.
- [x] add guards to not send a message with a request id that has already been allocated. One example of a potential bug is that the choosing of a request id can pick an id that already exists since its just picking random numbers.
- [ ] Add an explicit guard here when the future is already completed.
- [x] Add a guard in _async_response so that it doesn't overwrite an existing future if the request id is already in use
- [ ] There are cases where a response is received and it checks against the queue and pulls out the message, but then if it doesn't already exist it continues anyway. Why would this ever be allowed to happen? seems like a bug or malformed response so it also needs logging rather than being ignored since it could hide concurrency bugs
- [ ] Remove the multiple levels of futures happening between asyncio.ensure_future then a sync method _async_response which calls an async method _wait_response which directly awaits on a future. I don't think all these levels of indirection are needed
- [ ] Don't pass around tuples of (response, exception). Just use the future built in exceptions

allenporter:
I've made a few tiny changes of cleanup to get more familiar with the code base and understand it better.  There are now three changes I would like to focus on.  Let me know if you have any feedback before I get started.

## Reduce the class hierarchy

I've also noticed there are a lot of duplicate initializations happening of local variables which is really confusing. For example `RoborockClient` initializes `_waiting_queue` but so does `RoborockMqttClient`, etc. Also the loggers are set inconsistently, sometimes in the base class and sometimes in the child class, so log messages are inconsistent and hard to track down to the source class of which API is being used. Another example is `_send_msg_raw` is duplicated between the v1 MQTT client and the base class with duplicate implementations.

There are a few phases:

(1) Drop unnecessary hierarchy:

One other example of the type of simple change we can make is we have this graph:

`RoborockLocalClientV1 `-> (`RoborockLocalClient`,  `RoborockClientV1`)
`RoborockLocalClient` -> (`RoborockClient`, `asyncio.Protocol`)
`RoborockClientV1` -> `RoborockClient`

And all the code used by `RoborockLocalClient` is only used by `RoborockLocalClientV1` so we should drop it entirely.

[...]

PR #229, opened 2024-10-28
chore: Add some async improvements

Lash-L:
Relates to #228

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -23,7 +23,7 @@ from .roborock_message import (
     RoborockMessage,
 )
 from .roborock_typing import RoborockCommand
-from .util import RoborockLoggerAdapter, get_rand_int, get_running_loop_or_create_one
+from .util import RoborockLoggerAdapter, get_next_int, get_running_loop_or_create_one
 
 _LOGGER = logging.getLogger(__name__)
 KEEPALIVE = 60
@@ -114,7 +114,7 @@ class RoborockClient:
     ) -> Coroutine[Any, Any, tuple[Any, VacuumError | None]]:
         queue = RoborockFuture(protocol_id)
         if request_id in self._waiting_queue:
-            new_id = get_rand_int(10000, 32767)
+            new_id = get_next_int(10000, 32767)
             _LOGGER.warning(
                 f"Attempting to create a future with an existing request_id... New id is {new_id}. "
                 f"Code may not function properly."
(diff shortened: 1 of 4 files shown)


humbertogontijo:
I think it would be best to just keep it simple and create a counter to be used as the request_id. What you think?

humbertogontijo:
I don't know if it can conflict between cloud and local. But we can do something like top to bottom on cloud and bottom to top on local

Lash-L:
> I don't know if it can conflict between cloud and local. But we can do something like top to bottom on cloud and bottom to top on local

Sorry, I misspoke, I was specifically talking about commands run via the roborock app. i.e. I do something on my phone's app, it happens to be request id 34212 and that happens to be the next sequential id for HA. But it is probably rare enough that that edge case shouldn't even be really considered.

Lash-L:
Okay made the changes, what do you think about that?

humbertogontijo (approved):

roborock/api.py
```diff
@@ -113,6 +113,13 @@ def _async_response(
         self, request_id: int, protocol_id: int = 0
     ) -> Coroutine[Any, Any, tuple[Any, VacuumError | None]]:
         queue = RoborockFuture(protocol_id)
+        if request_id in self._waiting_queue:
+            new_id = get_next_int(10000, 32767)
+            _LOGGER.warning(
+                f"Attempting to create a future with an existing request_id... New id is {new_id}. "
+                f"Code may not function properly."
+            )
+            request_id = new_id
```
allenporter, line 122: I think this ends up writing the future with a different key, but does not change the request id chosen by the caller. I think this means the response will still be left hanging and/or associated with the wrong request?

roborock/roborock_message.py
```diff
@@ -155,9 +155,9 @@ class MessageRetry:
 class RoborockMessage:
     protocol: RoborockMessageProtocol
     payload: bytes | None = None
-    seq: int = randint(100000, 999999)
+    seq: int = get_next_int(100000, 999999)
```
allenporter, line 158: I believe these numbers are set once statically, and not once per message creation, which i think was the intent.  I think this is the equivalent of setting a constant for all created messages.

Lash-L: Looks like you may be correct? I modified what was existing prior without thinking about it much, I thought data classes worked a bit like init but obviously not. Good catch

PR #234, opened 2024-11-10
feat: add some new roborock codes and add custom command

Lash-L:


humbertogontijo:
Wouldn't be better to return all parsed data? Something like get_parsed_map

Lash-L:
> Wouldn't be better to return all parsed data? Something like get_parsed_map

@humbertogontijo 

Eh - maybe? But after talking with Piotr for the map card, he really just needs the calibration points. 

Calibration points are already in a dictionary and serializable. The entire parsed map is not, which I'm afraid could cause some problems with Vacuum: Send Command

I think I would prefer to leave it like this and if a use case pops up, I will add a second one for get parsed map and think through a solution

PR #236, opened 2024-11-12
fix: bump semantic release

Lash-L:
I have no idea if this will work tbh.

But we are so behind on the semantic release versioning it isn't working with the newer python versioning required for the map parser.

I took this config directly from their migration guide.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,9 @@ jobs:
     concurrency: release
     permissions:
       id-token: write
+    environment:
+      name: pypi
+      url: https://pypi.org/project/python-roborock/
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


humbertogontijo:
I have limited knowledge on GitHub actions and semantic release lib. But we can try it out

humbertogontijo (dismissed):

Lash-L:
Can you give me a invite to the pypi repo again - I didn't accept it in time I guess?

I believe we may have to specify a trusted publisher of this github action https://docs.pypi.org/trusted-publishers/

humbertogontijo:
Invite sent

Lash-L:
@humbertogontijo I should be able to test it now that's its setup just need a approval to be able to merge

humbertogontijo (approved):

PR #238, opened 2024-11-13
fix: semantic release versioning tag

Lash-L:
Okay I ran semantic release locally and it worked with this. Grabbed directly from their config


humbertogontijo (approved):

Lash-L:
Sorry, I had another commit getting pushed, it auto dismissed your review

PR #239, opened 2024-11-13
fix: set python version in ci

Lash-L:
Ci is still using the wrong version of python. I'm trying to explicitly set it here.


Lash-L:
I think just updating the python-semantic-rleease action to the latest version is good enough. But setting up python just to be safe anyways. Slows down release a little but I would like to stop iterating on this and just get releases out

humbertogontijo (approved):
Looks good

PR #241, opened 2024-11-14
fix: move github token to env for semantic release

Lash-L:


humbertogontijo (approved):

Lash-L:
@humbertogontijo 

Sorry this is being way more involved than I thought it would be

Could you try following these instructions

https://github.com/semantic-release/github/issues/175#issuecomment-757448750

If you make the GH_TOKEN, I could do the rest, or you could modify it directly yourself if you'd like, up to you. But I don't have permissions to do that.

humbertogontijo:
I think GH_TOKEN was already set. But I regenerated the token and set a new value to it

humbertogontijo:
There's also a PYPI_TOKEN env variable available

Lash-L:
Theoretically, pypi token shouldn't be needed. Right now it is just having failures pushing the new version commit.

PR #243, opened 2024-11-14
fix: update_gh_token

Lash-L:


humbertogontijo (approved):

Lash-L:
Getting rather annoyed with this whole process.

Could you check this for me? https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions Ensure it has full permissions

Lash-L:
> Getting rather annoyed with this whole process.
> 
> Could you check this for me? [docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions) Ensure it has full permissions

Ah I think I am seeing part of the problem now. Since I do not have full permissions, the github_token that it tries to use for me also does not have full permissions. I think if I just only use GH_TOKEN, it should work. Gonna make some changes and then a new pr

PR #247, opened 2024-11-20
fix: add config github actions

Lash-L:
Will try once without this as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
         run: |
           git config user.name "github-actions"
           git config user.email "actions@users.noreply.github.com"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -85,13 +85,16 @@ jobs:
     - name: Install Dependencies
       run: poetry install
       shell: bash
+    - name: Configure Git
+        run: |
+          git config user.name "github-actions"
+          git config user.email "actions@users.noreply.github.com"
+          git remote set-url origin https://${{ secrets.GH_TOKEN }}@github.com/<your-username>/<your-repo>.git
```
humbertogontijo: I think the url still have some place holders here

Lash-L: oops yes. Great catch

Lash-L: Fixed!

PR #249, opened 2024-11-20
fix: specify x-access-token

Lash-L:


humbertogontijo (approved):

Lash-L:
@humbertogontijo Can you try a manual push using the GH_TOKEN? Still getting auth issues

humbertogontijo:
I'm away from my pc. Can you create a simple PR? I guess if I merge it here it will be a commit from my user

Lash-L:
I was more interested in the specific token's auth - if you merge it I don't know if that will answer the question.

I'm going to get rid of semantic release I think. It has been too much of a pain. going to adopt another means of automatically doing it all.

humbertogontijo:
I was able to change readme file from mobile. It's running

humbertogontijo:
You now are a owner of the repository. I think you can do everything on a fork(even a release) before creating a PR. Maybe that will be an easier approach. 
And you can create a PR after a successfully release. What you think?

PR #250, opened 2024-11-20
fix: switch to new method of releases

Lash-L:
This one works by pushing the release out based on github releases. When you make a release there (it will auto draft releases), it pushes it. Doesn't do the fancy changelog and version updating, but I'm okay with that for now if I can stop hitting my head against the wall trying to get semantic release to work


humbertogontijo (approved):

humbertogontijo:
Approved. About the semantic release, I think it tries to work for every language and every exception. That makes it really complex. 
But I will agree with any solution you find best.
You can also work on a fork to make everything faster(not depending on my approvals while testing). I guess your fork would work like this repo since you are also a pypi owner

Lash-L:
Okay now that I have ownership maybe that will change my approach a bit. I will try testing a bit more on a separate branch.

Although - I don't see the settings page or anything else that seems to imply that I have ownership role - perhaps it takes a while?

Lash-L:
Gonna close this in favor of the fixes on the other

PR #251, opened 2024-11-20
fix: repair semantic release

Lash-L:
This works on the release branch. If I have the valid permissions, it should work here as well.

You could potentially add : "semantic-release" to overrides for branch protections. I do seem to have more permissions, but I can't see the settings page to do it myself. Hopefully it should work with my github_token

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,26 +75,11 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
-    - name: Set up Python 3.11
-      uses: actions/setup-python@v5
-      with:
-        python-version: "3.11"
-
-    - uses: snok/install-poetry@v1.4.1
-
-    - name: Install Dependencies
-      run: poetry install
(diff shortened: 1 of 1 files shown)


humbertogontijo (approved):

Lash-L:
@humbertogontijo Can you add "semantic-release" as a override for branch protection rules - is that possible?

You can see it committed here:
https://github.com/humbertogontijo/python-roborock/commits/release/

I think even if my GITHUB_TOKEN doesn't have the rights to commit, if you override it -it will work?

Alternatively - you can follow this for creating a token?
https://github.com/orgs/community/discussions/25305#discussioncomment-10728028

Problem is is that the release branch isn't protected, so I can try things there but it doesn't really help as it isn't true to what the main branch is like

humbertogontijo:
I cant find any override for github actions. It is only available for apps

humbertogontijo:
I added a DEPLOY_KEY secret to release environment. You can try it

PR #254, opened 2024-12-04
fix: change to deploy_key

Lash-L:


humbertogontijo (approved):

Lash-L:
@humbertogontijo What do you think about increasing my permissions so I can get this figured out. I think it would be a lot easier if I was able to create keys and test things.

You can change it back right after we get this fixed if you would like. This whole thing is just a bit aggravating and I would like it done.

PR #255, opened 2024-12-05
fix: add testing part and use ghtoken

Lash-L:


humbertogontijo (approved):

Lash-L:
i'm just trying stuff here, i'll ping you @humbertogontijo  if I think I figured something out. Everything I see shows that the GH_TOKEN should work. so I might try to get the GH_TOKEN out with an echo if it will let me and then try locally

PR #258, opened 2024-12-19
fix: add a mop mode to QRevoMaster

Lash-L:


rccoleman:
I'm was about to make a PR for this :)

Lash-L:
> I'm was about to make a PR for this :)

Feel free to use the actual codes if you have them. These are placeholders

rccoleman:
Yes, I have the correct mapping and I've tested it.  I'll make a PR in a bit.

PR #263, opened 2025-01-05
refactor: simplify future usage within the api clients

allenporter:
Simplify future usage within the api clients by separating how message results and exceptions are passed, using built in exception handling in futures. This raises exceptions instead of returning exceptions. This is a step of simplification on the path to fixing some concurrency issues in the client libraries.

Issue #228

### DIFF
--- a/roborock/version_a01_apis/roborock_mqtt_client_a01.py
+++ b/roborock/version_a01_apis/roborock_mqtt_client_a01.py
@@ -43,7 +43,7 @@ class RoborockMqttClientA01(RoborockMqttClient, RoborockClientA01):
         futures = []
         if "10000" in payload["dps"]:
             for dps in json.loads(payload["dps"]["10000"]):
-                futures.append(asyncio.ensure_future(self._async_response(dps, response_protocol)))
+                futures.append(self._async_response(dps, response_protocol))
         self._send_msg_raw(m)
         responses = await asyncio.gather(*futures, return_exceptions=True)
         dps_responses: dict[int, typing.Any] = {}
@@ -54,7 +54,7 @@ class RoborockMqttClientA01(RoborockMqttClient, RoborockClientA01):
                     self._logger.warning("Timed out get req for %s after %s s", dps, self.queue_timeout)
                     dps_responses[dps] = None
                 else:
-                    dps_responses[dps] = response[0]
+                    dps_responses[dps] = response
         return dps_responses
 
     async def update_values(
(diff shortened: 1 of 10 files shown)


allenporter:
I am sending this for initial feedback, but i have not tested exhaustively. I'd like to further ensure there is sufficient test coverage for these cases, etc.

roborock/version_a01_apis/roborock_mqtt_client_a01.py
```diff
@@ -54,7 +54,7 @@ async def send_message(self, roborock_message: RoborockMessage):
                     self._logger.warning("Timed out get req for %s after %s s", dps, self.queue_timeout)
                     dps_responses[dps] = None
                 else:
-                    dps_responses[dps] = response[0]
+                    dps_responses[dps] = response
```
humbertogontijo, line 57: I can't recall why the first position was used here. Is it safe to assign the entire array?

allenporter: My impression is that `_async_response` returns a tuple of (response, err), and this code was getting the response object out of the tuple. (The change here is related to every other change in this PR, it's not a random change).

humbertogontijo: I see

humbertogontijo (dismissed):
Seems like good improvements to me.
Any thoughts @Lash-L ?

Lash-L (dismissed):
I think this looks good to me, i'll want to do further testing before it's fully in HA, but i don't think that needs to hold up the package release

humbertogontijo:
Alright. I can merge it once conflicts are solved

allenporter:
I figured out a way to write tests some tests that use a fake MQTT broker, so let me get that working then i'll come back to this and resolve the conflicts and make sure everything passes.

Lash-L:
Using this as a quick chat

@allenporter @humbertogontijo 

Now that it is in the repo - we should all be able to merge in PRs.

But it should still be set so that you have to have one review. So whoever makes the PR should wait for one of the other two to approve it before merging.

allenporter:
OK, these changes should have improved test coverage now and ready for another look after rebasing/merging.

Lash-L (approved):
This all looks good to me - before we do the next bump in core though i'll want to run the package a good bit any make sure nothing breaks.

For now though, I think we are good to merge this, it probably makes sense for me to just check it after the whole refactoring process is done

PR #272, opened 2025-01-10
chore: use asyncio mode in tests

allenporter:
Update tests to use asyncio mode in pytest options to avoid needing to tag every test as async. I ran into problems when adding more asyncio tests for the mqtt broker that were resolved by changing this.  Also makes one test async where the constructor depends on the current asyncio loop and logs a warning.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,7 +1,6 @@
 from unittest.mock import patch
 
 import paho.mqtt.client as mqtt
-import pytest
 
 from roborock import (
     HomeData,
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -67,3 +67,6 @@ select=["E", "F", "UP", "I"]
 
 [tool.ruff.lint.per-file-ignores]
 "*/__init__.py" = ["F401"]
+
+[tool.pytest.ini_options]
```
Lash-L: I did this in my PR for the web tests, it should be in main

https://github.com/humbertogontijo/python-roborock/blob/3913e0178b49e2e917625e08824b8c19801d34a7/pyproject.toml#L72

allenporter: Thanks, rebased to main.

humbertogontijo (approved):

PR #278, opened 2025-01-12
test: add end-to-end tests for the MQTT client

allenporter:
Add end to end tests of the MQTT client using a fake socket. The tests patch the python socket library used by MQTT and return a fake socket that intercepts the send/recv methods, to avoid using real sockets during the test.

This copies some of the paho mqtt test fixtures and replies with real encoded mqtt packets. Really, this is probably too low level of a test to do in this client library, given we should just mock out the MQTT client. However, that is difficult to do because the client subclasses the MQTT client. Having the end to end tests means we can refactor the class inheritance structure with higher confidence that things won't break.

This adds tests for the async interfaces for connect and disconnect, and also tests with one sample message.

### DIFF
--- a/tests/mock_data.py
+++ b/tests/mock_data.py
@@ -21,14 +21,14 @@ USER_DATA = {
         "r": {
             "r": "US",
             "a": "https://api-us.roborock.com",
-            "m": "ssl://mqtt-us.roborock.com:8883",
+            "m": "tcp://mqtt-us.roborock.com:8883",  # Skip SSL code in MQTT client library
             "l": "https://wood-us.roborock.com",
         },
     },
     "tuyaDeviceState": 2,
     "avatarurl": "https://files.roborock.com/iottest/default_avatar.png",
 }
-
+LOCAL_KEY = "key123"
 HOME_DATA_RAW = {
     "id": 123456,
     "name": "My Home",
@@ -199,7 +199,7 @@ HOME_DATA_RAW = {
             "name": "Roborock S7 MaxV",
             "attribute": None,
             "activeTime": 1672364449,
-            "localKey": "key123",
+            "localKey": LOCAL_KEY,
             "runtimeEnv": None,
             "timeZoneId": "America/Los_Angeles",
             "iconUrl": "no_url",
@@ -339,3 +339,5 @@ BASE_URL_REQUEST = {
 }
 
 GET_CODE_RESPONSE = {"code": 200, "msg": "success", "data": None}
+
+MQTT_PUBLISH_TOPIC = "rr/m/o/user123/6ac2e6f8/abc123"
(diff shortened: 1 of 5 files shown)


tests/mock_data.py
```diff
@@ -339,3 +339,5 @@
 }
 
 GET_CODE_RESPONSE = {"code": 200, "msg": "success", "data": None}
+
+MQTT_PUBLISH_TOPIC = "rr/m/o/user123/6ac2e6f8/abc123"
```
Lash-L: Might make sense to reference the users id and device here similar to some other instances where you changed to a reference.

tests/test_api.py
```diff
@@ -126,3 +143,85 @@ async def test_get_prop():
...
+    received_requests: Queue,
+    response_queue: Queue,
+    mqtt_client: RoborockMqttClientV1,
+) -> None:
+    """Test sending an arbitrary MQTT message and parsing the response."""
+
+    response_queue.put(mqtt_packet.gen_connack(rc=0, flags=2))
+    response_queue.put(mqtt_packet.gen_suback(1, 0))
+    await mqtt_client.async_connect()
+    assert mqtt_client.is_connected()
```
Lash-L: You do this a few times and I could see us continuing to need to do it - does it make sense to make a fixture or helper that returns a connected client?

Lash-L (commented):
Two general comments - nothing major, glad you got this figured out!

I think in the future I'd like to figure out how to make some of this more generic and extendable. But that can come after the refactor.

I've always had this idea of getting a user per each device to run a script(that I could write pretty easily) that gets a bunch of Roborock responses for that specific device. Then we could change out the test data for each device and have a really robust test suite to make sure all devices are working and we don't have any major feature or bug gaps. Since all Roborock models have slightly different communication it would be a good sanity test

I think this is pretty far down the line, but something to keep in the he back of our minds

allenporter:
> I think in the future I'd like to figure out how to make some of this more generic and extendable. But that can come after the refactor.

Yeah, seems good. If you have ideas for what would be useful things to cover, let me know. I think we'll probably need to have tests with their own response handler e.g. parse the request, pull out something or assert something, and send a customized response.

> 
> I've always had this idea of getting a user per each device to run a script(that I could write pretty easily) that gets a bunch of Roborock responses for that specific device. Then we could change out the test data for each device and have a really robust test suite to make sure all devices are working and we don't have any major feature or bug gaps. Since all Roborock models have slightly different communication it would be a good sanity test

Yeah, that sounds like a great idea. I have a little [repository for Rainbird](https://github.com/allenporter/pyrainbird/tree/main/tests/testdata) sprinklers lwith test files for each [specific device](https://github.com/allenporter/pyrainbird/blob/main/tests/testdata/schedule_esp_me.yaml) and I found it pretty useful.

tests/mock_data.py
allenporter: Done.

Lash-L (approved):
Looks good to me!

Feel free to squash and merge, just change the commit to use chore instead of test so it passes commitlint

PR #279, opened 2025-01-12
chore: add codecov to CI

allenporter:
Add pytest codecov module and codecov to CI


allenporter:
I think this doesn't work when PRs come from forks, so not sure if this is worthwile.

Lash-L:
> I think this doesn't work when PRs come from forks, so not sure if this is worthwile.

how does HA make it work then?

I feel like i always get codecov messages but I always merge from my fork -> core

PR #282, opened 2025-01-13
chore: client library inheritance fixes and simplifications

allenporter:
Update the base classes to use the `abc` package, which provides the infrastructure for defining abstract base classes. Remove variables that are overridden in multiple places so that variables are only set once in the class hierarchy. The fixes include:
- Use `@abstractmethod` to avoid needing the `raise NotImplemented` reducing untested lines of code
- Set `_endpoint` only once, rather than multiple times. Removed the places where it is not used. (Question: Local commands use it when secured with "abc" but does it do anything?)
- Set `_logger` once for each client in the client itself so that the log messages are associated with the actual client being used. This will let us tell the difference between a local client disconnecting and a cloud client disconnecting which is difficult to do today
- Remove duplicate definition of `_send_msg_raw` in both the v1 mqtt client and the base mqtt client

Issue #228

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -7,6 +7,7 @@ import base64
 import logging
 import secrets
 import time
+from abc import ABC, abstractmethod
 from typing import Any
 
 from .containers import (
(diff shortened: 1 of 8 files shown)


Lash-L:
commitlint gets mad if the subject is capatilized, idk why.I've just never changed it. Semantic release I don't believe requires that to be the case

Lash-L (approved):
Looks good, a good bit of redundant code removed

PR #283, opened 2025-01-13
fix: bump commitlint and allow caps

Lash-L:
This should allow caps and get us past the commitlint breaking change.

Doing this as a fix to alos check and see if semantic release is working on this new repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: wagoid/commitlint-github-action@v5.4.5
+      - uses: wagoid/commitlint-github-action@v6.2.0
   lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


Lash-L:
@allenporter The pytest are failing here?

allenporter:
I re-ran it and it massed. I can't tell if there is a race or it actually took 4 seconds to do the handoff between threads, or if it is flaky for another reason. I may need to turn on debug on the CI to track it down.  Are those windows nodes typically slow?

allenporter:
> I re-ran it and it massed. I can't tell if there is a race or it actually took 4 seconds to do the handoff between threads, or if it is flaky for another reason. I may need to turn on debug on the CI to track it down. Are those windows nodes typically slow?

I think it is hitting the timeout. I saw windows flake on another PR's run and it took over two minutes.

What's the background for why this runs on windows? does it need to?

allenporter:
> > > I re-ran it and it massed. I can't tell if there is a race or it actually took 4 seconds to do the handoff between threads, or if it is flaky for another reason. I may need to turn on debug on the CI to track it down. Are those windows nodes typically slow?
> > 
> > 
> > I think it is hitting the timeout. I saw windows flake on another PR's run and it took over two minutes.
> > What's the background for why this runs on windows? does it need to?
> 
> Background was just extra testing to make super we didn't have anything platform specific. Probably not needed

OK, i will remove then if that is OK.  FWIW I have yet to ever have any windows (or mac compatibility) issues with python, so it should be pretty safe, but we can revisit of course if there is ever a regression on a specific platform.

PR #286, opened 2025-01-13
chore: remove level of inheritance in mqtt client

allenporter:
Pull the overrides related to the MQTT client library into a separate class, to avoid a level of inheritance. This makes it more clear which methods are part of the interaction with the MQTT client vs which are part of other base classes. It also makes it clear which overrides are MQTT client library specific vs part of the other parent classes.

Issue #228


humbertogontijo (approved):
I think this was a nice touch. Thanks

PR #292, opened 2025-01-13
chore: update CI to run on one platform, but multiple python versions

allenporter:
This will run the library on python 3.11, 3.12, 3.13 but drop windows and mac platforms. The motivation is that incompatibility issues seem more likely to happen from new python versions rather than different platforms. This matches most other project norms i've seen.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,11 +36,9 @@ jobs:
       matrix:
         python-version:
           - "3.11"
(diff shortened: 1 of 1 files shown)


Lash-L:
Good by me. I think since none of our libraries require any platform specific wheels (i.e. scikit-like packages) this is fine and should also speed our tests up.

Lash-L (approved):

Issue #222, opened 2024-08-20
InvalidStateError in RoborockFuture using roborock client


allenporter:
I don't know the exact steps to reproduce this, but noticed this while changing a device settings in HomeAssistant. It seems like this is an internal consistency issue and not something that that the caller should be able to introduce, but i'm not positive.

```
2024-08-20 03:49:28.365 ERROR (MainThread) [homeassistant] Error doing job: Exception in callback RoborockFuture._resolve((None, None)) (None):   File "/usr/local/lib/python3.12/asyncio/base_events.py", line 639, in run_forever
    self._run_once()
  File "/usr/local/lib/python3.12/asyncio/base_events.py", line 1977, in _run_once
    handle._run()
  File "/usr/local/lib/python3.12/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/lib/python3.12/asyncio/selector_events.py", line 970, in _read_ready
    self._read_ready_cb()
  File "/usr/local/lib/python3.12/asyncio/selector_events.py", line 1027, in _read_ready__data_received
    self._protocol.data_received(data)
  File "/home/vscode/.local/lib/python3.12/site-packages/roborock/local_api.py", line 39, in data_received
    self.on_message_received(parser_msg)
  File "/home/vscode/.local/lib/python3.12/site-packages/roborock/version_1_apis/roborock_client_v1.py", line 444, in on_message_received
    queue.resolve((data.payload, None))
  File "/home/vscode/.local/lib/python3.12/site-packages/roborock/roborock_future.py", line 22, in resolve
    self.loop.call_soon_threadsafe(self._resolve, item)
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/home/vscode/.local/lib/python3.12/site-packages/roborock/roborock_future.py", line 19, in _resolve
    self.fut.set_result(item)
asyncio.exceptions.InvalidStateError: invalid state
```

I believe this happened when i adjusted the same property two times in a row in a short succession (with a short pause between)

Lash-L:
I have also seen this issue but have not had a good way to recreate or debug. Would be open to any suggestions or even some debug improvements to figure out the cause

allenporter:
[...]

PR #300, opened 2025-01-21
fix: set unique sequence numbers on outgoing messages

allenporter:
Fix the message constructors to use a factory rather than a fixed constant. The bug is that the function is called only once, rather than on every construction. Add a test to verify that the messages are initialized with unique values.

Issue #228
Issue #222

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ ruff = "*"
 codespell = "*"
 pyshark = "^0.6"
 aioresponses = "^0.7.7"
+freezegun = "^1.5.1"
 
 [tool.semantic_release]
 branch = "main"
--- a/poetry.lock
+++ b/poetry.lock
@@ -273,6 +273,20 @@ testing = ["covdefaults (>=2.3)", "coverage (>=7.6.1)", "diff-cover (>=9.2)", "p
 typing = ["typing-extensions (>=4.12.2)"]
 
 [[package]]
+name = "freezegun"
+version = "1.5.1"
+description = "Let your Python tests travel through time"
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -45,6 +45,7 @@ ruff = "*"
 codespell = "*"
 pyshark = "^0.6"
 aioresponses = "^0.7.7"
+freezegun = "^1.5.1"
```
Lash-L, line 48: I think you should run poetry lock? I think dev dependencies also show up in the lock file. Not 100% though

Lash-L (approved):

pyproject.toml
allenporter: OK. To add this i rand `poetry add freezegun`.  When i run `poetry lock` it changes some other packages separate from freezegun: `pre-commit` and `identify`. I am not familiar with poetry, but is that expected and what you're looking for?

Lash-L: Oh if you added it with poetry add it should be fine.

Likely something else is pinning it and poetry is just using that. No need to push your poetry lock file then.

allenporter:
I think this fix may explain most of the issues we're seeing like https://github.com/home-assistant/core/issues/129582 or the invalidate state id since its assigning the same number to all requests.

pyproject.toml
allenporter: I ended up needing to rebase which had conflicts on the lock file so i went ahead and did it again and this time i see freezegun in there. Perhaps i did it wrong the first time.

Lash-L (approved):

PR #301, opened 2025-01-21
fix: threadsafe waiting queue

allenporter:
Make the waiting queue a threadsafe queue. 

This attempts to push the protocol to be a request key, but it seems to have odd interactions with local clients that munge the protocol. Will do some testing on a live system before marking ready for review. Parts of this may end up getting reverted.

Issue #228

### DIFF
--- a/roborock/roborock_future.py
+++ b/roborock/roborock_future.py
@@ -1,16 +1,64 @@
 from __future__ import annotations
 
+import logging
 from asyncio import Future
+from dataclasses import dataclass
+from threading import Lock
 from typing import Any
 
 import async_timeout
 
-from .exceptions import VacuumError
+from .exceptions import UnknownMethodError, VacuumError
+from .roborock_message import RoborockMessageProtocol
+
+_LOGGER = logging.getLogger(__name__)
+_TRIES = 3
+
+
+@dataclass(frozen=True)
+class RequestKey:
+    """A key for a Roborock message request."""
+
+    request_id: int
+    protocol: RoborockMessageProtocol | int = 0
+
+    def __str__(self) -> str:
+        """Get the key for the request."""
+        return f"{self.request_id}-{self.protocol}"
+
+
+class WaitingQueue:
+    """A threadsafe waiting queue for Roborock messages."""
+
+    def __init__(self) -> None:
+        """Initialize the waiting queue."""
+        self._lock = Lock()
+        self._queue: dict[RequestKey, RoborockFuture] = {}
+
+    def put(self, request_key: RequestKey, future: RoborockFuture) -> None:
+        """Create a future for the given protocol."""
+        _LOGGER.debug("Putting request key %s in the queue", request_key)
+        with self._lock:
+            if request_key in self._queue:
+                raise ValueError(f"Request key {request_key} already exists in the queue")
+            self._queue[request_key] = future
+
+    def safe_pop(self, request_key: RequestKey) -> RoborockFuture | None:
+        """Get the future from the queue if it has not yet been popped, otherwise ignore."""
+        _LOGGER.debug("Popping request key %s from the queue", request_key)
+        with self._lock:
+            return self._queue.pop(request_key, None)
 
 
 class RoborockFuture:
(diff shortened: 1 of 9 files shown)


roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -53,15 +54,19 @@ async def send_message(self, roborock_message: RoborockMessage):
             response_protocol = request_id + 1
```
allenporter, line 55: Is this trying to set the response protocol to the next thing, e.g. HELLO_REQUEST to HELLO_RESPONSE?  I think it ends up munging things up and i can't tell what this is trying to do.

humbertogontijo: Yes
I believe that it might be a mapping for dps request and response. But since I didn't have that I noticed that the response protocol was always a number higher

allenporter: OK great, maybe i can figure out another way to use the new protocol types. I'll think about this more and maybe:
(1) Add some tests to exercise these cases explicitly
(2) address first as a separate change
(3) come back to this

roborock/version_1_apis/roborock_client_v1.py
```diff
@@ -460,19 +462,21 @@ def on_message_received(self, messages: list[RoborockMessage]) -> None:
...
+                            self._logger.debug("Received response for unknown request id %s", request_key)
                 else:
-                    queue = self._waiting_queue.get(data.seq)
+                    request_key = RequestKey(data.seq, protocol)
+                    queue = self._waiting_queue.safe_pop(request_key)
                     if queue:
                         queue.set_result(data.payload)
                     else:
-                        self._logger.debug("Received response for unknown request id %s", data.seq)
+                        self._logger.debug("Received response for unknown request id %s", request_key)
```
Lash-L: Does it make sense for us to just keep the logic for if the item is not in the queue inside the waiting queue?

Although i guess that gets a bit tricky with the logger

roborock/roborock_future.py
```diff
@@ -1,16 +1,64 @@
...
+from threading import Lock
 from typing import Any
 
 import async_timeout
 
 from .exceptions import VacuumError
+from .roborock_message import RoborockMessageProtocol
+
+_LOGGER = logging.getLogger(__name__)
+_TRIES = 3
```
Lash-L: Is this used anywhere?

roborock/roborock_future.py
```diff
@@ -1,16 +1,64 @@
...
+        """Initialize the waiting queue."""
+        self._lock = Lock()
+        self._queue: dict[RequestKey, RoborockFuture] = {}
+
+    def put(self, request_key: RequestKey, future: RoborockFuture) -> None:
+        """Create a future for the given protocol."""
+        _LOGGER.debug("Putting request key %s in the queue", request_key)
+        with self._lock:
+            if request_key in self._queue:
+                raise ValueError(f"Request key {request_key} already exists in the queue")
```
Lash-L, line 42: Maybe this should be a subclass of RoborockException for easier error handling?

Lash-L:
@allenporter - do you want me to do a full review on this? Or is it still a WIP?

roborock/roborock_future.py
allenporter: Removed.

roborock/version_1_apis/roborock_client_v1.py
allenporter: Done.

roborock/roborock_future.py
allenporter: My thinking is that if this happens its a bug/shouldn't happen case and so it shouldn't be treated like a server side error. My thinking is: what do we expect a caller to do to handle this?

allenporter:
> @allenporter - do you want me to do a full review on this? Or is it still a WIP?

Still WIP -- I am worried the +1 logic commented above is not quite right and may raise exceptions.

allenporter:
Will revisit this with new aiomqtt based client.

PR #302, opened 2025-01-21
fix: fix queue timeout variable and set default in tests of 10 seconds

allenporter:
Update the class hierarchy for queue_timeout so that it is only set in the base classes. Right now it is passed up through the constructors, but sometimes uses the default arguments and get overwritten. This is an example of the problems with having multiple inheritance.

Issue #228

Update the tests to set a timeout of 10 up from 4 to see if it helps the tests be more resilient to slowness.

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -31,8 +31,9 @@ class RoborockClient(ABC):
     """Roborock client base class."""
 
     _logger: logging.LoggerAdapter
+    queue_timeout: int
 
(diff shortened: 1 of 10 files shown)


allenporter:
I think there is a bug in the queue timeout code. "roborock.exceptions.RoborockTimeout: id=0 Timeout after 4 seconds" as it doesn't seem to respect constructor arguments.

PR #303, opened 2025-01-21
chore: fix flaky tests by cleaning up threads

allenporter:
Explicitly stop the mqtt loop when the client library shuts down to avoid leaking threads. Explicitly release clients during test cleanup, but make it best effort.

### DIFF
--- a/roborock/cloud_api.py
+++ b/roborock/cloud_api.py
@@ -74,6 +74,12 @@ class RoborockMqttClient(RoborockClient, ABC):
         self._waiting_queue: dict[int, RoborockFuture] = {}
         self._mutex = Lock()
 
+    async def async_release(self) -> None:
+        """Release the MQTT client."""
+        await super().async_release()
+        if self._mqtt_client:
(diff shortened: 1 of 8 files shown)


allenporter:
It seems like propperly cleaning up the mqtt threads fixes the issue. it seemed like there could be 12 threads lying around at the end of a few tests which may have put too much pressure on the machine.

roborock/cloud_api.py
```diff
@@ -74,6 +74,11 @@ def __init__(self, user_data: UserData, device_info: DeviceData, queue_timeout:
         self._waiting_queue: dict[int, RoborockFuture] = {}
         self._mutex = Lock()
 
+    async def async_release(self) -> None:
+        """Release the MQTT client."""
+        await super().async_release()
+        await self.event_loop.run_in_executor(None, self._mqtt_client.loop_stop)
```
Lash-L: Should this be called in async_disconnect() as well?

allenporter: I was considering this originally, but changed it to be conservative. I'll go ahead and make this change.

Lash-L (approved):

PR #307, opened 2025-01-24
feat: add commands to add a new device

Lash-L:
Some findings from my recent reverse engineering

### DIFF
--- a/roborock/web_api.py
+++ b/roborock/web_api.py
@@ -9,7 +9,7 @@ import secrets
 import time
 
 import aiohttp
-from aiohttp import ContentTypeError
+from aiohttp import ContentTypeError, FormData
 
 from roborock.containers import HomeData, HomeDataRoom, ProductResponse, RRiot, UserData
 from roborock.exceptions import (
@@ -112,7 +112,6 @@ class RoborockApiClient:
         base_url = user_data.rriot.r.a
         prepare_request = PreparedRequest(base_url)
         hid = await self._get_home_id(user_data)
-        from aiohttp import FormData
 
         data = FormData()
         data.add_field("hid", hid)
@@ -138,7 +137,9 @@ class RoborockApiClient:
 
     async def add_device(self, user_data: UserData, s: str, t: str) -> dict:
         """This will add a new device to your account
-        it is recommended to only use this if you know what you are doing."""
+        it is recommended to only use this during a pairing cycle with a device.
+        Please see here: https://github.com/Python-roborock/Roborockmitmproxy/blob/main/handshake_protocol.md
+        """
         if (
             user_data.rriot is None
             or user_data.rriot.r is None
--- a/roborock/web_api.py
+++ b/roborock/web_api.py
@@ -67,9 +67,25 @@ class RoborockApiClient:
         md5.update(self._device_identifier.encode())
         return base64.b64encode(md5.digest()).decode()
 
-    def _get_hawk_authentication(self, rriot: RRiot, url: str) -> str:
+    def _process_extra_hawk_values(self, values: dict | None) -> str:
+        if values is None:
+            return ""
+        else:
+            sorted_keys = sorted(values.keys())
+            result = []
+            for key in sorted_keys:
+                value = values.get(key)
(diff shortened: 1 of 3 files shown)


roborock/web_api.py
```diff
@@ -67,22 +67,107 @@ def _get_header_client_id(self):
...
+                formdata_str,
             ]
         )
         mac = base64.b64encode(hmac.new(rriot.h.encode(), prestr.encode(), hashlib.sha256).digest()).decode()
-        return f'Hawk id="{rriot.u}", s="{rriot.s}", ts="{timestamp}", nonce="{nonce}", mac="{mac}"'
+        return f'Hawk id="{rriot.u}",s="{rriot.s}",ts="{timestamp}",nonce="{nonce}",mac="{mac}"'
+
+    async def nc_prepare(self, user_data: UserData, timezone: str) -> dict:
+        """This gets a few critical parameters for adding a device to your account."""
+        if (
```
allenporter, line 105: For a follow up: I've seen this kind of validation in a few of theclients. Should any of these fields be made required as part of `UserData`? e.g. either not `None` or `__post__init__` validation?

roborock/web_api.py
```diff
@@ -67,22 +67,107 @@ def _get_header_client_id(self):
...
+        if prepare_response is None:
+            raise RoborockException("prepare_response is None")
+        if not prepare_response.get("success"):
+            raise RoborockException(f"{prepare_response.get('msg')} - response code: {prepare_response.get('code')}")
+
+        return prepare_response["result"]
+
+    async def add_device(self, user_data: UserData, s: str, t: str) -> dict:
+        """This will add a new device to your account
+        it is recommended to only use this if you know what you are doing."""
```
allenporter: I'm not sure i understand how this API is intended to be used if its not recommended to be called. Is this some prep work for the future or something else?

I think adding some additional context, even for future maintainers of the library, would be helpful. e.g. if its just for experimentation, thats also fine to say explicitly.

roborock/web_api.py
```diff
@@ -67,22 +67,107 @@ def _get_header_client_id(self):
...
-                "",
+                params_str,
+                formdata_str,
             ]
         )
         mac = base64.b64encode(hmac.new(rriot.h.encode(), prestr.encode(), hashlib.sha256).digest()).decode()
-        return f'Hawk id="{rriot.u}", s="{rriot.s}", ts="{timestamp}", nonce="{nonce}", mac="{mac}"'
+        return f'Hawk id="{rriot.u}",s="{rriot.s}",ts="{timestamp}",nonce="{nonce}",mac="{mac}"'
+
+    async def nc_prepare(self, user_data: UserData, timezone: str) -> dict:
```
allenporter, line 103: Will this be made part of normal client initialization at some point or is this for special use cases only? I notice `UserData` is a constructor argument to the other clients so wondering if this should be required going forward.

roborock/web_api.py
```diff
@@ -67,22 +67,107 @@ def _get_header_client_id(self):
...
+            user_data.rriot is None
+            or user_data.rriot.r is None
+            or user_data.rriot.u is None
+            or user_data.rriot.r.a is None
+        ):
+            raise RoborockException("Your userdata is missing critical attributes.")
+        base_url = user_data.rriot.r.a
+        prepare_request = PreparedRequest(base_url)
+        hid = await self._get_home_id(user_data)
+        from aiohttp import FormData
```
allenporter: why do this inline rather than at the top? is it expensive or something?

roborock/web_api.py
Lash-L: I think when we get around to collecting a bunch of different volunteered user data for different devices for testing, we will have a better idea of what is 100% always going to be there and we can remove a lot of the | None = None

roborock/web_api.py
Lash-L: Accidental

roborock/web_api.py
Lash-L: Good call, fixed.

allenporter (approved):

PR #308, opened 2025-01-26
fix: remove the __del__ warning for disconnected clients

allenporter:
This had some problems in home assistants unit test. I think it is reasonable to just require clients to shut down things properly and not have this.

This is needed to update python-roborock in Home Assistant.

### DIFF
--- a/roborock/api.py
+++ b/roborock/api.py
@@ -47,10 +47,6 @@ class RoborockClient(ABC):
         }
         self.is_available: bool = True
 
-    def __del__(self) -> None:
-        if self.is_connected():
-            self._logger.debug("Roborock is connected while being released")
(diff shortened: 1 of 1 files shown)


Lash-L:
Failing tests? Not sure if it's just a stubborn test

Lash-L (approved):

allenporter:
It appears that some tests are flaky, darn. I'll try to see if i can reproduce and take a look.

Maybe faking the sockets and/or using the real threads is too complex, or maybe its spotting a real issue with how we're using the mqtt client, I can't tell yet.

allenporter:
(Re-running the test and it passed so i think its flaky)

PR #309, opened 2025-01-26
fix: don't wait for the mqtt thread to fully shutdown when disconecting.

allenporter:
The motivation is because this speeds up the tests to not have to wait for all threads to fully shut down. This can take a moment since it does a join of the thread.


Lash-L (approved):

allenporter:
This also had a flaky test. Looking at the test timeout, it shows a bunch of threads still running, so this change has the desired effect, but i think it leaves the threads stacked up at the end. Maybe it's too resource intensive to do this.

PR #310, opened 2025-01-26
fix: never create a new asyncio loop

allenporter:
Always reuse the existing asyncio loop rather than creating a new one.

This is needed to upgrade python-roborock in Home Assistant since the async shutdown call must happen in the same asyncio loop that was used to create the client.

This requires inlining all calls to get the currently running loop, because the constructor for the MQTT client needs to be created from outside the asyncio loop because the MQTT client does blocking I/O in the constructor. (example: https://github.com/home-assistant/core/pull/124266). In the future we can move this back to the constructor, but that requires separately creating the MQTT client in a way that doesn't block, which we can do in the future when using a shared MQTT client as proposed by @Lash-L.

### DIFF
--- a/roborock/cloud_api.py
+++ b/roborock/cloud_api.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 import threading
 from abc import ABC
@@ -158,7 +159,8 @@ class RoborockMqttClient(RoborockClient, ABC):
             if disconnected_future := self._sync_disconnect():
                 # There are no errors set on this future
                 await disconnected_future
-            await self.event_loop.run_in_executor(None, self._mqtt_client.loop_stop)
+            loop = asyncio.get_running_loop()
+            await loop.run_in_executor(None, self._mqtt_client.loop_stop)
 
     async def async_connect(self) -> None:
         async with self._mutex:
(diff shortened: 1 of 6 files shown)


roborock/cloud_api.py
```diff
@@ -158,7 +159,8 @@ async def async_disconnect(self) -> None:
             if disconnected_future := self._sync_disconnect():
                 # There are no errors set on this future
                 await disconnected_future
-            await self.event_loop.run_in_executor(None, self._mqtt_client.loop_stop)
+            loop = asyncio.get_running_loop()
```
Lash-L, line 162: My only thought is if there is any chance that an event loop doesn't exist at these points. But even for non-ha users, it should as it's needed for the web api so it should be okay

Lash-L (approved):

roborock/cloud_api.py
allenporter: Yeah my thinking is that if it has, to exist to be correct. If it doesn't exist then failing loudly is best.

We can label this as a breaking change? Not sure how semantic commit handles this.. "refactor"?

Lash-L: I don't think we need to mark it as a breaking change - partially as i am also not sure how to do that with semantic release haha

It should be fine as most users will create the web api first and will get an event loop there if it doesn't exist.

PR #331, opened 2025-02-11
fix: add some extra data protocol checking

Lash-L:
Kind of just a intermediary step so users don't need to get any kind of logging warnings for something we know of.

### DIFF
--- a/roborock/roborock_message.py
+++ b/roborock/roborock_message.py
@@ -36,6 +36,7 @@ class RoborockDataProtocol(RoborockEnum):
     TASK_CANCEL_IN_MOTION = 132
     CHARGE_STATUS = 133
     DRYING_STATUS = 134
+    OFFLINE_STATUS = 135
 
     @classmethod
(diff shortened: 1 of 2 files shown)


roborock/roborock_message.py
```diff
@@ -36,6 +36,7 @@ class RoborockDataProtocol(RoborockEnum):
     TASK_CANCEL_IN_MOTION = 132
     CHARGE_STATUS = 133
     DRYING_STATUS = 134
+    OFFLINE_STATUS = 135
```
allenporter, line 39: Did you want to do something with this that you're adding?

allenporter (approved):
Looks good. You can also consider logging once per type.

roborock/roborock_message.py
Lash-L: Nah I don't really know anything about it yet. Just adding it for labeling sake to make life easier in the future.

PR #333, opened 2025-02-14
add cli status

regevbr:


regevbr:
@Lash-L would appreciate your CR

Lash-L (approved):

Lash-L:
Fine by me - I can't really recommend using it regularly though. i.e. this makes sense to only ever use as a once off cli call. Never put into any kind of bash script for regular calling.

regevbr:
> Fine by me - I can't really recommend using it regularly though. i.e. this makes sense to only ever use as a once off cli call. Never put into any kind of bash script for regular calling.

Agreed, for debugging and development only

PR #347, opened 2025-03-14
feat: add home_data_v3

Lash-L:


roborock/web_api.py
```diff
@@ -332,6 +332,29 @@ async def get_home_data_v2(self, user_data: UserData) -> HomeData:
...
             raise RoborockException("home_response result was an unexpected type")
 
+    async def get_home_data_v3(self, user_data: UserData) -> HomeData:
+        """This is the same as get_home_data, but uses a different endpoint and includes non-robotic vacuums."""
+        rriot = user_data.rriot
+        if rriot is None:
+            raise RoborockException("rriot is none")
+        home_id = await self._get_home_id(user_data)
+        if rriot.r.a is None:
+            raise RoborockException("Missing field 'a' in rriot reference")
```
allenporter, line 340: Swap order so the two rriot checks are nearby?

```suggestion
        if rriot.r.a is None:
            raise RoborockException("Missing field 'a' in rriot reference")
        home_id = await self._get_home_id(user_data)
```

roborock/web_api.py
```diff
@@ -332,6 +332,29 @@ async def get_home_data_v2(self, user_data: UserData) -> HomeData:
...
+        home_request = PreparedRequest(
+            rriot.r.a,
+            {
+                "Authorization": self._get_hawk_authentication(rriot, "/v3/user/homes/" + str(home_id)),
+            },
+        )
+        home_response = await home_request.request("get", "/v3/user/homes/" + str(home_id))
```
allenporter: why is the cast needed?

roborock/web_api.py
```diff
@@ -332,6 +332,29 @@ async def get_home_data_v2(self, user_data: UserData) -> HomeData:
...
+        if not home_response.get("success"):
+            raise RoborockException(home_response)
+        home_data = home_response.get("result")
+        if isinstance(home_data, dict):
+            return HomeData.from_dict(home_data)
+        else:
+            raise RoborockException("home_response result was an unexpected type")
```
allenporter:
```suggestion
        home_data = home_response.get("result")
        if isinstance(home_data, dict):
            return HomeData.from_dict(home_data)
        raise RoborockException(f"home_response result was an unexpected type: {home_data}")
```

allenporter (changes requested):

roborock/web_api.py
allenporter: Thinking its helpful to have details about the failture ("I wonder what type it was then?")

roborock/web_api.py
Lash-L: likely isn't, this is just an old function copied with some small changes. I'll fix it up a bit.

roborock/web_api.py
Lash-L: I removed one rriot check as I am instead setting the value to not optional.

allenporter (approved):

PR #348, opened 2025-03-15
Fix Links to documentation

Craneson:
The links to the documentation and to the supported devices were not working because they were quoted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Install this via pip (or your favourite package manager):
 
 ## Functionality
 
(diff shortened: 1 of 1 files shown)


allenporter (approved):

allenporter:
Hi, can you fix the commit messages in this PR to be in the format that is expected by the linter?

allenporter (changes requested):
Requesting changes to fix CI from commit lint

Lash-L (approved):

Lash-L:
We can actually fix it via the squash @allenporter - then commitlint will only look at the squash commit message

PR #354, opened 2025-03-17
feat: allow passing in clientsession

Lash-L:
This allows passing in a clientsession object. If one isn't provided, it will auto create one.


allenporter (approved):
Nice!

PR #355, opened 2025-03-21
feat: allow forcing of updating cache variables

Lash-L:
This whole system likely needs some love and refactoring. But for now, we shouldn't be returning a cached value for get_status.

We should update the cached value, but it doesn't make sense for us to return it, as it is the most important command for us to get.

### DIFF
--- a/roborock/version_1_apis/roborock_client_v1.py
+++ b/roborock/version_1_apis/roborock_client_v1.py
@@ -107,9 +107,9 @@ class AttributeCache:
             raise err
         return self._value
 
-    async def async_value(self):
+    async def async_value(self, force: bool = False):
         async with self._mutex:
-            if self._value is None:
+            if self._value is None or force:
                 return await self.task.reset()
             return self._value
 
@@ -175,7 +175,7 @@ class RoborockClientV1(RoborockClient, ABC):
         return self._status_type
 
     async def get_status(self) -> Status:
-        data = self._status_type.from_dict(await self.cache[CacheableAttribute.status].async_value())
+        data = self._status_type.from_dict(await self.cache[CacheableAttribute.status].async_value(force=True))
         if data is None:
             return self._status_type()
         return data


roborock/version_1_apis/roborock_client_v1.py
```diff
@@ -175,7 +175,7 @@ def status_type(self) -> type[Status]:
         return self._status_type
 
     async def get_status(self) -> Status:
-        data = self._status_type.from_dict(await self.cache[CacheableAttribute.status].async_value())
+        data = self._status_type.from_dict(await self.cache[CacheableAttribute.status].async_value(force=True))
```
allenporter, line 178: Is this equivalent to something like this?
```
data = await self.send_command(RoborockCommand.GET_STATUS, None)
```

It seems like if the goal is to skip the cache then not using it at all seems more straight forward, even if its a little more code.

Are there places where this data will still be served from cache by other commands or is it in fact never used like that like i'm assuming?

Lash-L: It does get used for the automatic protocol listeners/updaters.

I also believe send_command() will check the cache first. 

I'm not even sure if it makes a ton of sense for the cache to exist at all in it's current state, but it would be a pretty large refactor that doesn't make sense to do unless we are doing the larger refactor (i.e. more explicit devices in the library)

allenporter: So then `self._send_command`?  I'm thinking whatever command is passed into the cache itself could be called directly:
https://github.com/Python-roborock/python-roborock/blob/03b0eaa6074ccaafb6e3965842b4c040da910e8d/roborock/version_1_apis/roborock_client_v1.py#L160

Lash-L: Likely yes - but I worry about the cache and the latest information getting out of sync. It might not be a problem, but i'm not sure

I.e. a scenario where:

I update manually (without the cache)

A protocol message comes in and the cache updates overriding anything that was received from the manual update

allenporter: I see. Can `CacheableAttribute.status` go away entirely as a breaking change?

Generally, I agree on the larger overhaul. It seems odd that an entity in home assistant is overwriting coordinator data: https://github.com/home-assistant/core/blob/b2942d61b3f521763916277cfd0f99d00bcd7ec8/homeassistant/components/roborock/entity.py#L185 -- the responsibilities are a little backward there.

Lash-L: I think yes - but I think it is going to be a bit of a pain.

If it's fine with you, i'd like to move forward with this direction.

My next big change in this library is still the MqttManager, and I worry changing this now/ how protocol listeners update now will make that more difficult.

I think it makes sense to finish the manager and then look at the attributecache/ protocol listeners and how they can be better handled.

This is just a slight bug fix, so I'd like to get it in before everything else.

allenporter: Yes, sounds good. Thanks for talking me through it.

allenporter (approved):

PR #364, opened 2025-04-09
Refactor Mqtt Manager and Roborock Device

Lash-L:


roborock/mqtt_manager.py
```diff
@@ -0,0 +1,113 @@
...
+    publish_function: Coroutine[None]
+    unsubscribe_function: Coroutine[None]
+    subscribe_function: Coroutine[None]
+
+
+class RoborockMqttManager:
+    client_wrappers: dict[str, ClientWrapper] = {}
+    _instance: Self = None
+
+    def __new__(cls) -> RoborockMqttManager:
```
allenporter: I don't think this class should impose that itself must be a singleton, but instead have this managed by the callers.


The caller can create  a single mqtt client for each account, which is used for all communication. This removes the need for this class itself to track the notion of different sessions for different user accounts.

Can this be removed for now and we figure out how to address the downstream implications in a follow up?


Instead of `RoborockDevice` creating its own `RoborockMqttManager`, it would be passed in through its constructor. Essentially `_new_connect` args can be constructor args.

roborock/mqtt_manager.py
```diff
@@ -0,0 +1,113 @@
...
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
+    async def connect(self, user_data: UserData):
```
allenporter: I think this class should not get into the business of dealing with roborock API internals in this class and have a little shim separation later.

I just made up this naming so feel free to change but you get the idea:

```python
@dataclass
class SessionParams:
    host: str
    port: int
    hashed_user: str
    hashed_password: str

    
def create_mqtt_session_params(user_data: UserData) -> SessionParams:
    """Create MQTT Session params from Roborock API data."""
    ....
```

This can separate logic, make things easier to test, etc.


Similarly for the device, instead of taking a `DeviceData` I think this class only needs to accept a `device_duid` string and should be responsible for dispatching messages to some callback.  The payload decoding should live next to the payload encoding, not in here. (or the payload decoder can simply be a `Callable` we stick somewhere else that use the `local_key`.  )

roborock/device_trait.py
```diff
@@ -0,0 +1,79 @@
+import datetime
+from abc import ABC, abstractmethod
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
+
+from . import RoborockCommand
+from .containers import Consumable, DeviceFeatures, DnDTimer, RoborockBase
+
+
+@dataclass
```
allenporter, line 9: The general shape looks good with the base class that supports listeners and handling updates.

I'm wondering if we  can merge the trait and the dataclass objects. The API right now is like this:

```
if device.dnd:
    print(device.dnd.status.enabled)
```

But perhaps that extra level of `status` is worth trying to avoid? Then it could look like this:

```
if device.dnd:
    print(device.dnd.enabled)
```

The down side of this is that the dataclass has to become mutable, and we will likely want more subclassing (e.g. There is already `RoborockBase` for handling parsing of objects, but should it also have listeners? or make another class hierarchy? just want to avoid multiple paths to inherit `RoborockBase`).

roborock/device_trait.py
```diff
@@ -0,0 +1,79 @@
...
+class DndTrait(DeviceTrait):
+    handle_command: RoborockCommand = RoborockCommand.GET_DND_TIMER
+    _status_type: type[DnDTimer] = DnDTimer
+    status: DnDTimer
+
+    def __init__(self, send_command: Callable[..., Awaitable[None]]):
+        super().__init__(send_command)
+
+    @classmethod
+    def supported(cls, features: DeviceFeatures) -> bool:
```
allenporter: Right now the logic for this mapping is spread across a few places, the device features object and the trait itself, but wondering if it can live in a single place.

I think it either needs to be:
- in the device features code, directly point to the trait, then look up the reference when building the device.  i'm feeling like the first option may be simplest, though it flips around the dependencies it may have to move if going that way.
- Define an enums the trait can reference then the parsing code can look it up in the right place.  This may be a challenge since there are multiple ways to determine if something is supported.
- push the "is supported" logic only in the trait (and figure out the right apis to pass the needed data through when building the device). This may be a challenge since there are multiple ways to determine if something is supported.

roborock/device_trait.py
Lash-L: So I just pushed a change that attempts to do this. It's a bit weird with the to and from dict logic. I started wondering though if the logic we use right now is overkill. We do all sorts of camelization and decamelization, but maybe it is better to just be explicit. It is less 'engineered'/ less fancy. But it is probably less prone to errors? I didn't touch the support features stuff you talked about. But want to take a look at that in the diff and see what you think about it/ if you have other feedback? We'd basically remove containers for anything that is a device trait.

roborock/mqtt_manager.py
Lash-L: I believe this has all been handled, so i'm going to resolve it

roborock/mqtt_manager.py
Lash-L: I believe this has all been handled by your changes, so i'm going to resolve it

roborock/roborock_message.py
```diff
@@ -161,11 +161,17 @@ class RoborockMessage:
     random: int = field(default_factory=lambda: get_next_int(10000, 99999))
     timestamp: int = field(default_factory=lambda: math.floor(time.time()))
     message_retry: MessageRetry | None = None
+    _parsed_payload: dict | None = None
+
+    def get_payload(self) -> dict | None:
+        if self.payload and not self._parsed_payload:
+            self._parsed_payload = json.loads(self.payload.decode())
+        return self._parsed_payload
```
allenporter: Could be equivalent to this?

```
    @cache
    def get_payload(self) -> dict | None:
        return json.loads(self.payload.decode())
```

roborock/roborock_message.py
```diff
@@ -161,11 +161,17 @@ class RoborockMessage:
...
 
     def get_request_id(self) -> int | None:
-        if self.payload:
-            payload = json.loads(self.payload.decode())
-            for data_point_number, data_point in payload.get("dps").items():
+        payload = self.get_payload()
+        if payload:
```
allenporter: Can be one-liner here and below:
```suggestion
        if payload := self.get_payload()
```

allenporter: If below these get too long for single lines you could switch to early return:

```
if self.protocol not in {...]
    return None
if (payload := self.get_payload()) is None:
    return None
... 
```

roborock/roborock_device.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, user_data: UserData, device_info: DeviceData):
+        self.user_data = user_data
+        self.device_info = device_info
+        self.data = None
+        self._logger = RoborockLoggerAdapter(device_info.device.name, _LOGGER)
+        self._mqtt_endpoint = base64.b64encode(Utils.md5(user_data.rriot.k.encode())[8:14]).decode()
+        rriot = user_data.rriot
+        hashed_user = md5hex(rriot.u + ":" + rriot.k)[2:10]
+        url = urlparse(rriot.r.m)
+        mqtt_password = rriot.s
```
allenporter: Looks unused

roborock/roborock_device.py
```diff
@@ -0,0 +1,156 @@
...
+            return self.send_message(method, params, use_cloud)
+
+        if trait.supported(self.device_info.device_features):
+            trait_instance = trait(_send_command)
+            self._all_supported_traits.append(trait(_send_command))
+            self._command_to_trait[trait.handle_command] = trait_instance
+            return trait_instance
+        return None
+
+    async def connect(self):
```
allenporter, line 68: I know you have a TODO above, but i think it would be best to just now pass in a connected session and not add this method, or add a TODO remove here i guess.

roborock/roborock_device.py
```diff
@@ -0,0 +1,156 @@
...
+                username=hashed_user,
+                password=md5hex(rriot.s + ":" + rriot.k)[16:],
+            )
+        )
+        self.input_topic = f"rr/m/i/{rriot.u}/{hashed_user}/{device_info.duid}"
+        self.output_topic = f"rr/m/o/{rriot.u}/{hashed_user}/{device_info.duid}"
+
+    def determine_supported_traits(self, trait: type[DeviceTrait]):
+        def _send_command(
+            method: RoborockCommand | str, params: list | dict | int | None = None, use_cloud: bool = True
```
allenporter, line 57: Can we omit `use_cloud` if we're not using it yet?

roborock/containers.py
```diff
@@ -840,6 +1037,22 @@ class DeviceData(RoborockBase):
     device: HomeDataDevice
     model: str
     host: str | None = None
+    product_nickname: RoborockProductNickname | None = None
```
allenporter, line 939: Where is this set or passed in?

roborock/code_mappings.py
```diff
@@ -50,6 +50,144 @@ def items(cls: type[RoborockEnum]):
         return cls.as_dict().items()
 
 
+class RoborockProductNickname(StrEnum):
```
allenporter: Given there is a 1:1 between this enum and the `short_model_to_enum` we can specify this in a single enum.

e.g. I believe you can do something like this:
```
   class Color(Enum):
       RED = (255, 0, 0)
       GREEN = (0, 255, 0)
       BLUE = (0, 0, 255)
```

And it can be 
```
>>> Color.RED.value
(255, 0, 0)
```

Or you can give the fields name:

```
from enum import Enum
from collections import namedtuple

RGB = namedtuple('RGB', ['red', 'green', 'blue'])

class Color(Enum):
  RED = RGB(255, 0, 0)
  GREEN = RGB(0, 255, 0)
  BLUE = RGB(0, 0, 255)

print(Color.RED.value.red)  # Output: 255
print(Color.BLUE.value.blue) # Output: 255
```

You can still build a map after the fact iterating over the values. Nice because this saves a lot of verticle space and makes it easier to update.

roborock/containers.py
```diff
@@ -294,144 +296,339 @@ class HomeDataDevice(RoborockBase):
...
-        support_floor_direction=new_feature_set_mod_8 and bool(2048 & converted_new_feature_set),
-        wifi_manage_supported=bool(128 & converted_new_feature_set),
-        back_charge_auto_wash_supported=bool(4096 & converted_new_feature_set),
-        support_incremental_map=bool(8192 & converted_new_feature_set),
-        offline_map_supported=bool(16384 & converted_new_feature_set),
-    )
+    """Represents the features supported by a Roborock device."""
+
+    # Features derived from robot_new_features
+    is_map_carpet_add_support: bool
```
allenporter: We can add the details about the bits into the metadata for the field itself like this:
```
from dataclasses import dataclass, field
from typing import Any

@dataclass
class Example:
    is_map_carpet_add_support: bool = field(metadata={'feature': '1073741824'})
```

You can use `fields()` to iterate over all the fields and access the `.metadata` to get the dict.  The field could be named different things for different feature types e.g.  (`"upper_32_bits":` or `"feature_bit_str":`) and change the logic.

roborock/device_trait.py
```diff
@@ -0,0 +1,57 @@
...
+from .containers import DeviceFeatures, RoborockBase
+
+
+@dataclass
+class DeviceTrait(ABC):
+    handle_command: RoborockCommand
+
+    def __init__(self, send_command: Callable[..., Awaitable[None]]):
+        self.send_command = send_command
+        self.status: RoborockBase | None = None
```
allenporter: What is this going to be used for? Can we remove for now if not used?

roborock/device_trait.py
```diff
@@ -0,0 +1,57 @@
...
+    @abstractmethod
+    def supported(cls, features: DeviceFeatures) -> bool:
+        raise NotImplementedError
+
+    @abstractmethod
+    def from_dict(cls, data: dict) -> bool:
+        raise NotImplementedError
+
+    def on_message(self, data: dict) -> None:
+        self.status = self.from_dict(data)
```
allenporter: Typically `from_dict` is a factor method so here i think its used more like `update`.

roborock/device_trait.py
allenporter: I think we still can find ways to simplify this, but not yet seeing it.  Maybe it's but cutting some of the requirements for now that are complex, then adding them in a follow up? The challenge is to have a nice API you like, then figure out how to add workarounds to add in extra constraints w/o uglifying the API.

Can we cut requirements for now that every trait has its own subscriber? I feel like that adds a little complexity i think we can work up to.


How about reducing the number of states a trait can be in? Right now we have these states to consider:
- Can be None (not supported)
- Can be present, but all fields are None (supported but not populated yet yet)
- Can be present with a status (support but failed population on first message?)
- Can be present with some fields None and some not None (does it happen? maybe depends on device? unclear to caller)

I think we need to remove 2 of these if we can... e.g. present with all None fields seems like it could be nice to avoid if its not meaningful.

Parsing wise, it feels like `Dnd` should be a `@dataclass` given all that initialization. Thee are also parsing libraries we can use e.g. mashumaro

Lash-L: Will responding to the rest when I'm at my machine. As far as serialization goes. We used dacite in the past and did camelization of atttibutes as sometimes they would randomly throw in a snake case attribute. This added some complexity, and when I removed the dacite dependency, I just wrote my own serializer as there were a few things that were happening that didn't work well with the libraries I tried(including madhumaru). Perhaps while we are already making a breaking change, we can switch to a library that may handle things slightly differently.

roborock/containers.py
Lash-L: oh that's a cool idea. i like it

roborock/containers.py
Lash-L: It is set in post init

roborock/device_trait.py
Lash-L: Maybe:

The "update" method can actually be a class method and return a instance of the trait. 

Before we get any data it is None.

So long as it is supported, when we get data we update it. 

self.dnd = DndTrait.update(data)

Not sure if that potentially opens the door to async problems.

Thoughts @allenporter

Lash-L: Although it gets a bit trickier with something like Consumable. Where different devices have different attributes on them.  and therefore have to have some None attributes

PR #366, opened 2025-04-12
Add an aiomqtt based MQTT session module

allenporter:
Add an aiomqtt based MQTT session module. This is based on #364 with some additional unit tests and support for retries.

The session is created by invoking `create_mqtt_session`.  A caller may subscribe to a topic, and the session keeps  track of which callbacks to invoke for each topic. The client is run as a background task that will run until shutdown. Once connected, the client will wait for messages to be received in a loop. If the connection is lost, the client will be re-created and reconnected. There is backoff to avoid spamming the broker with connection attempts. The client
will automatically re-establish any subscriptions when the connection is re-established.

It is very difficult to exercise the failure cases with aiomqtt, so these will need to be exercised manually. This does not contain any roborock specific information.


allenporter:
Looks like the tests fail when run from CI. Taking a look.

roborock/mqtt/roborock_session.py
```diff
@@ -0,0 +1,213 @@
...
+        await start_future
+
+    async def close(self) -> None:
+        """Cancels the MQTT loop and shutdown the client library."""
+        if self._background_task:
+            self._background_task.cancel()
+            try:
+                await self._background_task
+            except asyncio.CancelledError:
+                pass
```
Lash-L, line 89: I guess it likely doesn't matter as if you're getting a Cancellederror - likely the object will no longer exists - but do we want to set self._healthy here?

roborock/mqtt/roborock_session.py
```diff
@@ -0,0 +1,213 @@
...
+                yield client
+        finally:
+            async with self._client_lock:
+                self._client = None
+
+    async def _process_message_loop(self, client: aiomqtt.Client) -> None:
+        _LOGGER.debug("Processing MQTT messages")
+        async for message in client.messages:
+            _LOGGER.debug("Received message: %s", message)
+            for listener in self._listeners.get(message.topic.value) or []:
```
Lash-L: is the or [] better than .get(message.topic.value, []) because the second part of a .get() is always evaluated in python?

roborock/mqtt/roborock_session.py
```diff
@@ -0,0 +1,213 @@
...
+
+        return lambda: self._listeners[topic].remove(callback)
+
+    async def publish(self, topic: str, message: bytes) -> None:
+        """Publish a message on the topic."""
+        _LOGGER.debug("Sending message to topic %s: %s", topic, message)
+        async with self._client_lock:
+            if not self._client:
+                raise RoborockException("MQTT client not connected")
+            coro = self._client.publish(topic, message)
```
Lash-L: Does the coroutine need to be made within the lock?

roborock/mqtt/roborock_session.py
```diff
@@ -0,0 +1,213 @@
...
+                timeout=params.timeout,
+                logger=_MQTT_LOGGER,
+            ) as client:
+                _LOGGER.debug("Connected to MQTT broker")
+                # Re-establish any existing subscriptions
+                async with self._client_lock:
+                    self._client = client
+                    for topic in self._listeners:
+                        _LOGGER.debug("Re-establising subscription to topic %s", topic)
+                        await client.subscribe(topic)
```
Lash-L, line 166: Should we try except this? Incase we fail to subscribe to a specific topic (not sure how likely this is - but i think theoretically possible)

We wouldn't want to fail entirely would we? Not sure how much it matters though as that would add some complexity.

Lash-L (commented):
Looks Amazing!

roborock/mqtt/roborock_session.py
allenporter: My assumption is that this will fall through and set `self._healthy = False` at the exit for this function. I've gone ahead an added a test for the `connected` attribute based on this prompting (though not clear it actually exercises that case)

roborock/mqtt/roborock_session.py
allenporter: Good point. If you have a bunch of devices i can see this could be sending a bunch of requests in a row and it would be a shame to lose the whole connection.

If this fails, I think it needs to keep retrying which might be the complicated part.  We would need to keep retrying in the background with additional backoff, essentially a mirror of the backoff happening for the primary client.


I'm wondering if perhaps that needs to be restructured, pulled out into a separate object for managing the backoff state. Maybe something like a `DeviceMqttSession` object.

allenporter: How about I extract that in a followup?

roborock/mqtt/roborock_session.py
allenporter: I think your version is more clear here, updated.

roborock/mqtt/roborock_session.py
allenporter: My assumption was that it would not be expensive since it didn't await, but maybe its more clear to try not to do it at all. My worry was that `self._client` would get reset or shutdown, so maybe its ok to just make a copy of the reference to the varaible.

The worst case here is the client is shutting down in paralle with the publish call happening which should be fine and just cause it to fail.

Lash-L (approved):
honestly looks great. Happy to merge this in and then make any needed adjustments with the roborock device when that is added. It's possible some flaws may pop up during implementation

roborock/mqtt/roborock_session.py
Lash-L: I think either way is fine, this may be a bit safer, and like you said it shouldn't be expensive.

I was just curious about the why so I knew going forward.

roborock/mqtt/roborock_session.py
Lash-L: Ah I guess in my head i have always viewed pass as essentially a return. I only ever user it in empty functions as a placeholder. I didn't realize it was basically the same as '...'

PR #397, opened 2025-06-29
chore: Minor refactoring creating functions for transforming bytes

allenporter:
This is in preparation for sharin with new device/mqtt code.

### DIFF
--- a/roborock/protocol.py
+++ b/roborock/protocol.py
@@ -359,3 +359,56 @@ class _Parser:
 
 MessageParser: _Parser = _Parser(_Messages, True)
 BroadcastParser: _Parser = _Parser(_BroadcastMessage, False)
+
+
+Decoder = Callable[[bytes], list[RoborockMessage]]
+Encoder = Callable[[RoborockMessage], bytes]
+
+
+def create_mqtt_decoder(local_key: str) -> Decoder:
+    """Create a decoder for MQTT messages."""
+
+    def decode(data: bytes) -> list[RoborockMessage]:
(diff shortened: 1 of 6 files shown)


roborock/protocol.py
```diff
@@ -359,3 +359,56 @@ def build(
...
+    # It is defined outside the decode function to maintain state across calls.
+    buffer: bytes = b""
+
+    def decode(bytes: bytes) -> list[RoborockMessage]:
+        """Parse the given data into Roborock messages."""
+        nonlocal buffer
+        buffer += bytes
+        parsed_messages, remaining = MessageParser.parse(buffer, local_key=local_key)
+        buffer = remaining
+        return parsed_messages
```
Lash-L, line 402: In your testing have you found that the buffer is often being used?

Lash-L (approved):
Looks good - I think this is a good code cleanup step - makes it a bit clearer to see what is happening opposed to the previous function calls

roborock/protocol.py
allenporter: I have not looked at this, just assumed it was needed. Intuitively it makes sense to me that tcp streams can arrive in chunks that may not align to the message. Maybe will remember to track explicitly how often this is used. (I also worry about errors if messages are not valid, what happens, e.g. can it recover?

allenporter: (Could explain why devices go bad and don't recover)

allenporter: I would prefer to leave this unresolved for history, but needs to be resolved to merge)

PR #399, opened 2025-06-30
feat: Add a DeviceManager to perform discovery

allenporter:
Add a device manager that will query home data then create a Device object for each discovered device. The API exposes the home API such that it can be wrapped by the caller with a caching layer.

The device object is a placeholder and will likely be changed in the future to support capabilities.

### DIFF
--- a/roborock/devices/README.md
+++ b/roborock/devices/README.md
@@ -7,7 +7,9 @@ This page documents the full lifecycle of device discovery across Cloud and Netw
 ### Login
 
 - Login can happen with either email and password or email and sending a code. We
-  currently prefer email with sending a code.
+  currently prefer email with sending a code -- however the roborock no longer
+  supports this method of login. In the future we may want to migrate to password
+  if this login method is no longer supported.
 - The Login API provides a `userData` object with information on connecting to the cloud APIs
 - This `rriot` data contains per-session information, unique each time you login.
   - This contains information used to connect to MQTT
@@ -15,7 +17,9 @@ This page documents the full lifecycle of device discovery across Cloud and Netw
 
 ## Home Data
 
-The `HomeData` includes information about the various devices in the home. We use `v3`.
+The `HomeData` includes information about the various devices in the home. We use `v3`
+and it is notable that if devices don't show up in the `home_data` response it is likely
+that a newer version of the API should be used.
 
 - `products`: This is a list of all of the products you have on your account. These objects are always the same (i.e. a s7 maxv is always the exact same.)
   - It only shows the products for devices available on your account
--- a/roborock/devices/README.md
+++ b/roborock/devices/README.md
@@ -1,0 +1,94 @@
+# Roborock Device Discovery
+
+This page documents the full lifecycle of device discovery across Cloud and Network.
+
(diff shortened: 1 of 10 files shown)


roborock/devices/README.md
```diff
@@ -0,0 +1,94 @@
+# Roborock Device Discovery
+
+This page documents the full lifecycle of device discovery across Cloud and Network.
+
+## Init account setup
+
+### Login
+
+- Login can happen with either email and password or email and sending a code. We
+  currently prefer email with sending a code.
```
Lash-L: May be worth noting that the roborock app no longer supports code login. In the future we may have to migrate to password if they fully depreciate that login method.

roborock/devices/README.md
```diff
@@ -0,0 +1,94 @@
...
+- The Login API provides a `userData` object with information on connecting to the cloud APIs
+- This `rriot` data contains per-session information, unique each time you login.
+  - This contains information used to connect to MQTT
+  - You get an `-eu` suffix in the API URLs if you are in the eu and `-us` if you are in the us
+
+## Home Data
+
+The `HomeData` includes information about the various devices in the home. We use `v3`.
```
Lash-L: May be worth noting that if devices don't show up in the home_data response, you likely need to use a newer version of the home_data api.

But maybe that is intuitively clear

roborock/devices/device.py
```diff
@@ -0,0 +1,65 @@
...
+    def device_version(self) -> str:
+        """Return the device version.
+
+        At the moment this is a simple check against the product version (pv) of the device
+        and used as a placeholder for upcoming functionality for devices that will behave
+        differently based on the version and capabilities.
+        """
+        if self._device_info.pv == "1.0":
+            return DeviceVersion.V1
+        elif self._device_info.pv in ["A1", "A01"]:
```
Lash-L: I don't think pv can be either one of those things - have you seen otherwise?

tests/devices/test_device_manager.py
```diff
@@ -0,0 +1,80 @@
...
+    assert devices == []
+
+
+async def test_with_device() -> None:
+    """Test the DeviceManager created with devices returned from the API."""
+    device_manager = await create_device_manager(USER_DATA, mock_home_data)
+    devices = await device_manager.get_devices()
+    assert [(device.duid, device.name, device.device_version) for device in devices] == [
+        ("abc123", "Roborock S7 MaxV", DeviceVersion.V1),
+    ]
```
Lash-L: Seems overly complicated? Just a placeholder for when you add multiple devices?

roborock/devices/README.md
allenporter: This had not occurred to me, so glad you mentioned it. Updated.

roborock/devices/device.py
allenporter: AI suggested.  How about this then: https://github.com/home-assistant/core/blob/23c304fc75b3da89e3e2f98a8af89399a4c646d6/homeassistant/components/roborock/__init__.py#L236C22-L236C25

Lash-L (approved):

PR #406, opened 2025-07-01
Make error code consistent for dock and robot

amaximus:
When everything is ok the robot reports 'none' while the dock 'ok', while they both should report 'none' reflecting no errors.


allenporter (commented):
Appreciate consistency, though I don't know that it's worth a breaking API change here.

PR #409, opened 2025-07-01
feat: Update device manager and device to establish an MQTT subscription

allenporter:
Update the device manager and device to establish an Mqtt "channel" which is the device-specific subscription in the mqtt session. The channel will be used for decoding messages and sending/receiving RPCs.

Lots of future work needed here including:
- Handle non-mqtt devices
- Add message decoding 
- Add message encoding and receiving of RPCs
- Local connections

### DIFF
--- a/roborock/devices/device_manager.py
+++ b/roborock/devices/device_manager.py
@@ -1,5 +1,6 @@
 """Module for discovering Roborock devices."""
 
+import asyncio
 import logging
 from collections.abc import Awaitable, Callable
 
@@ -56,6 +57,7 @@ class DeviceManager:
         device_products = home_data.device_products
         _LOGGER.debug("Discovered %d devices %s", len(device_products), home_data)
 
+        # These are connected serially to avoid overwhelming the MQTT broker
         new_devices = {}
         for duid, (device, product) in device_products.items():
             if duid in self._devices:
@@ -77,11 +79,10 @@ class DeviceManager:
 
     async def close(self) -> None:
         """Close all MQTT connections and clean up resources."""
-        for device in self._devices.values():
-            await device.close()
+        tasks = [device.close() for device in self._devices.values()]
         self._devices.clear()
-        if self._mqtt_session:
-            await self._mqtt_session.close()
+        tasks.append(self._mqtt_session.close())
+        await asyncio.gather(*tasks)
 
 
 def create_home_data_api(email: str, user_data: UserData) -> HomeDataApi:
--- a/roborock/devices/device_manager.py
+++ b/roborock/devices/device_manager.py
@@ -113,7 +113,7 @@ async def create_device_manager(user_data: UserData, home_data_api: HomeDataApi)
     mqtt_session = await create_mqtt_session(mqtt_params)
 
     def device_creator(device: HomeDataDevice, product: HomeDataProduct) -> RoborockDevice:
-        mqtt_channel = MqttChannel(mqtt_session, device.duid, user_data.rriot, mqtt_params)
+        mqtt_channel = MqttChannel(mqtt_session, device.duid, device.local_key, user_data.rriot, mqtt_params)
         return RoborockDevice(user_data, device, product, mqtt_channel)
 
     manager = DeviceManager(home_data_api, device_creator, mqtt_session=mqtt_session)
--- a/roborock/devices/device_manager.py
+++ b/roborock/devices/device_manager.py
@@ -10,7 +10,12 @@ from roborock.containers import (
     UserData,
 )
 from roborock.devices.device import RoborockDevice
+from roborock.mqtt.roborock_session import create_mqtt_session
+from roborock.mqtt.session import MqttSession
+from roborock.protocol import create_mqtt_params
 from roborock.web_api import RoborockApiClient
+
+from .mqtt_channel import MqttChannel
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


roborock/devices/device_manager.py
```diff
@@ -59,6 +75,14 @@ async def get_devices(self) -> list[RoborockDevice]:
         """Get all discovered devices."""
         return list(self._devices.values())
 
+    async def close(self) -> None:
+        """Close all MQTT connections and clean up resources."""
+        for device in self._devices.values():
+            await device.close()
```
Lash-L: Make sense for us to gather this and the next one?

roborock/devices/mqtt_channel.py
```diff
@@ -0,0 +1,44 @@
...
+    This currently only supports listening to messages and does not yet
+    support RPC functionality.
+    """
+
+    def __init__(self, mqtt_session: MqttSession, duid: str, rriot: RRiot, mqtt_params: MqttParams):
+        self._mqtt_session = mqtt_session
+        self._duid = duid
+        self._rriot = rriot
+        self._mqtt_params = mqtt_params
+        self._unsub: Callable[[], None] | None = None
```
allenporter:
```suggestion
```

roborock/devices/mqtt_channel.py
```diff
@@ -0,0 +1,43 @@
...
+    @property
+    def _subscribe_topic(self) -> str:
+        """Topic to receive responses from the device."""
+        return f"rr/m/o/{self._rriot.u}/{self._mqtt_params.username}/{self._duid}"
+
+    async def subscribe(self, callback: Callable[[bytes], None]) -> Callable[[], None]:
+        """Subscribe to the device's response topic.
+
+        The callback will be called with the message payload when a message is received.
+        If already subscribed, raises ValueError.
```
allenporter:
```suggestion
```

roborock/devices/mqtt_channel.py
```diff
@@ -0,0 +1,43 @@
...
+
+        Returns a callable that can be used to unsubscribe from the topic.
+        """
+        if self._unsub:
+            raise ValueError("Already subscribed to the response topic")
```
allenporter:
```suggestion
```

roborock/devices/device_manager.py
```diff
@@ -34,21 +39,32 @@ def __init__(
...
 
-        self._devices = {
-            duid: self._device_creator(device, product) for duid, (device, product) in device_products.items()
-        }
+        new_devices = {}
+        for duid, (device, product) in device_products.items():
+            if duid in self._devices:
+                continue
+            new_device = self._device_creator(device, product)
+            await new_device.connect()
```
Lash-L, line 66: Should we gather? Or are you attempting to help prevent any kind of mqtt load

roborock/devices/mqtt_channel.py
```diff
@@ -0,0 +1,115 @@
...
+                except Exception as e:
+                    _LOGGER.exception("Uncaught error in message handler callback: %s", e)
+
+        return await self._mqtt_session.subscribe(self._subscribe_topic, message_handler)
+
+    async def _resolve_future_with_lock(self, message: RoborockMessage) -> None:
+        """Resolve waiting future with proper locking."""
+        if (request_id := message.get_request_id()) is None:
+            _LOGGER.debug("Received message with no request_id")
+            return
```
Lash-L, line 76: In the future we will want to allow this to do a separate callback for protocol data?

Data that is auto sent as updates? I think it wont have a request id - or at the very least it wont have a request id that we have a matching id for. But that can come in the future, i just wanted to call it out.

roborock/devices/mqtt_channel.py
```diff
@@ -0,0 +1,115 @@
...
+                try:
+                    callback(message)
+                except Exception as e:
+                    _LOGGER.exception("Uncaught error in message handler callback: %s", e)
+
+        return await self._mqtt_session.subscribe(self._subscribe_topic, message_handler)
+
+    async def _resolve_future_with_lock(self, message: RoborockMessage) -> None:
+        """Resolve waiting future with proper locking."""
+        if (request_id := message.get_request_id()) is None:
```
Lash-L, line 74: I could see a debug line here being helpful

Lash-L (commented):
Looks good - I'll be curious to see the next iteration. Just a couple of comments

roborock/devices/device_manager.py
allenporter: i think connecting serially to avoid overloading mqtt makes sense., but if we see latency is too high we can revisit. Added a comment for now.

roborock/devices/device_manager.py
allenporter: Good idea, done.

roborock/devices/mqtt_channel.py
allenporter: Yeah, my thinking was:
- `subscribe` takes a callback. That is used for all messages received and the device can listen to them to update it's local state.
- the rpc version also has a callback to associate with the original request. that can be used to reply to the caller

An RPC calls both. A message with no incoming RPC only calls the first one. My thinking is this can help separate responsibilities a bit.

Does that cover the use case you have in mind?

roborock/devices/mqtt_channel.py
allenporter: ok, added on line 64 before calling this. Does that work?

roborock/devices/mqtt_channel.py
Lash-L: Yes I think it does!

PR #410, opened 2025-07-04
feat: Add a local channel, similar to the MQTT channel

allenporter:
Add a local channel, similar to the MQTT channel. This is not yet used by the device or device manager.

### DIFF
--- a/roborock/devices/local_channel.py
+++ b/roborock/devices/local_channel.py
@@ -1,0 +1,143 @@
+"""Module for communicating with Roborock devices over a local network."""
+
+import asyncio
+import logging
+from collections.abc import Callable
+from dataclasses import dataclass
+from json import JSONDecodeError
+
+from roborock.exceptions import RoborockConnectionException, RoborockException
+from roborock.protocol import Decoder, Encoder, create_local_decoder, create_local_encoder
+from roborock.roborock_message import RoborockMessage
+
+_LOGGER = logging.getLogger(__name__)
+_PORT = 58867
+
+
+@dataclass
+class _LocalProtocol(asyncio.Protocol):
+    """Callbacks for the Roborock local client transport."""
+
+    messages_cb: Callable[[bytes], None]
+    connection_lost_cb: Callable[[Exception | None], None]
+
+    def data_received(self, data: bytes) -> None:
+        """Called when data is received from the transport."""
+        self.messages_cb(data)
+
+    def connection_lost(self, exc: Exception | None) -> None:
+        """Called when the transport connection is lost."""
+        self.connection_lost_cb(exc)
+
+
+class LocalChannel:
+    """Simple RPC-style channel for communicating with a device over a local network.
+
+    Handles request/response correlation and timeouts, but leaves message
+    format most parsing to higher-level components.
+    """
+
+    def __init__(self, host: str, local_key: str):
+        self._host = host
+        self._transport: asyncio.Transport | None = None
+        self._protocol: _LocalProtocol | None = None
+        self._subscribers: list[Callable[[RoborockMessage], None]] = []
+        self._is_connected = False
+
+        # RPC support
+        self._waiting_queue: dict[int, asyncio.Future[RoborockMessage]] = {}
+        self._decoder: Decoder = create_local_decoder(local_key)
+        self._encoder: Encoder = create_local_encoder(local_key)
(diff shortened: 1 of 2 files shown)


roborock/devices/local_channel.py
```diff
@@ -0,0 +1,143 @@
...
+        try:
+            self._transport, self._protocol = await loop.create_connection(lambda: protocol, self._host, _PORT)
+            self._is_connected = True
+        except OSError as e:
+            raise RoborockConnectionException(f"Failed to connect to {self._host}:{_PORT}") from e
+
+    async def close(self) -> None:
+        """Disconnect from the device."""
+        if self._transport:
+            self._transport.close()
```
Lash-L, line 70: Maybe a warning if not self._transport? 


That way if we have that happening we can identify it?

roborock/devices/local_channel.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(self, host: str, local_key: str):
+        self._host = host
+        self._transport: asyncio.Transport | None = None
+        self._protocol: _LocalProtocol | None = None
+        self._subscribers: list[Callable[[RoborockMessage], None]] = []
+        self._is_connected = False
+
+        # RPC support
+        self._waiting_queue: dict[int, asyncio.Future[RoborockMessage]] = {}
+        self._decoder: Decoder = create_local_decoder(local_key)
```
Lash-L, line 49: I guess something i didn't ask before.

How do you see these objects adapting to best support A01 and B01 devices? Different decoder/encoder functions?

Logic within the decoding to determine what to do?

roborock/devices/local_channel.py
```diff
@@ -0,0 +1,143 @@
...
+                try:
+                    callback(message)
+                except Exception as e:
+                    _LOGGER.exception("Uncaught error in message handler callback: %s", e)
+
+    def _connection_lost(self, exc: Exception | None) -> None:
+        """Handle connection loss."""
+        _LOGGER.warning("Connection lost to %s", self._host, exc_info=exc)
+        self._transport = None
+        self._is_connected = False
```
Lash-L, line 94: Do we want to attempt to reconnect? Or would that happen further upstream?

Lash-L:
Also seems like tests are still timing out unfortuantely

roborock/devices/local_channel.py
allenporter: Yeah, I think if devices have different behavior then these should just be different encoder/decoer functions passed in. It's kind of annoying to pass two things, so i could also see making it a single object for the specific device or channel.

roborock/devices/local_channel.py
allenporter: I wasn't sure yet, so just ommited for now.

I have two thoughts for options:
(1)  The device can handle it, whenever it sends a command.  First check if connected, then reconnect, etc. Home assistnat, for example, has a nice poll mechanism that can backoff on failures, so if the caller is polling (nicely) that could trigger something that would cause a reconnect if the connection was previously loss.
(2) start a background task for this channel responsible for reconnecting with backoff, similar to what mqtt session is doing

Very  open to input/discussion

roborock/devices/local_channel.py
allenporter: Done.

Lash-L (approved):

roborock/devices/local_channel.py
Lash-L: Back off would certainly be nice...

I think that's the route I would lean BUT it seems to cripple the library for the sake of the integration. But maybe we could work around that by having like a check connection and reconnect function in the library that the integration just uses

A task may be the better option

We also have to think about the fact that no matter what this is happening at least once a day during the 3am disconnect (would be nice if we could get it to stop showing up in HA - but may not be feasible/accurate(i.e. hiding the reality for the sake of looking nice


Those are my thoughts! Sorry if they are all over the place, middle of the night feed rn

allenporter: Thanks, that is very helpful. Background task seems like the way to go then: The point about managing the health of the connection knowing that it can do odd things like disconnect at 3am makes sense to me, such that we'd only consider the device unavaailble if disconnected for a significant amount of time.  We can just have backoff in here same as the mqtt channel and manage it internally just fine.

PR #411, opened 2025-07-05
Update tests to improve CI logging to better diagnose timeout

allenporter:
Remove tests that often timeout on CI. This will result in lower test coverage, but probably not worth it given the flakes.


Lash-L:
Looks to still flake unfortunately

Lash-L (approved):

PR #413, opened 2025-07-14
chore: Create module for v1 request encoding

allenporter:
Extract the v1 request encoding into a separate module that will be shared with the new device library.

### DIFF
--- a/roborock/protocols/v1_protocol.py
+++ b/roborock/protocols/v1_protocol.py
@@ -13,12 +13,6 @@ from roborock.roborock_message import MessageRetry, RoborockMessage, RoborockMes
 from roborock.roborock_typing import RoborockCommand
 from roborock.util import get_next_int
 
-# All mqtt commands are sent securely. Only local commands in this list are secured.
-COMMANDS_SECURED = {
-    RoborockCommand.GET_MAP_V1,
-    RoborockCommand.GET_MULTI_MAP,
-}
-
 CommandType = RoborockCommand | str
 ParamsType = list | dict | int | None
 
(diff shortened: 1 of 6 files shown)


roborock/protocols/v1_protocol.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def create_local_payload_encoder(security_data: SecurityData) -> Callable[[CommandType, ParamsType], RoborockMessage]:
+    """Create a payload encoder for V1 commands over local connection."""
+
+    def _get_payload(method: CommandType, params: ParamsType) -> RoborockMessage:
+        """Build the payload for a V1 command."""
+        request = RequestMessage(method=method, params=params)
+        is_secured = request.method in COMMANDS_SECURED
+        payload = request.as_payload(security_data if is_secured else None)
```
Lash-L: I want to say that secured commands are only possible via cloud.

But maybe they are possible local - they just only give cloud responses.

tests/devices/test_v1_protocol.py
Lash-L, line 1: empty file?

Lash-L (dismissed):
looks good, primarily just moving things so not much to say

roborock/protocols/v1_protocol.py
allenporter: Is another way to say this that: These commands `GET_MAP_V1` should only ever be sent to cloud?  And we can remove the security data entirely for local commands?

allenporter: I've simplified this by removing the code related to security data for local payload encoding.

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -21,26 +22,16 @@ def __init__(self, device_data: DeviceData, queue_timeout: int = 4):
...
-            timestamp=timestamp, protocol=request_protocol, payload=payload, message_retry=message_retry
-        )
-
     async def _send_command(
         self,
         method: RoborockCommand | str,
         params: list | dict | int | None = None,
     ):
-        roborock_message = self.build_roborock_message(method, params)
+        if method in CLOUD_REQUIRED:
```
allenporter, line 30: This is now a breaking change to just hard fail when trying to do this.

PR #415, opened 2025-07-14
Add Saros 10R modes

starkillerOG:


starkillerOG:
Yes I already performed a firmware update. I should be fully up to date.

I did not specifically test this code with the Saros 10R, but can easily do that if required.
I just checked all the modes using the diagnostic info.
But I can easily test it if required.

I will fix the linting tomorrow, no problem.

roborock/code_mappings.py
```diff
@@ -438,6 +458,17 @@ class RoborockMopIntensityS8MaxVUltra(RoborockMopIntensityCode):
...
 
 
+class RoborockMopIntensitySaros10R(RoborockMopIntensityCode):
+    off = 200
+    low = 201
+    medium = 202
+    high = 203
+    custom = 204
+    extreme = 250
+    vac_followed_by_mop = 235
```
Kranzes, line 468: Hi, I am getting similar codes on my Saros 10 (non R) but this one I don't get, are you sure it's also under the `water_box_mode` field?

starkillerOG: Yes, I checked quite throughly.
It happens when selecting the mode "vac followed by mop" instead of "vacuum & mop", then don't change any of the settings like water flow, otherwise it goes back to "vacuum & mop"

Kranzes: if water_box_mode is set to 235 when in vac followed by mop, what does changing the mop intensity to "extreme" (for example) (while in followed by mop mode) change actually?

starkillerOG: It will go back to the "vacuum & mop" and apply the "extreme" setting for the water flow.
(I think roborock could have made this better or more clear in their UI of the app...)

roborock/code_mappings.py
```diff
@@ -438,6 +458,17 @@ class RoborockMopIntensityS8MaxVUltra(RoborockMopIntensityCode):
     custom_water_flow = 207
 
 
+class RoborockMopIntensitySaros10R(RoborockMopIntensityCode):
+    off = 200
+    low = 201
+    medium = 202
+    high = 203
```
Kranzes, line 465: Are these actually called "low" "medium" and "high" for you in the Roborock app? For me it's "mild" "standard" and "Intense".

roborock/code_mappings.py
Kranzes: I don't know what it is for my Saros 10, but 235 doesn't work and throws an error.

roborock/code_mappings.py
starkillerOG: Screenshot_20250721_215417_Roborock
Screenshot_20250721_215412_Roborock
Screenshot_20250721_215405_Roborock
Screenshot_20250721_215359_Roborock
Screenshot_20250721_215354_Roborock
Screenshot_20250721_215337_Roborock

starkillerOG:
@Lash-L the tests are now passing, could you merge?

Lash-L:
> @Lash-L the tests are now passing, could you merge?

Sorry I've been out of town two weekends in a row. Will do now

Lash-L:
@allenporter what is your general level of confidence that a HA version bump would not break everything? Most changes have been isolated or moves

PR #416, opened 2025-07-16
feat: Add a v1 protocol channel that can send messages across MQTT or Local connections, preferring local

allenporter:
Add a V1Channel class that wraps the MQTT and Local channels and handles getting the network info from mqtt and decidign which channel is available. The local connection does not handle retries yet, which will be fixed in a follow up change.

Most of the PR is for tests exercising corner cases. This PR was heavily coded by Co-pilot.

### DIFF
--- a/roborock/devices/device_manager.py
+++ b/roborock/devices/device_manager.py
@@ -10,13 +10,13 @@ from roborock.containers import (
     HomeDataProduct,
     UserData,
 )
-from roborock.devices.device import RoborockDevice
+from roborock.devices.device import DeviceVersion, RoborockDevice
 from roborock.mqtt.roborock_session import create_mqtt_session
 from roborock.mqtt.session import MqttSession
 from roborock.protocol import create_mqtt_params
 from roborock.web_api import RoborockApiClient
 
-from .mqtt_channel import MqttChannel
+from .v1_channel import create_v1_channel
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -114,8 +114,16 @@ async def create_device_manager(user_data: UserData, home_data_api: HomeDataApi)
     mqtt_session = await create_mqtt_session(mqtt_params)
 
     def device_creator(device: HomeDataDevice, product: HomeDataProduct) -> RoborockDevice:
-        mqtt_channel = MqttChannel(mqtt_session, device.duid, device.local_key, user_data.rriot, mqtt_params)
-        return RoborockDevice(user_data, device, product, mqtt_channel)
+        # Check device version and only support V1 for now
+        if device.pv != DeviceVersion.V1.value:
+            raise NotImplementedError(
+                f"Device {device.name} has version {device.pv}, but only V1 devices "
+                f"are supported through the unified interface. Use the legacy "
+                f"RoborockMqttClientV1 or RoborockLocalClientV1 for this device."
+            )
+        # Create V1 channel that handles both MQTT and local connections
+        v1_channel = create_v1_channel(user_data, mqtt_params, mqtt_session, device)
+        return RoborockDevice(user_data, device, product, v1_channel)
 
     manager = DeviceManager(home_data_api, device_creator, mqtt_session=mqtt_session)
     await manager.discover_devices()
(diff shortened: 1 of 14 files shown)


roborock/devices/device_manager.py
```diff
@@ -114,8 +114,16 @@ async def create_device_manager(user_data: UserData, home_data_api: HomeDataApi)
...
 
     def device_creator(device: HomeDataDevice, product: HomeDataProduct) -> RoborockDevice:
-        mqtt_channel = MqttChannel(mqtt_session, device.duid, device.local_key, user_data.rriot, mqtt_params)
-        return RoborockDevice(user_data, device, product, mqtt_channel)
+        # Check device version and only support V1 for now
+        if device.pv != DeviceVersion.V1.value:
+            raise NotImplementedError(
+                f"Device {device.name} has version {device.pv}, but only V1 devices "
+                f"are supported through the unified interface. Use the legacy "
+                f"RoborockMqttClientV1 or RoborockLocalClientV1 for this device."
```
Lash-L: not super important but this is inaccurate. don't use v1 client for non v1 devices

roborock/devices/v1_channel.py
```diff
@@ -0,0 +1,215 @@
...
+        This will establish MQTT connection first, and also attempt to set up
+        local connection if possible.
+        """
+
+        if self._mqtt_unsub:
+            raise ValueError("Already connected to the device")
+        self._callback = callback
+
+        # First establish MQTT connection
+        self._mqtt_unsub = await self._mqtt_channel.subscribe(self._on_mqtt_message)
```
Lash-L, line 93: Is it worth doing error handling here? Or are we just doing it above subscribe?

roborock/devices/v1_channel.py
```diff
@@ -0,0 +1,215 @@
...
+        falling back to MQTT if needed.
+        """
+        _LOGGER.debug("Sending command: %s, params=%s", method, params)
+        if self._local_channel:
+            try:
+                return await self._send_local_decoded_command(method, response_type=response_type, params=params)
+            except RoborockException as e:
+                _LOGGER.info("Local command failed, falling back to MQTT: %s", e)
+
+        return await self._send_mqtt_decoded_command(method, response_type=response_type, params=params)
```
Lash-L, line 157: I'm wondering if this should only be an else. and not always happen

If the user has a invalid command or the vacuum is truly off the following will happen:

1) Send the message locally -> Wait for the response
2) Exception!
3) Send the message via cloud -> wait for the response

Maybe it's fine - but in my mind I think it makes sense to only fall back to mqtt decoded if a) self._lcaol_channel is None or b) the user specifically askes to send a mqtt message or c) We know for a fact this message has to be done via cloud.

roborock/devices/v1_channel.py
```diff
@@ -0,0 +1,215 @@
...
+
+    async def _send_mqtt_raw_command(self, method: CommandType, params: ParamsType | None = None) -> dict[str, Any]:
+        """Send a raw command without response handling."""
```
Lash-L: I think response handling may be inaccurate here.

There is still decoding going on - we are getting back json - it's just not in the specific object. Maybe without response parsing would be clearer?

Lash-L:
Sorry that took me so long - should be faster for any follow ups

roborock/devices/v1_channel.py
allenporter: My current thinking is to propagate to the caller. I've added pydoc explaining the expectations.  (We can change, though i take your main point as it's not clear whats supposed to happen)

roborock/devices/device_manager.py
allenporter: Fixed.

roborock/devices/v1_channel.py
allenporter: Changed to be an else. You're right, we should keep connection health as a separate concept.

roborock/devices/v1_channel.py
allenporter: Agreed, re-worded.

Lash-L (approved):

PR #417, opened 2025-07-16
chore: Move a01 encoding and decoding to a separate module

allenporter:
Update the a01 encoding and decoding to live in a separate module that can be shared with the new device code. Increase test coverage of the encoding and decoding.

This removes the caching since it was not implemented, and it can be re-added in the new device library.

### DIFF
--- a/roborock/version_a01_apis/roborock_client_a01.py
+++ b/roborock/version_a01_apis/roborock_client_a01.py
@@ -140,15 +140,16 @@ class RoborockClientA01(RoborockClient, ABC):
                 try:
                     data_points = decode_rpc_response(message)
                 except RoborockException as err:
-                    self._logger.error("Failed to decode message %s: %s", message, err)
+                    self._logger.debug("Failed to decode message: %s", err)
                     continue
                 for data_point_number, data_point in data_points.items():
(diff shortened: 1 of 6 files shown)


roborock/version_a01_apis/roborock_client_a01.py
```diff
@@ -43,111 +40,121 @@
 _LOGGER = logging.getLogger(__name__)
 
 
-@dataclasses.dataclass
-class A01ProtocolCacheEntry:
-    post_process_fn: Callable
-    value: typing.Any | None = None
-
-
 # Right now this cache is not active, it was too much complexity for the initial addition of dyad.
```
Lash-L: I think this comment is no longer accurate

Lash-L (approved):
Seems pretty straightforward. Nice change. I don't have a a01 device to explicitly test it, but test look good

roborock/version_a01_apis/roborock_client_a01.py
allenporter: Removed.

allenporter:
Comments were addressed, this is ready for approval/merge.

PR #419, opened 2025-07-27
feat: Add Saros 10 code mappings

Kranzes:
The codes are equal to #415, but the names are different. I also can't confirm the `vac_followed_by_mop` code so I havent added it yet.


Kranzes:
Do we have any specific order for the classes in the code? It got a bit messed up during a rebase...

Lash-L (approved):

PR #424, opened 2025-08-05
chore: Refactor some of the internal channel details used by the device.

allenporter:
Refactor some of the internal channel details used by the device by separating V1 API connection logic from encoding logic

This creates an "RPC Channel" to separate the logic for establishing connections from doing the "RPC" payload request encoding and response parsing for sending `RoborockCommand` commands.    The existing `send_command` on the channel has been renamed to `send_message` to be more clear since it is sending a `RoborockMessage`. The fallback logic is now behind the RPC interface. 

The motivation for this is because we will likely try to push much of the device connection logic into the background. I'm working on a separate PR that adding A01 protocol and it seems like separating the connection logic from the RPC sending is helpful since when interacting with the devices you only care about sending RPCs.

### DIFF
--- a/roborock/devices/v1_rpc_channel.py
+++ b/roborock/devices/v1_rpc_channel.py
@@ -1,0 +1,158 @@
+"""V1 Rpc Channel for Roborock devices.
+
+This is a wrapper around the V1 channel that provides a higher level interface
+for sending typed commands and receiving typed responses. This also provides
+a simple interface for sending commands and receiving responses over both MQTT
+and local connections, preferring local when available.
+"""
+
+import logging
+from collections.abc import Callable
+from typing import Any, Protocol, TypeVar, overload
+
+from roborock.containers import RoborockBase
+from roborock.protocols.v1_protocol import (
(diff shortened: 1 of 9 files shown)


roborock/devices/v1_rpc_channel.py
```diff
@@ -0,0 +1,148 @@
...
+        self,
+        method: CommandType,
+        *,
+        params: ParamsType = None,
+    ) -> Any:
+        """Send a raw command and return the decoded response. Must be implemented by subclasses."""
+        raise NotImplementedError
+
+
+class CombinedChannel(BaseV1RpcChannel):
```
Lash-L: Should this be labeled CombinedV1Channel?

allenporter: Updated. I named it `CombinedV1RpcChannel` to be consistent with `V1RpcChannel` and `BaseV1RpcChannel`. Also `PayloadEncodedV1Channel` is now `PayloadEncodedV1RpcChannel`.

PR #425, opened 2025-08-07
feat: Support both a01 and v1 device types with traits

allenporter:


roborock/devices/device.py
```diff
@@ -4,101 +4,72 @@
...
-        user_data: UserData,
         device_info: HomeDataDevice,
-        product_info: HomeDataProduct,
-        v1_channel: V1Channel,
+        channel: Channel,
+        traits: list[Trait],
     ) -> None:
         """Initialize the RoborockDevice.
 
         The device takes ownership of the V1 channel for communication with the device.
```
Lash-L: generic class but comment talks about V1

roborock/devices/device.py
```diff
@@ -111,10 +82,7 @@ def _on_message(self, message: RoborockMessage) -> None:
...
-        """Get the current status of the device.
-
-        This is a placeholder command and will likely be changed/moved in the future.
-        """
-        status_type: type[Status] = ModelStatus.get(self._product_info.model, S7MaxVStatus)
-        return await self._v1_channel.rpc_channel.send_command(RoborockCommand.GET_STATUS, response_type=status_type)
+    @property
+    def traits(self) -> Mapping[str, Trait]:
+        """Return the traits of the device."""
+        return MappingProxyType(self._trait_map)
```
Lash-L, line 88: Why can't this just stay as a dict as is? I don't really know what MappingProxyType is

roborock/devices/traits/status.py
```diff
@@ -0,0 +1,48 @@
...
+    """Unified Roborock device class with automatic connection setup."""
+
+    name = "status"
+
+    def __init__(self, product_info: HomeDataProduct, rpc_channel: V1RpcChannel) -> None:
+        """Initialize the RoborockDevice.
+
+        The device takes ownership of the V1 channel for communication with the device.
+        Use `connect()` to establish the connection, which will set up the appropriate
+        protocol channel. Use `close()` to clean up all connections.
```
Lash-L: stale comment

roborock/devices/v1_channel.py
```diff
@@ -63,6 +64,11 @@ def __init__(
         self._callback: Callable[[RoborockMessage], None] | None = None
         self._networking_info: NetworkInfo | None = None
 
+    @property
+    def is_connected(self) -> bool:
+        """Return whether MQTT connection is available."""
```
Lash-L:
```suggestion
        """Return whether any connection is available."""
```

roborock/devices/device.py
allenporter: When this returns a `dict` its actually mutable so a caller could theoretically change it. It may be overkill here...

roborock/devices/traits/status.py
```diff
@@ -0,0 +1,43 @@
...
+    """Unified Roborock device class with automatic connection setup."""
+
+    name = "status"
+
+    def __init__(self, product_info: HomeDataProduct, rpc_channel: V1RpcChannel) -> None:
+        """Initialize the StatusTrait."""
+        self._product_info = product_info
+        self._rpc_channel = rpc_channel
+
+    async def get_status(self) -> Status:
```
Lash-L, line 37: So does each trait have its own unique functions? This may negate the need for the traits property, as iterating through them wouldn't be helpful as you need to do different things Per trait

Lash-L (approved):

roborock/devices/traits/status.py
Lash-L: although I guess the traits() property just gives a mapping, so you can still do everything specifically there as well.

allenporter: My assumption is traits will have multiple functions of related properties. However, maybe that is not the case, so i take your point that if every trait really is one thing then its just a grab bag of functions?

allenporter: Maybe i take your point as traits can be flattened properties on the device, which i agree with. I was assuming we'd go to something like this:

```
if device.status:
    print(await device.status.get_status())
```

Lash-L: Most of the time they come down to two things:

Get the latest version of myself

change myself or perform an action

sometimes there are multiple different ways you can make changes, but typically, there is just one way to update data and every single trait would need to have that.

I don't think there is a problem with:

`
tasks = []
if self.status:
	tasks.append(self.status.get_status())
`

but we will just have a long set of code opposed to:

`
await asyncio.gather(*trait.update() for trait in self.traits)
`

PR #426, opened 2025-08-09
chore: update container parsing using native typing and dataclass

allenporter:
Streamline the container parsing using modern python constructs for iterating over types. This adds support for nested dictionaries, and consolidates a few versions of similar logic into fewer places.

### DIFF
--- a/tests/test_containers.py
+++ b/tests/test_containers.py
@@ -110,16 +110,12 @@ def test_ignore_unknown_keys() -> None:
     """Test that we don't fail on unknown keys."""
     data = {
         "ignored_key": "This key should be ignored",
-        "simple": {"name": "Nested", "value": 100},
-        "items": ["item1", "item2"],
-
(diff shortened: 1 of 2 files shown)


tests/test_containers.py
```diff
@@ -23,6 +29,74 @@
...
+
+    simple: SimpleObject | None = None
+    items: list[str] | None = None
+    value: int | None = None
+    nested_dict: dict[str, SimpleObject] | None = None
+    nested_list: list[SimpleObject] | None = None
+    any: Any | None = None
+
+
+def test_simple_object() -> None:
```
Lash-L, line 52: So i'm not sure if we have this test elsewhere (i don't think so)

Can you add a specific test case where there is a extra key that we have never seen and is not part of the description. This has been the result of crashes before.

allenporter: Added.

PR #427, opened 2025-08-09
feat: Update the cli cache to also store network info

allenporter:
This allows using the CLI to talk to local devices without connecting to the server at all once local device information has been cached. A follow up step will support caching network info for the new device manager API, and will use the CLI to do this.

Additional improvements:
- Put the code for discovering and caching devices into a helper function
- Make the `~/.roborock` file easier to read by dumping with indent


Lash-L:
I think it makes sense to find a way to update the home data - although I guess the user could simply delete their cache manually

But home_data is stored, so if a user got a new device, they need to get the new home_data

Lash-L (dismissed):

Lash-L (approved):

PR #428, opened 2025-08-09
feat: add the ability to update supported_features via cli

Lash-L:
Heavily AI assisted

I kept the information like "Feature Info", etc in the table so that in the future, when we added more features, we can recacluate it for all devices.


allenporter (dismissed):
Very cool. I assume we'll want to be able to collect these from multiple people and re-generate the table (e.g. you and i have different devices so we can't merge them). As a future step would we have a place in the repo we check them in that we can merge with new devices then generate the markdown from that?

Lash-L:
> Very cool. I assume we'll want to be able to collect these from multiple people and re-generate the table (e.g. you and i have different devices so we can't merge them). As a future step would we have a place in the repo we check them in that we can merge with new devices then generate the markdown from that?

Definitely a good thing to add. Right now if you do it from within the repo, it will automatically update the existing one.

But some way to merge should be fairly trivial and would be helpful so people don't have to clone and commit to give the info.

I also store the featureinfo so that if we get new features to evaluate, we can re evaluate all of them at once

Lash-L:
Changed this to put the features in a yaml file so that it is easier to update.

roborock/cli.py
```diff
@@ -1,16 +1,17 @@
 import asyncio
 import json
 import logging
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from pathlib import Path
 from typing import Any
 
 import click
+import yaml
```
pelson, line 9: This dependency didn't get added to the `pyproject.toml`. Found this when doing a `uv run --with python-roborock python -c "import roborock.cli"`

allenporter: It's in the "dev" dependency group.

allenporter: It's in the "dev" dependency group.

Issue #392, opened 2025-06-21
Roborock Q10 S5 support


Shadester:
I got a new Q10 S5 and it seems like I can get the device in products, but not in devices?

{
  "api": "获取家庭详细信息",
  "result": {
    "id": 3807116,
    "name": "My Home",
    "lon": null,
    "lat": null,
    "geoName": null,
    "products": [
      {
        "id": "1O9BlCxWe4loekRE624qPv",
        "name": "Roborock Q10 Series",
        "model": "roborock.vacuum.ss07",
        "iconUrl": null,
        "attribute": null,
        "capability": 0,
        "category": "robot.vacuum.cleaner",
        "schema": [
          {
            "id": 101,
            "name": "RPC Request",
            "code": "rpc_request",
            "mode": "rw",
            "type": "RAW",
            "property": "null"
          },
          {
            "id": 102,
            "name": "RPC Response",
            "code": "rpc_response",
            "mode": "rw",
            "type": "RAW",
            "property": "null"
          },
          {
            "id": 120,
            "name": "错误代码",
            "code": "error_code",
            "mode": "ro",
            "type": "ENUM",
            "property": "{\"range\": []}"
          },
          {
            "id": 121,
            "name": "设备状态",
            "code": "state",
            "mode": "ro",
            "type": "VALUE",
            "property": "null"
          },
          {
            "id": 122,
            "name": "设备电量",
            "code": "battery",
            "mode": "ro",
            "type": "ENUM",
            "property": "{\"range\": []}"
          },
          {
            "id": 123,
            "name": "吸力档位",
            "code": "fan_power",
            "mode": "rw",
            "type": "ENUM",
            "property": "{\"range\": []}"
          },
          {
            "id": 124,
            "name": "拖地档位",
            "code": "water_box_mode",
            "mode": "rw",
            "type": "RAW",
            "property": "null"
          },
          {
            "id": 125,
            "name": "主刷寿命",
            "code": "main_brush_life",
            "mode": "ro",
            "type": "ENUM",
            "property": "{\"range\": []}"
          },
          {
            "id": 126,
            "name": "边刷寿命",
            "code": "side_brush_life",
            "mode": "ro",
            "type": "ENUM",
            "property": "{\"range\": []}"
          },
          {
            "id": 127,
            "name": "滤网寿命",
            "code": "filter_life",
            "mode": "ro",
            "type": "ENUM",
            "property": "{\"range\": []}"
          },
          {
            "id": 135,
[...]

PR #429, opened 2025-08-09
feat: add some basic B01 support

Lash-L:
Relates to #392

This is some basic B01 information that I got from information users have gotten me and a little reverse engineering. We will need more help from users to get it in a good state. But I wanted to add this so that we can get it to a point where users can test.

### DIFF
--- a/roborock/devices/b01_channel.py
+++ b/roborock/devices/b01_channel.py
@@ -1,0 +1,29 @@
+"""Thin wrapper around the MQTT channel for Roborock B01 devices."""
+
+from __future__ import annotations
+
+import logging
+
+from roborock.protocols.b01_protocol import (
+    CommandType,
+    ParamsType,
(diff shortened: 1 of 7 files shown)


roborock/devices/b01_channel.py
```diff
@@ -0,0 +1,29 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+
+
+async def send_decoded_command(
+    mqtt_channel: MqttChannel,
+    dps: int,
+    command: CommandType,
+    params: ParamsType,
+) -> dict:
```
allenporter: Can this be `dict[int, Any]` given its the signature of `decode_rpc_response`

```suggestion
) -> dict[int, Any]:
```

allenporter:
Ah messed up the lint commit...

PR #431, opened 2025-08-10
feat: add a new type for supported features

Lash-L:
Got these while trying to nail down a better way to fix all the errors with incorrect names for modes or smart mode errors.

This theoretically should make it so we can more dynamically create mop/fan modes. TBD though.

Open to better naming if you can think of it.

### DIFF
--- a/roborock/device_features.py
+++ b/roborock/device_features.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field, fields
-from enum import IntEnum
+from enum import IntEnum, StrEnum
 from typing import Any
 
 from roborock import RoborockProductNickname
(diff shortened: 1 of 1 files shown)


roborock/device_features.py
```diff
@@ -59,6 +59,121 @@ class NewFeatureStrBit(IntEnum):
...
+
+
+PEARL_FEATURES = NEW_DEFAULT_FEATURES + SINGLE_LINE_CAMERA_FEATURES + [ProductFeatures.MOP_SPIN_MODULE]
+PEARL_PLUS_FEATURES = NEW_DEFAULT_FEATURES + RGB_CAMERA_FEATURES + [ProductFeatures.MOP_SPIN_MODULE]
+ULTRON_FEATURES = NEW_DEFAULT_FEATURES + DUAL_LINE_CAMERA_FEATURES + [ProductFeatures.MOP_SHAKE_MODULE]
+ULTRONSV_FEATURES = NEW_DEFAULT_FEATURES + RGB_CAMERA_FEATURES + [ProductFeatures.MOP_SHAKE_MODULE]
+TANOSS_FEATURES = NEW_DEFAULT_FEATURES + [ProductFeatures.MOP_SHAKE_MODULE]
+TOPAZSPOWER_FEATURES = NEW_DEFAULT_FEATURES + [ProductFeatures.MOP_SHAKE_MODULE]
+
+product_feature_map = {
```
allenporter, line 126: `PRODUCT_FEATURE_MAP`?

Lash-L: Off my machine and on my phone, can do tomorrow 👍🏻

allenporter:
Just an idea, we can also consider storing these mappings a configuration file: https://github.com/allenporter/pyrainbird/blob/main/pyrainbird/resources/models.yaml

(I inherited this code base, not by creation)

allenporter (approved):
Let's handle in a followup

PR #432, opened 2025-08-10
Add an explicit module for caching

allenporter:
Add an explicit module for caching that initially is extended to support storing network information and avoiding the need for complext Rborocksession.

### DIFF
--- a/roborock/devices/cache.py
+++ b/roborock/devices/cache.py
@@ -1,0 +1,47 @@
+"""This module provides caching functionality for the Roborock device management system.
+
+This module defines a cache interface that you may use to cache device
+information to avoid unnecessary API calls. Callers may implement
+this interface to provide their own caching mechanism.
+"""
+
+from dataclasses import dataclass, field
+from typing import Protocol
+
+from roborock.containers import HomeData, NetworkInfo
+
+
+@dataclass
+class CacheData:
+    """Data structure for caching device information."""
+
+    home_data: HomeData | None = None
+    """Home data containing device and product information."""
+
+    network_info: dict[str, NetworkInfo] = field(default_factory=dict)
+    """Network information indexed by device DUID."""
+
+
+class Cache(Protocol):
+    """Protocol for a cache that can store and retrieve values."""
+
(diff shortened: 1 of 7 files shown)


roborock/devices/cache.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+from dataclasses import dataclass, field
+from typing import Protocol
+
+from roborock.containers import HomeData, NetworkInfo
+
+
+@dataclass
+class CacheData:
```
Lash-L, line 15: I think adding local key would be very helpful here. That theoretically would then have everything you need for local usage

roborock/devices/device_manager.py
```diff
@@ -57,19 +56,27 @@ def __init__(
...
         self._home_data_api = home_data_api
+        self._cache = cache
         self._device_creator = device_creator
         self._devices: dict[str, RoborockDevice] = {}
         self._mqtt_session = mqtt_session
 
     async def discover_devices(self) -> list[RoborockDevice]:
         """Discover all devices for the logged-in user."""
-        home_data = await self._home_data_api()
+        cache_data = await self._cache.get()
```
Lash-L, line 73: I think we should have a way to force an update

roborock/devices/cache.py
allenporter: I believe this is in `home_data.devices.local_key`?

Lash-L: Ah yes sorry - was conflating user data in my head - my bad

roborock/devices/device_manager.py
allenporter: I would like to see if we can address this in the cache implementation e.g. caller can flush their own cache implementation if they don't want caching. If that doesn't work we can add this as an explicit flag here?

The reason why i am thinking about this is because we also may need to flush network info as well sometimes, so it seems nice to invalidate the whole cache and try to refresh.

Lash-L: Okay sounds good!

Lash-L (approved):

PR #437, opened 2025-08-10
feat: add dynamic clean modes

Lash-L:
Allow us to dynamically set the valid modes.

Needs to be integrated but wanted some feedback on structure


Lash-L:
> Structure looks good to me. I might suggest adding some higher level comments simiiar to what were discussing about why these various behaviors exist for different mode (e.g. if there is some physical feature involved, explaining that)

Good call out - can do

PR #438, opened 2025-08-10
fix: merge the local api with the local v1 api

allenporter:
Remove the `local_api` module and flatten the class hierarchy. This removes the need for a generic abstract message and potentially fields in `RoborockMessage` that are protocol specific. Now all clients only call their own `_send_message` method so it can pass any arguments it wants.


Lash-L (approved):
Looks good - good simplification

PR #439, opened 2025-08-11
chore: avoid re-parsing RoborockMessages and replace with passing explicit arguments

allenporter:
Now pass explicit arguments to the send message functions rather than encoding then decoding the messages.


Lash-L (approved):
Does this give any kind of performance boost? Or was it more so just to avoid redundant code? anyways - LGTM

allenporter:
> Does this give any kind of performance boost? Or was it more so just to avoid redundant code? anyways - LGTM

It was just about deleting code, and separating the different response types.

PR #448, opened 2025-08-13
feat: improve dynamic clean modes

Lash-L:
There are some things i have to figure out (i.e. typing in some places)

And how to get locations to the trait

And how much of a breaking change is okay vs if i should try to make both possible at the same time

### DIFF
--- a/roborock/device_features.py
+++ b/roborock/device_features.py
@@ -205,7 +205,7 @@ _BASE_PRODUCT_FEATURE_MAP: dict[RoborockProductNickname, list[ProductFeatures]] 
     RoborockProductNickname.RUBYSLITE: [ProductFeatures.MOP_ELECTRONIC_MODULE],
 }
 
-product_feature_map: dict[RoborockProductNickname, list[ProductFeatures]] = {
+PRODUCT_FEATURE_MAP: dict[RoborockProductNickname, list[ProductFeatures]] = {
     product: (
         features
         + ([ProductFeatures.DEFAULT_CLEANMODECUSTOM] if product not in PRODUCTS_WITHOUT_CUSTOM_CLEAN else [])
@@ -538,7 +538,7 @@ class DeviceFeatures:
                     kwargs[f.name] = product_nickname not in blacklist
             elif (product_features := f.metadata.get("product_features")) is not None:
                 if product_nickname is not None:
-                    available_features = product_feature_map.get(product_nickname, [])
+                    available_features = PRODUCT_FEATURE_MAP.get(product_nickname, [])
                     if any(feat in available_features for feat in product_features):  # type: ignore
                         kwargs[f.name] = True
 


roborock/version_1_apis/roborock_client_v1.py
```diff
@@ -154,6 +154,10 @@ def __init__(self, device_info: DeviceData, security_data: SecurityData | None)
         """Initializes the Roborock client."""
         super().__init__(device_info)
         self._status_type = CustomStatus(device_info.device_features, region=device_info.region)
+        if security_data is not None:
```
allenporter: this looks like it was already checked in, odd rebase?

Lash-L: weird - on my side that doesn't show as a change.

image

roborock/cli.py
```diff
@@ -235,7 +235,7 @@ async def status(ctx, device_id):
     devices = home_data.devices + home_data.received_devices
     device = next(device for device in devices if device.duid == device_id)
     product_info: dict[str, HomeDataProduct] = {product.id: product for product in home_data.products}
-    device_data = DeviceData(device, product_info[device.product_id].model)
+    device_data = DeviceData(device, product_info[device.product_id].model, region=cache_data.user_data.region)
```
Lash-L: We are kind of in a weird situation.

We need the DeviceFeatures in order to create the valid options for Status - so it is stored in DeviceData. But we need DeviceData in order to set up the client - which we need to call get_init_status to get the DeviceFeatures.

Maybe the client just has a function to update it's "app_init_status" and then it updates it's status? Any thoughts?

roborock/device_features.py
```diff
@@ -174,6 +175,51 @@ class ProductFeatures(StrEnum):
...
+        RoborockProductNickname.TANOSSE,
+        RoborockProductNickname.TANOSSC,
+        RoborockProductNickname.ULTRONLITE,
+        RoborockProductNickname.ULTRONE,
+        RoborockProductNickname.RUBYPLUS,
+        RoborockProductNickname.RUBYSLITE,
+        RoborockProductNickname.RUBYSC,
+        RoborockProductNickname.RUBYSE,
+    }
+    for product, feature_list in product_feature_map.items():
```
allenporter: Don't reall love updating the map like this. Can we use the logic here to build the entire map from the start?

allenporter (commented):
It may be worth clarifying in `clean_modes.py` what is internal vs part of the public api. One way you can do this is by setting `__all__ = [ ... ]` to the list of public things or empty if this is totally private and internal.

my assumption is that file is moslty internal and the actual things we want callers to check will be in a trait when those are ready.

roborock/device_features.py
Lash-L: So this is how they do it in their codebase. And it would be the easiest to update in the future as I can just copy the newer versions.

The issue with changing it is that there are not in checks, meaning this automatically works for any new devices. Alternatively, we would have to add it for each product when it comes out.

I definitely agree it is not clean, but it is unfortunately how they designed it.

Lash-L:
> It may be worth clarifying in `clean_modes.py` what is internal vs part of the public api. One way you can do this is by setting `__all__ = [ ... ]` to the list of public things or empty if this is totally private and internal.
> 
> my assumption is that file is moslty internal and the actual things we want callers to check will be in a trait when those are ready.

It is actually basically all external. You provide the device features, we will tell you which of the modes your device supports.

roborock/device_features.py
allenporter: I'm fine with a dynamic feature map, i'm just saying this mutating a global variatble python object `product_feature_map`.  Create the `product_feature_map` right from the start rather than calling a method to update this global variable.

roborock/device_features.py
```diff
@@ -173,6 +205,16 @@ class ProductFeatures(StrEnum):
     RoborockProductNickname.RUBYSLITE: [ProductFeatures.MOP_ELECTRONIC_MODULE],
 }
 
+product_feature_map: dict[RoborockProductNickname, list[ProductFeatures]] = {
```
allenporter: PRODUCT_FEATURE_MAP?

allenporter (approved):

PR #449, opened 2025-08-13
feat: improve B01 support

Lash-L:
I think the next step is to make a B1Channel that does the decoding and handles things for the device? I would like your opinion, as B01 does not have response matching.

### DIFF
--- a/roborock/b01_containers.py
+++ b/roborock/b01_containers.py
@@ -1,0 +1,417 @@
+from dataclasses import dataclass, field
+
+from .code_mappings import RoborockModeEnum
+from .containers import RoborockBase
+
+
+class WorkStatusMapping(RoborockModeEnum):
+    """Maps the general status of the robot."""
+
+    SLEEPING = ("sleeping", 0)
+    WAITING_FOR_ORDERS = ("waiting_for_orders", 1)
+    PAUSED = ("paused", 2)
+    DOCKING = ("docking", 3)
+    CHARGING = ("charging", 4)
+    SWEEP_MOPING = ("sweep_moping", 5)
+    SWEEP_MOPING_2 = ("sweep_moping_2", 6)
+    MOPING = ("moping", 7)
+    UPDATING = ("updating", 8)
+    MOP_CLEANING = ("mop_cleaning", 9)
+    MOP_AIRDRYING = ("mop_airdrying", 10)
+
+
+class SCWindMapping(RoborockModeEnum):
+    """Maps suction power levels."""
+
+    SILENCE = ("quiet", 0)
+    STANDARD = ("balanced", 1)
+    STRONG = ("turbo", 2)
+    SUPER_STRONG = ("max", 3)
+    MAX = ("max_plus", 4)
+
+
+class WaterLevelMapping(RoborockModeEnum):
+    """Maps water flow levels."""
+
+    LOW = ("low", 0)
+    MEDIUM = ("medium", 1)
+    HIGH = ("high", 2)
+
+
+class CleanTypeMapping(RoborockModeEnum):
+    """Maps the type of cleaning (Vacuum, Mop, or both)."""
+
+    VACUUM = ("vacuum", 0)
+    VAC_AND_MOP = ("vac_and_mop", 1)
+    MOP = ("mop", 2)
+
+
+class CleanRepeatMapping(RoborockModeEnum):
+    """Maps the cleaning repeat parameter."""
+
+    ONCE = ("once", 0)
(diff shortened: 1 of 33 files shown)


roborock/devices/mqtt_channel.py
```diff
@@ -69,7 +69,8 @@ def message_handler(payload: bytes) -> None:
                 return
             for message in messages:
                 _LOGGER.debug("Received message: %s", message)
-                asyncio.create_task(self._resolve_future_with_lock(message))
+                if message.version == b"1.0":
```
allenporter: I believe a01 needs responses. does that get covered by this?

roborock/devices/mqtt_channel.py
```diff
@@ -95,6 +96,15 @@ async def _resolve_future_with_lock(self, message: RoborockMessage) -> None:
             else:
                 _LOGGER.debug("Received message with no waiting handler: request_id=%s", request_id)
 
+    async def send_message_no_wait(self, message: RoborockMessage) -> None:
```
allenporter: How about we just call this `publish`? i have a pending PR I have not sent yet to remove RPCs form this method and push it up a level, which is renaming this also.

roborock/protocols/b01_protocol.py
```diff
@@ -23,7 +23,7 @@
 
 def encode_mqtt_payload(dps: int, command: CommandType, params: ParamsType) -> RoborockMessage:
     """Encode payload for B01 commands over MQTT."""
-    dps_data = {"dps": {dps: {"method": command, "params": params or []}}}
+    dps_data = {"dps": {dps: {"method": str(command), "msgId": "1751755654575", "params": params or []}}}
```
allenporter: can you extract this msg id to a constant and give it a self documenting name that explains what it is?

allenporter (changes requested):

roborock/devices/mqtt_channel.py
Lash-L: I can remove my logic in mqtt manager and just defer to your pr if that works for you. This was more so getting things working, not necessarily thinking through the best designs

roborock/protocols/b01_protocol.py
Lash-L: It actually shouldn't be a constant, I was just doing that for testing sake. I will make it dynamic.

roborock/b01_containers.py
```diff
@@ -0,0 +1,417 @@
...
+    F_2110 = ("fault_2110", 2110)
+    F_2111 = ("fault_2111", 2111)
+    F_2112 = ("fault_2112", 2112)
+    F_2113 = ("fault_2113", 2113)
+    F_2114 = ("fault_2114", 2114)
+    F_2115 = ("fault_2115", 2115)
+
+
+@dataclass
+class B01Props(RoborockBase):
```
Lash-L, line 353: This actually does not work right now. As RoborockBase is in the other file,  so it doesn't have access to the typing in here. Need to figure out a work around

Lash-L: Should be resolved now

tests/test_containers.py
```diff
@@ -326,3 +326,104 @@ def test_no_value():
     assert s.dock_type == RoborockDockTypeCode.unknown
     assert -9999 not in RoborockDockTypeCode.keys()
     assert "missing" not in RoborockDockTypeCode.values()
+
+
+def test_b01props_deserialization():
+    """Test that B01Props can be deserialized after its module is dynamically imported."""
+    from roborock.b01_containers import (
```
allenporter: this could just be top level i think

roborock/containers.py
```diff
@@ -109,6 +110,11 @@ def _decamelize(s: str):
 @dataclass
 class RoborockBase:
     _ignore_keys = []  # type: ignore
+    _registry: ClassVar[dict[str, type["RoborockBase"]]] = {}
```
allenporter: what is this needed for?

allenporter (commented):
Generally looks fine, just a question

roborock/containers.py
Lash-L: It's part of the fix to get roborockbase working on objects not in containers.py

Open to other solutions as well

allenporter: What does it do and how does it work?

I think just removing `from __future__ import annotations` from that file will make things work.

Lash-L: Believe you are correct. Annotations must have been there before and this fixed it while it was. This works without registry now though so I deleted it

PR #450, opened 2025-08-14
feat: add container and function for app_init_status

Lash-L:


roborock/containers.py
```diff
@@ -725,6 +725,33 @@ class NetworkInfo(RoborockBase):
     rssi: int | None = None
 
 
+@dataclass
+class AppInitStatusLocalInfo(RoborockBase):
+    """"""
```
allenporter: Add docstring or remove? here and below

allenporter (changes requested):

allenporter (approved):

PR #451, opened 2025-08-17
Extract common module for managing pending RPCs

allenporter:
Extract common module for managing pending RPCs. The logic may move elsewhere in follow up PRs.


roborock/devices/pending.py
```diff
@@ -0,0 +1,45 @@
...
+            if (future := self._waiting_queue.pop(key, None)) is not None:
+                future.cancel()
+
+    async def resolve(self, key: K, value: V) -> None:
+        """Resolve waiting future with proper locking."""
+        async with self._queue_lock:
+            if (future := self._waiting_queue.pop(key, None)) is not None:
+                future.set_result(value)
+            else:
+                _LOGGER.debug("Received unsolicited message: %s", key)
```
Lash-L, line 45: So for B01 - this is going to debug for every message we receive. Probably fine, but may be a bit of log spam for users when they turn on debug mode. Probably not that impactful though

allenporter: OK, we can it use an mqtt channel that does not use this. I think i want to move this out into the v1 rpc channel.

PR #452, opened 2025-08-18
chore: remove dependencies on `get_request_id` by changing where RequestMessage request_id is set

allenporter:
Remove some of the payload encoding abstractions so that we don't need to do a round trip for getting `request_id`. We can just get it directly when it is assigned to the outgoing message.

In a follow up change we will remove `get_request_id` from `RoborockMessage` where the last dependency is in the new mqtt API. This can set us up for making the mqtt channel more appropriate to use for b01 where it doesn't depend on v1 and a01 request id semantics.

### DIFF
--- a/roborock/version_1_apis/roborock_mqtt_client_v1.py
+++ b/roborock/version_1_apis/roborock_mqtt_client_v1.py
@@ -9,7 +9,7 @@ from roborock.cloud_api import RoborockMqttClient
 
 from ..containers import DeviceData, UserData
 from ..exceptions import CommandVacuumError, RoborockException, VacuumError
-from ..protocols.v1_protocol import create_mqtt_payload_encoder, create_security_data
+from ..protocols.v1_protocol import RequestMessage, create_security_data
 from ..roborock_message import (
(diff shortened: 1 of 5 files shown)


roborock/version_1_apis/roborock_mqtt_client_v1.py
```diff
@@ -28,12 +28,12 @@ def __init__(self, user_data: UserData, device_info: DeviceData, queue_timeout:
         rriot = user_data.rriot
         if rriot is None:
             raise RoborockException("Got no rriot data from user_data")
-        security_data = create_security_data(rriot)
         RoborockMqttClient.__init__(self, user_data, device_info)
+        security_data = create_security_data(rriot)
```
Lash-L, line 32: If we are using this in the base class, why define it below as well?

allenporter: It's not setting it as a member in the base class, its just used for some diagnostics, so setting it here so it can be accessed below. Ideally it wouldn't be in the base class given that it's only used in one type of subclass.

PR #453, opened 2025-08-18
feat: fix a01 and b01 response handling in new api

allenporter:
This builds on #452 by removing additional response decoding in `RoborockMessage`, separating out the different protocols better. This now has separate RPC logic for a01, b01, and v1. This removes the `PendingRpc` abstraction that was just added since it is no longer necessary since we can handle all responses within a single function call using a subscribe callback.

### DIFF
--- a/roborock/devices/a01_channel.py
+++ b/roborock/devices/a01_channel.py
@@ -66,7 +66,7 @@ async def send_decoded_command(
         """Handle incoming messages and resolve the future."""
         try:
             decoded = decode_rpc_response(response_message)
-        except RoborockException as ex: 
+        except RoborockException as ex:
             _LOGGER.info("Failed to decode a01 message: %s: %s", response_message, ex)
             return
         for key, value in decoded.items():


roborock/devices/a01_channel.py
```diff
@@ -39,5 +44,46 @@ async def send_decoded_command(
...
-    return decode_rpc_response(response)  # type: ignore[return-value]
+
+    # We only block on a response for queries
+    param_values = {int(k): v for k, v in params.items()}
+    if not (
+        query_values := param_values.get(int(RoborockDyadDataProtocol.ID_QUERY))
+        or param_values.get(int(RoborockZeoProtocol.ID_QUERY))
+    ):
+        await mqtt_channel.publish(roborock_message)
+        return {}
```
Lash-L: I think this comment could be clearer for others looking through. With the not here it is a bit tricky to quickly read and confused me for a second (maybe that's just because i'm tired)

I think something like "If the message is not a query, we don't block" may be clearer for this if block.

roborock/devices/a01_channel.py
```diff
@@ -39,5 +44,46 @@ async def send_decoded_command(
...
     _LOGGER.debug("Sending MQTT command: %s", params)
     roborock_message = encode_mqtt_payload(params)
-    response = await mqtt_channel.send_message(roborock_message)
-    return decode_rpc_response(response)  # type: ignore[return-value]
+
+    # We only block on a response for queries
+    param_values = {int(k): v for k, v in params.items()}
+    if not (
+        query_values := param_values.get(int(RoborockDyadDataProtocol.ID_QUERY))
+        or param_values.get(int(RoborockZeoProtocol.ID_QUERY))
```
Lash-L: technically a redundant if as ID_QUERY is the same.

roborock/devices/a01_channel.py
```diff
@@ -39,5 +44,46 @@ async def send_decoded_command(
...
+        """Handle incoming messages and resolve the future."""
+        try:
+            decoded = decode_rpc_response(response_message)
+        except RoborockException:
+            return
+        for key, value in decoded.items():
+            if key in query_values:
+                result[key] = value
+        if len(result) != len(query_values):
+            return
```
Lash-L: We should definitely log here what is happening

roborock/devices/a01_channel.py
```diff
@@ -39,5 +44,46 @@ async def send_decoded_command(
...
+    # This could be simplified if we can assume there is a single response.
+    finished = asyncio.Event()
+    result: dict[int, Any] = {}
+
+    def find_response(response_message: RoborockMessage) -> None:
+        """Handle incoming messages and resolve the future."""
+        try:
+            decoded = decode_rpc_response(response_message)
+        except RoborockException:
+            return
```
Lash-L: We should log here

roborock/devices/a01_channel.py
```diff
@@ -39,5 +44,46 @@ async def send_decoded_command(
...
+        for key, value in decoded.items():
+            if key in query_values:
+                result[key] = value
+        if len(result) != len(query_values):
+            return
+        _LOGGER.debug("Received query response: %s", result)
+        if not finished.is_set():
+            finished.set()
+
+    unsub = await mqtt_channel.subscribe(find_response)
```
Lash-L, line 82: Okay maybe i'm just getting confused by the asyncness and maybe that is leading to my confusion with the returns above as well.

My first thought when I was reading this was that we are adding a subscription everytime that we send a message. Then when a message goes through, we are going through all of the subscriptions and if they aren't a match, we end up returning None on them.

However, looking at it, it seems that subscribe holds exactly one callback, so if I had three different send_commands I wanted to send at once - wouldn't that override the subscription callback for each one? Or does the lock within roborock_session prevent that - if that's the case, doesn't that dramatically increase the amount of time that I will be waiting to get a response?

Maybe it's okay, as if you are using the api correctly, all IDs that you are querying should be done with one function call - but there could be commands that we don't utilize in A01 that should have responses as well

Lash-L (changes requested):

roborock/devices/a01_channel.py
allenporter: You're right, the code I was using for mqtt session changed a bunch of times, but i reverted it for this PR. 

The intent was to use a version that allowed any number of callbacks. I think a version fo this was reverted so this no longer makes sense.

roborock/devices/a01_channel.py
allenporter: Changed to use a single constant and added a comment.  I did know these used the same values but wasn't sure that depending on only one constant here made sense readability wise. So, now there's a comment.

roborock/devices/a01_channel.py
allenporter: Updated

roborock/devices/a01_channel.py
allenporter: Done

roborock/devices/a01_channel.py
allenporter: Done.

roborock/devices/a01_channel.py
allenporter: Updated this to remove the limitation of a single subscribe call for the mqtt channel. This matches the behavior of local channel.

PR #457, opened 2025-08-21
chore: add snapshot tests for parsing device wire formats

allenporter:
Add some raw device payloads to the repo for capturing the device message formats. This captures a prettyprint of the json output which can be manually copied/used in other more focused tests (e.g. trait tests) as needed, or we can automate extracting the trait messages into separate data files as well.

Add a dev dependency on syrupy for snapshot tests that can be updated with `--snapshot-update`


tests/protocols/testdata/v1_protocol/get_dnd.json
```diff
@@ -0,0 +1 @@
+{"t": 1755785801, "dps": {"102": "{\"id\":20002,\"result\":[{\"start_hour\":22,\"start_minute\":0,\"end_hour\":8,\"end_minute\":0,\"enabled\":1}]}"}}
```
Lash-L, line 1: So the only potential issue:

Some devices give more information than this. I may not have it represented in the container, but my newer vacuum gives extra information for DND about what is allowed during DND.

allenporter: My assumption is that we'd want to add more samples from devices that are interesting or from debug reports. This came from my device verbatim, so we can also add messages from yours for example.

Lash-L: In this case how is that done? an extra line of json? replace this json? new file? What did you have in mind

Lash-L (dismissed):

tests/protocols/testdata/v1_protocol/get_dnd.json
allenporter: I would say add another file. We could put the device name in the file, or put in a subdirectory per device.

Lash-L (approved):

PR #459, opened 2025-09-02
chore: move broadcast_protocol to its own file

Lash-L:
Extracting it out for some other future planned changes.

### DIFF
--- a/roborock/broadcast_protocol.py
+++ b/roborock/broadcast_protocol.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+from asyncio import BaseTransport, Lock
+
+from construct import (  # type: ignore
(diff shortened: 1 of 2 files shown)


roborock/broadcast_protocol.py
```diff
@@ -0,0 +1,71 @@
...
+class RoborockProtocol(asyncio.DatagramProtocol):
+    def __init__(self, timeout: int = 5):
+        self.timeout = timeout
+        self.transport: BaseTransport | None = None
+        self.devices_found: list[BroadcastMessage] = []
+        self._mutex = Lock()
+
+    def __del__(self):
+        self.close()
+
```
allenporter: While you're here, let's not have this and use the explicit close below only

```suggestion
```

allenporter (dismissed):

PR #460, opened 2025-09-03
fix: handle auth expiring

Lash-L:
Seemingly, tokens now sometimes expire. I'm not sure how long it takes, but the issue is being seen in a few issues. Our error handling logic was wrong in the old system (It was still operating under some mqtt3 logic instead of 5.)

Now, we identify bad auth errors, and pass forward a exception that makes that clear. Then we can prompt a reauth on the HA side.


Lash-L:
We will likely need to do something similar in the new mqtt side once everything is finalized.

roborock/cloud_api.py
```diff
@@ -78,21 +79,27 @@ def __init__(self, user_data: UserData, device_info: DeviceData) -> None:
         self._encoder: Encoder = create_mqtt_encoder(device_info.device.local_key)
 
     def _mqtt_on_connect(self, *args, **kwargs):
-        _, __, ___, rc, ____ = args
+        rc: ReasonCodes = args[3]
         connection_queue = self._waiting_queue.get(CONNECT_REQUEST_ID)
-        if rc != mqtt.MQTT_ERR_SUCCESS:
-            message = f"Failed to connect ({mqtt.error_string(rc)})"
+        if rc != 0:
```
allenporter: Does reason_code.is_failure work here?

roborock/cloud_api.py
```diff
@@ -78,21 +79,27 @@ def __init__(self, user_data: UserData, device_info: DeviceData) -> None:
         self._encoder: Encoder = create_mqtt_encoder(device_info.device.local_key)
 
     def _mqtt_on_connect(self, *args, **kwargs):
-        _, __, ___, rc, ____ = args
+        rc: ReasonCodes = args[3]
```
allenporter: How about we just make this 

`_mqtt_on_connect(client, userdata: ..., flags: .., reason_code: ..., properties):`

roborock/cloud_api.py
```diff
@@ -78,21 +79,27 @@ def __init__(self, user_data: UserData, device_info: DeviceData) -> None:
...
         connection_queue = self._waiting_queue.get(CONNECT_REQUEST_ID)
-        if rc != mqtt.MQTT_ERR_SUCCESS:
-            message = f"Failed to connect ({mqtt.error_string(rc)})"
+        if rc != 0:
+            message = f"Failed to connect ({str(rc)})"
```
allenporter: I think you can omit `str` here

allenporter (changes requested):

roborock/cloud_api.py
```diff
@@ -76,43 +80,63 @@ def __init__(self, user_data: UserData, device_info: DeviceData) -> None:
...
-
-    def _mqtt_on_connect(self, *args, **kwargs):
-        _, __, ___, rc, ____ = args
+        self.previous_attempt_was_subscribe = False
+        self._topic = f"rr/m/o/{self._mqtt_user}/{self._hashed_user}/{self.device_info.device.duid}"
+
+    def _mqtt_on_connect(
+        self,
+        client: mqtt.Client,
+        data: object,
```
allenporter:
```suggestion
        userdata: object,
```

roborock/cloud_api.py
```diff
@@ -178,3 +202,33 @@ def _send_msg_raw(self, msg: bytes) -> None:
         )
         if info.rc != mqtt.MQTT_ERR_SUCCESS:
             raise RoborockException(f"Failed to publish ({mqtt.error_string(info.rc)})")
+
+    async def validate_connection(self) -> None:
+        """Override the default validate connection to try to re-subscribe rather than disconnect."""
```
allenporter: I'm having trouble following everything going on here because of:
- the new state member variable
- keepalive
- interaction with the parent class, invoking parent calss validate connection 4 different ways

Could this have an opening description of (1) big picture whats supposed to be happening here then (2) at a detailed level what these flows are and how the 3 things interact? Im also unsure how this pare relates to the error handling issue above.

A suggestion for simplification would be to no longer call the parent class validate_connection logic.  Instead move that down into the local and cloud clients and inline it here so there are fewer interactions with the parent class.

roborock/cloud_api.py
```diff
@@ -76,43 +80,63 @@ def __init__(self, user_data: UserData, device_info: DeviceData) -> None:
...
+    ):
         connection_queue = self._waiting_queue.get(CONNECT_REQUEST_ID)
-        if rc != mqtt.MQTT_ERR_SUCCESS:
-            message = f"Failed to connect ({mqtt.error_string(rc)})"
+        if rc.is_failure:
+            message = f"Failed to connect ({rc})"
             self._logger.error(message)
             if connection_queue:
-                connection_queue.set_exception(VacuumError(message))
+                # These are the ReasonCodes relating to authorization issues.
```
allenporter: You could consider extracting a constant like this:
```
AUTH_ERROR_REASON_CODES = {24, 25, 133, 134, 135, 144}
```
which could make this comment unnecessary

Lash-L:
Did some simplification. I don't think i specifically want to request re-auth, as waiting is a fine solution.

allenporter (changes requested):

roborock/cloud_api.py
Lash-L: I actually removed this logic as I think that trying to force a reauth is the wrong approach now.

roborock/cloud_api.py
Lash-L: Take a look now, i broke things out into some helper functions

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -138,6 +138,12 @@ async def ping(self) -> None:
             response_protocol=RoborockMessageProtocol.PING_RESPONSE,
         )
 
+    async def validate_connection(self) -> None:
```
allenporter: This can now be private `_validate_connection` on both unless you think it has been assumed to be part of the public API.

roborock/cloud_api.py
```diff
@@ -178,3 +196,62 @@ def _send_msg_raw(self, msg: bytes) -> None:
         )
         if info.rc != mqtt.MQTT_ERR_SUCCESS:
             raise RoborockException(f"Failed to publish ({mqtt.error_string(info.rc)})")
+
+    async def unsubscribe(self):
```
allenporter: Can this be private?

allenporter: Can you add typing information to the return value?

roborock/cloud_api.py
```diff
@@ -178,3 +196,62 @@ def _send_msg_raw(self, msg: bytes) -> None:
...
+    async def reconnect(self) -> None:
+        """Reconnect to the MQTT broker."""
+        await self.async_disconnect()
+        await self.async_connect()
+
+    async def validate_connection(self) -> None:
+        """Override the default validate connection to try to re-subscribe rather than disconnect.
+        When something seems to be wrong with our connection, we should follow the following steps:
+        1. Try to unsubscribe and resubscribe from the topic.
+        2. If we don't end up getting a message, we should completely disconnect and reconnect to the MQTT broker.
```
allenporter, line 234: Thank you this is helpful. Can you explain how #2 works? Maybe renaming `previous_attempt_was_subscribe` to `received_message` and inverting the logic would fix my confusion and we're good.

roborock/cloud_api.py
```diff
@@ -178,3 +196,62 @@ def _send_msg_raw(self, msg: bytes) -> None:
...
+        3. We will continue to try to disconnect and reconnect until we get a message.
+        4. If we get a message, the next time connection is lost, We will go back to step 1.
+        """
+        if not self.should_keepalive():
```
allenporter, line 239: While you're here, can you also explain what this logic is supposed to be doing in words? I don't understand the point of it exactly.

It seems confusing because this logic below looks kinda like keepalive logic, but it runs when "not should keepalive" so i don't totally follow.

roborock/cloud_api.py
```diff
@@ -178,3 +196,62 @@ def _send_msg_raw(self, msg: bytes) -> None:
...
+            self._logger.info("Resetting Roborock connection due to keepalive timeout")
+            if self.previous_attempt_was_subscribe:
+                # If we have already tried to unsub and resub, and we are still in this state,
+                # we should try to reconnect.
+                return await self.reconnect()
+            try:
+                # Mark that we have tried to unsubscribe and resubscribe
+                self.previous_attempt_was_subscribe = True
+                if await self.unsubscribe() == 0:
+                    # If we fail to unsubscribe, reconnect to the broker
```
allenporter, line 249: I would have though "== 0" means success given the handling in these functions. Is this correct?

Lash-L: Good catch

roborock/cloud_api.py
Lash-L: Yeah, this logic is from a long time ago

But in this context:

if keepalive = True, the connection is fine and don't do anything.

if keepalive = false, we need to fix the connection. The current connection should not be kept alive

Lash-L:
Anything outstanding here @allenporter?

roborock/cloud_api.py
```diff
@@ -76,43 +81,57 @@ def __init__(self, user_data: UserData, device_info: DeviceData) -> None:
...
             if connection_queue:
                 connection_queue.set_exception(VacuumError(message))
             return
-        self._logger.info(f"Subscribed to topic {topic}")
+        self._logger.info(f"Subscribed to topic {self._topic}")
         if connection_queue:
             connection_queue.set_result(True)
 
     def _mqtt_on_message(self, *args, **kwargs):
+        self.received_message_since_last_disconnect = False
```
allenporter: Should this now be `True`?

Unless i misunderstood the naming of this.

Lash-L: Yes good call - it did not match the new naming. fixed.

PR #471, opened 2025-09-13
feat: Add a DnD trait and fix bugs in the rpc channels

allenporter:
Add a Do Not Disturb trait. This trait sends commands that change device state with data returned, which uncovers issues like:
- there is a bug in `from_dict` with reusing variable names
- Using `as_dict` to serialize the data has an issue adding additional timer fields that are moved to be properties

### DIFF
--- a/roborock/devices/traits/dnd.py
+++ b/roborock/devices/traits/dnd.py
@@ -27,16 +27,16 @@ class DoNotDisturbTrait(Trait):
 
     def __init__(self, rpc_channel: Callable[[], V1RpcChannel]) -> None:
         """Initialize the DoNotDisturbTrait."""
-        self._rpc_channel = rpc_channel
+        self._send_command = lambda *args, **kwargs: rpc_channel().send_command(*args, **kwargs)
 
     async def get_dnd_timer(self) -> DnDTimer:
         """Get the current Do Not Disturb (DND) timer settings of the device."""
-        return await self._rpc_channel().send_command(RoborockCommand.GET_DND_TIMER, response_type=DnDTimer)
+        return await self._send_command(RoborockCommand.GET_DND_TIMER, response_type=DnDTimer)
 
     async def set_dnd_timer(self, dnd_timer: DnDTimer) -> None:
         """Set the Do Not Disturb (DND) timer settings of the device."""
-        await self._rpc_channel().send_command(RoborockCommand.SET_DND_TIMER, params=dnd_timer.as_dict())
+        await self._send_command(RoborockCommand.SET_DND_TIMER, params=dnd_timer.as_dict())
 
     async def clear_dnd_timer(self) -> None:
         """Clear the Do Not Disturb (DND) timer settings of the device."""
-        await self._rpc_channel().send_command(RoborockCommand.CLOSE_DND_TIMER)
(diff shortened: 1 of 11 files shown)


roborock/devices/traits/dnd.py
```diff
@@ -0,0 +1,42 @@
...
+
+    name = "do_not_disturb"
+
+    def __init__(self, rpc_channel: Callable[[], V1RpcChannel]) -> None:
+        """Initialize the DoNotDisturbTrait."""
+        self._rpc_channel = rpc_channel
+
+    async def get_dnd_timer(self) -> DnDTimer:
+        """Get the current Do Not Disturb (DND) timer settings of the device."""
+        return await self._rpc_channel().send_command(RoborockCommand.GET_DND_TIMER, response_type=DnDTimer)
```
Lash-L: Think it makes sense to make a rpctrait that has a send command function to simplify things? 

Simplify the specific trait object as much as possible?

roborock/devices/traits/dnd.py
```diff
@@ -0,0 +1,42 @@
...
+from .trait import Trait
+
+_LOGGER = logging.getLogger(__name__)
+
+__all__ = [
+    "DoNotDisturbTrait",
+]
+
+
+class DoNotDisturbTrait(Trait):
```
Lash-L, line 22: Zooming out a bit, where do you see the data from get dnd timer being stored? In the device object? Still in the trait somehow?

allenporter: I don't know yet?

My thinking is on this next step is: the current set of commands isn't complex enough to have any real use cases, so I want to start adding them in. The whole trait syntax is not good yet, and needs to be rewritten, with a few examples. I was only thinking in the context of adding traits to the CLI for now.

One question i'm wondering if is if data needs to be stored at all here? but yeah i think it probably will.

roborock/devices/traits/dnd.py
allenporter: I don't love inheritance, would like to avoid it. If we add that then effectively the trait is a V1RpcChannel. I do agree with you this is still messy.  I've made one small change for now to not spread this everywhere. Is this ok as an incremental step?

allenporter: The specific example I worry about now is adding more constructor arguments like status trait. i'm curious to see how more of these traits are constructed or how it plays out with the other comment you have. I figure that with 3-5 examples this needs to get fixed/settled.

allenporter: I think i want to fix the combined channel complexity some other way

allenporter: I have pulled out https://github.com/Python-roborock/python-roborock/pull/475 to fix this issue behind the older API, then I can come back here and revert some of these changes. I think this will keep the trait code simple enough we can continue to use composition here.

allenporter: OK, the API changes here have been reverted.

roborock/devices/traits/dnd.py
Lash-L: I think we will need data stored somewhere for simplicity, but i'm open to other solutions.

There could be cases where one trait has multiple entities relying on it, so we don't want to call update() for each entity.

But 100% fine with what you're saying for now, fine with punting this down the road.

PR #473, opened 2025-09-13
feat: Improve session command for CLI

allenporter:
Improve the session command for the CLI to allow having commands subcommands present both on the normal CLI and the session command. This is prep before adding additional commands from additional traits.

The syntax will be an interactive session like this:

```
$ roborock session
roborock> list-devices
...
roborock> status --device_id <device_id>
```


Lash-L (approved):
Looks like a good QOL change

PR #477, opened 2025-09-14
feat: add a sound volume trait

allenporter:
Add a volume trait. The purpose is to have more trait use cases to figure out a better trait API.
This improves error response handling, found during adding the volume trait and making some mistakes.

This will  have merge conflicts with https://github.com/Python-roborock/python-roborock/pull/476 and will likely need another pass of review.


roborock/devices/traits/sound_volume.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self, rpc_channel: V1RpcChannel) -> None:
+        """Initialize the SoundVolumeTrait."""
+        self._rpc_channel = rpc_channel
+
+    async def get_volume(self) -> int:
+        """Get the current sound volume of the device."""
+        response = await self._rpc_channel.send_command(RoborockCommand.GET_SOUND_VOLUME)
+        if isinstance(response, list) and response:
+            return int(response[0])
```
Lash-L: I think this is something we should definitely try to figure out in the future system. I don't think we should always have to check if it is a list, I think you said it is just because you changed the typing for send_command

Lash-L (dismissed):

roborock/devices/traits/sound_volume.py
allenporter: Does the API always return a list? My impression was that it does not given some of the other code. If it does always return a list, then that's good to know.

Lash-L: I honestly can't remember. I think we may always wrap it in a list? We didn't have to do this explicit check before so I can't remember what we did

allenporter: I think this always returns a list with a number in it...

I had to take out the code that always pops the data structure out of the list for the clean summary https://github.com/Python-roborock/python-roborock/pull/476/files otherwise it only returns one number from the list.

Lash-L (approved):

PR #482, opened 2025-09-16
chore: Add test coverage of end to end trait parsing from raw responses

allenporter:
Add v1 trait test coverage using the .json files. The existing tests are not testing the full logic from decoding all the way to trait parsing. This at least uses both the logic in `decode_rpc_response` and the actual payload parsing logic in the trait.

### DIFF
--- a/tests/devices/test_v1_device.py
+++ b/tests/devices/test_v1_device.py
@@ -1,8 +1,8 @@
 """Tests for the Device class."""
 
 import pathlib
-from unittest.mock import AsyncMock, Mock
 from collections.abc import Awaitable, Callable
+from unittest.mock import AsyncMock, Mock
 
 import pytest
 from syrupy import SnapshotAssertion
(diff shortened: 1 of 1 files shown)


tests/devices/test_v1_device.py
```diff
@@ -70,19 +83,45 @@ async def test_device_connection(device: RoborockDevice, channel: AsyncMock) ->
...
+)
+async def test_device_trait_command_parsing(
+    device: RoborockDevice,
+    setup_rpc_channel: AsyncMock,
+    snapshot: SnapshotAssertion,
+    trait_name: str,
+    trait_type: type[Trait],
+    trait_method: Callable[..., Awaitable[object]],
+) -> None:
     """Test the device get_status command."""
```
Lash-L:
```suggestion
    """Test the device trait command."""
```

Lash-L (dismissed):
Just had that one change

Lash-L (approved):

PR #487, opened 2025-09-20
feat: Implement L01 protocol

Lash-L:
This is *mainly* working, but since we really need to get this done asap, I figured I would go ahead and put a PR up while the rest is being figured out.


roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -121,20 +117,58 @@ async def async_disconnect(self) -> None:
...
+        ping_message = RoborockMessage(
+            protocol=RoborockMessageProtocol.PING_REQUEST,
+        )
         await self._send_message(
-            roborock_message=_PING_REQUEST_MESSAGE,
-            request_id=_PING_REQUEST_MESSAGE.seq,
+            roborock_message=ping_message,
+            request_id=ping_message.seq,
             response_protocol=RoborockMessageProtocol.PING_RESPONSE,
         )
```
Lash-L: We don't need to set ping request statically like we did. I know we talked about this before but I wasn't fully sure.

This works

roborock/version_1_apis/roborock_client_v1.py
```diff
@@ -448,9 +448,15 @@ def on_message_received(self, messages: list[RoborockMessage]) -> None:
                                     "Received unsolicited map response for request_id %s", map_response.request_id
                                 )
                 else:
+                    if data.protocol == RoborockMessageProtocol.GENERAL_RESPONSE and data.payload is None:
```
allenporter: I'd say put this at the previous level:

```
elif data.protocol == RoborockMessageProtocol.GENERAL_RESPONSE and data.payload is None:
    # Api will often ...
    continue
else:
...
```

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -121,20 +117,58 @@ async def async_disconnect(self) -> None:
         async with self._mutex:
             self._sync_disconnect()
 
-    async def hello(self):
+    def _set_l01_encoder_decoder(self):
```
allenporter: I think we can simplify a little bit. We can have one path for setting encoder/decoders shared in both the L01 case and non-L01 case? 

In the constructor, both paths work identically (connect none and ack none are None), so they can look like this:
```
    # For L01 these are updated with nonces after the first hello
    self.set_encoder_decoder(connect_none=None, ack_none=None)
```
Then in the `do_hello` case they can be called again with the none values for L01.

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -121,20 +117,58 @@ async def async_disconnect(self) -> None:
...
+            response = await self._send_message(
+                roborock_message=request,
+                request_id=request.seq,
                 response_protocol=RoborockMessageProtocol.HELLO_RESPONSE,
             )
-        except Exception as e:
-            self._logger.error(e)
+            if response.version.decode() == "L01":
+                self._ack_nonce = response.random
+                self._set_l01_encoder_decoder()
```
allenporter: So, up until this point the encoder was used without a connect nonce or an ack nonce. But then for all future messages beyond the hello, these must be used? Is that a requirement or can the connect nonce be set up front in the encoder even for the hello message?

Is `response.random` set to None in the non-L01 protocols? Does it hurt to set it on future requests for non-L01 protocols too?

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -50,7 +40,7 @@ def connection_lost(self, exc: Exception | None) -> None:
 class RoborockLocalClientV1(RoborockClientV1, RoborockClient):
     """Roborock local client for v1 devices."""
 
-    def __init__(self, device_data: DeviceData, queue_timeout: int = 4):
+    def __init__(self, device_data: DeviceData, queue_timeout: int = 4, version: str | None = None):
```
allenporter: Is the L01 here different than the product version? I don't love that this supports discovery, so could we make it so that it defaults to V1  unllss we know for a fact that it's an L01 product? or is V1 a local only thing that we don't know about until we explicitly discover? I assume we have the option to know it or it would not be passed here.

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -153,6 +187,33 @@ async def _send_command(
     ):
         if method in CLOUD_REQUIRED:
             raise RoborockException(f"Method {method} is not supported over local connection")
+        if self._version == "L01":
```
allenporter: This all seems like the responsibility of `RequestMessage.encode_message` and should live in `protocol`. Firstly because the code follow this pattern so its cleaner and unit testable, but also so that it can be reused by the other client.

roborock/version_1_apis/roborock_local_client_v1.py
Lash-L: So the RoborockProtocol() step could be used to discover devices, and when we do that, we can determine what version it is. 

However, it requires a specific port being opened and it is likely that a lot of instances wont just 'work' with it. Hence why we need the fallback. 

So in a future version of the system, we can set version for (some devices) that we discover.


i.e.(Pseudo code)
```
RoborockProtocol().listen()
await asyncio.sleep(some_timeout)
found_devices = []
for device in protocol_responses:
	setup_device(version=device.version)
	found_devices.append(device.duid)
	
for device in home_data.all_devices:
    if device.duid not in found_devices:
        setup_device(version=None)
```

roborock/version_1_apis/roborock_local_client_v1.py
Lash-L: That's very true, we don't have the connect nonce or the ack nonce yet. so doing it in the constructor is not helpful.

roborock/version_1_apis/roborock_local_client_v1.py
Lash-L: response.random is not None for non-L01 protocols. However, it is actually fine if they exist, that is a good call. It isn't used on encrypting the 1.0 payload, so those values can be anything.

As well, the hello message does not include any payload, so the value doesn't matter there as well. 

The only potential issues to setting connect_nonce upfront is that if the user reconnects, it will be the same nonce, but the vac should handle that fine as it is a new transport thread.

roborock/version_1_apis/roborock_local_client_v1.py
Lash-L: Great call, as it turns out all the logic was similar enough to just carry over with a small change.

roborock/version_1_apis/roborock_local_client_v1.py
allenporter: So this is different from the product version info? That is, today we can determine v1, a01, a02 etc from the provided product info. This needs a different type of version discovery, then? I think version may be overloaded and it's worth disambiguating if that is the case.

Do we want to consider giving this a different name like "message version" or some other name given we already have another existing version concept?

I think having an enum for these could also be useful given there are only two supported i think.

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -121,20 +110,56 @@ async def async_disconnect(self) -> None:
...
+                raise RoborockException(f"Failed to connect to device with protocol {self._version}")
+        else:
+            # try 1.0, then L01
+            if not await self._do_hello("1.0"):
+                if not await self._do_hello("L01"):
+                    raise RoborockException("Failed to connect to device with any known protocol")
 
     async def ping(self) -> None:
+        # Realistically, this should be set here, but this is to be safe and for typing.
+        version = b"1.0" if self._version is None else self._version.encode()
```
allenporter: How about moving this to a property shared between the two callers of send_message? The logic is inversed in each so hard to tell if its different on purpose, but i think its basically the same thing just with an arbitrary ordering.

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -121,20 +110,56 @@ async def async_disconnect(self) -> None:
         async with self._mutex:
             self._sync_disconnect()
 
-    async def hello(self):
+    def _set_encoder_decoder(self):
+        """Updates the encoder decoder. For L01 these are updated with nonces after the first hello."""
```
allenporter: This is now always true that it updates with the ack nonce after hello, but still worth preserving the fact that this is needed for L01

roborock/version_1_apis/roborock_local_client_v1.py
```diff
@@ -121,20 +110,56 @@ async def async_disconnect(self) -> None:
...
+        self._encoder = create_local_encoder(self.device_info.device.local_key, self._connect_nonce, self._ack_nonce)
+        self._decoder = create_local_decoder(self.device_info.device.local_key, self._connect_nonce, self._ack_nonce)
+
+    async def _do_hello(self, version: str) -> bool:
+        """Perform the initial handshaking."""
+        self._logger.debug(f"Attempting to use the {version} protocol for client {self.device_info.device.duid}...")
```
allenporter: %s is preferred over f strings when logging to be slightly more efficient

allenporter (changes requested):

allenporter (approved):

PR #489, opened 2025-09-21
chore: Overhaul new device trait interfaces

allenporter:
Update traits to have a simpler syntax, but some runtime hackery.

This is an example for how to interact with the device with the new syntax:

```
    device_manager = await context.get_device_manager()
    device = await device_manager.get_device(device_id)

    if device.v1_properties is None:
        raise RoborockException(f"Device {device.name} does not support V1 protocol")

    status_trait = device.v1_properties.get_status
    await status_trait.refresh()
    print(status.state)
```

Common v1 properties are assembled into a single trait of properties and we can add more groups based on supported features (e.g. map related functionality or something else)


roborock/devices/device_manager.py
```diff
@@ -154,26 +146,19 @@ async def create_device_manager(
...
         match device.pv:
             case DeviceVersion.V1:
-                channel = create_v1_channel(user_data, mqtt_params, mqtt_session, device, cache)
-                traits.append(StatusTrait(product, channel.rpc_channel))
-                traits.append(DoNotDisturbTrait(channel.rpc_channel))
-                traits.append(CleanSummaryTrait(channel.rpc_channel))
-                traits.append(SoundVolumeTrait(channel.rpc_channel))
+                v1_channel = create_v1_channel(user_data, mqtt_params, mqtt_session, device, cache)
+                channel = v1_channel
+                traits.extend(v1.create_v1_traits(product, v1_channel.rpc_channel))
```
Lash-L: Can't we just set channel = create_v1_channel rather than the temp v1_channel object.

Lash-L: Same below

Lash-L: With this refactor, we no longer need to extend right? We can just outright set traits = the creation of traits.

roborock/devices/traits/v1/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, product: HomeDataProduct, rpc_channel: V1RpcChannel) -> None:
+        """Initialize the V1TraitProps with None values."""
+        self.status = StatusTrait(product)
+
+        for item in fields(self):
+            if (trait := getattr(self, item.name, None)) is None:
+                trait = item.type()
+                setattr(self, item.name, trait)
+            trait._rpc_channel = rpc_channel
```
Lash-L, line 52: A comment here would be helpful. I get what is happening, but as this is a bit hacky, it would probably be good to have a super quick comment of what is happening.

roborock/devices/traits/v1/properties.py
```diff
@@ -0,0 +1,95 @@
+"""Traits for properties of Roborock devices using the V1 API."""
```
Lash-L: Does this make sense to be a folder of property traits?

I know one issue in the past (i.e. containers.py) is that so much gets put into it that the file gets huge and is hard to really understand.

I recon there will be a good bit of traits in here and this file will probably get long, does it make sense to seperate each property trait into it's own file?

roborock/devices/traits/v1/properties.py
```diff
@@ -0,0 +1,95 @@
...
+# dynamically or somehow make them less code.
+
+
+@dataclass
+class SoundVolume(common.RoborockValueBase):
+    """Dataclass for sound volume."""
+
+    volume: int | None = field(default=None, metadata={"roborock_value": True})
+    """Sound volume level (0-100)."""
+
```
Lash-L: Is it more simple to just make a container object for this?

Lash-L: I see that there might be more going on here than I thought

allenporter: Yeah i don't think this is quite right yet and we should add a few more.  I have this TODO here that says "This is currently the pattern for holding all the commands that hold a single value, but it still seems too verbose. Maybe we can generate these dynamically or somehow make them less code."

Maybe there's a way we can stick the `commands` on this dataclass then register it. How about in a follow up CL i add like 4 more of these then make a better pattern?

roborock/devices/traits/v1/properties.py
allenporter: Yes, agree, i think merging them was a bad idea so split back into separate files like it was before.

roborock/devices/device_manager.py
allenporter: Updated.

allenporter: I thought typing needed this but I can see that is not the case, fixed.

roborock/devices/traits/v1/__init__.py
allenporter: Updated.

roborock/devices/traits/v1/properties.py
Lash-L: Sounds great

roborock/devices/traits/traits_mixin.py
```diff
@@ -31,20 +31,20 @@ class TraitsMixin:
     zeo: a01.ZeoApi | None = None
     """Zeo API, if supported."""
 
-    b01_properties: b01.B01PropsApi | None = None
+    b01_props_api: b01.B01PropsApi | None = None
```
Lash-L: I think it makes sense for b01_props_api and v1_propertieis to have the same naming scheme. Unless you have some reasoning otherwise

roborock/devices/traits/v1/clean_summary.py
```diff
@@ -0,0 +1,29 @@
...
+class CleanSummaryTrait(CleanSummary, common.V1TraitMixin):
+    """Trait for managing the clean summary of Roborock devices."""
+
+    command = RoborockCommand.GET_CLEAN_SUMMARY
+
+    @classmethod
+    def _parse_type_response(cls, response: common.V1ResponseData) -> Self:
+        """Parse the response from the device into a CleanSummary."""
+        if isinstance(response, dict):
+            return CleanSummaryTrait.from_dict(response)  # type: ignore[return-value]
```
Lash-L: hmm, could be in a follow up, but are we able to fix the typing warnings here?

roborock/devices/traits/traits_mixin.py
allenporter: Yes, agreed, done.

roborock/devices/traits/v1/clean_summary.py
allenporter: Thanks for pushing on this. I couldn't figure it out before, but found out the solution to use `cls()` to create the object which will ensure subclasses also work.

allenporter (commented):
PTAL, ready for another look.

PR #490, opened 2025-09-21
fix: cycle through iot urls

Lash-L:
So - new accounts only give back the proper iot url on their specific regions iot url. Previously, if you hit any iot url, you would get your data back.

The options are either

a) Have the user specific what region to use, which is not clear for the user as they are a bit random sometimes. Some of Africa seems to be US and some seems to be EU. HK is US and mainland china is CN. Japan is US. A lot of randomness that a user couldn't really guess.

or 
B) cycle through all of them until we get the right answer.

So i went with B.


roborock/web_api.py
```diff
@@ -52,37 +51,49 @@ class RoborockApiClient:
...
+                    elif response_code == 1001:
+                        raise RoborockMissingParameters(
+                            "You are missing parameters for this request, are you sure you entered your username?"
+                        )
+                    else:
+                        _LOGGER.warning(
+                            "Failed to get base url for %s with the following context: %s", self._username, response
+                        )
+                if response["data"]["countrycode"] is not None:
+                    self._country_code = response["data"]["countrycode"]
```
allenporter: Don't love methods that have side effects like this. Can this return a `@dataclass` instead of a string?

I think perhaps the base url and url cycling logic may also be better served in the caller (e.g. don't call this function unless trying to get the url and country etc). I think since there a few variables about what we have (url, country codes) vs not it may be good to do that in one spot.

allenporter (changes requested):

roborock/web_api.py
Lash-L: I get the first part, can you clarify what you mean for this part?

> I think perhaps the base url and url cycling logic may also be better served in the caller (e.g. don't call this function unless trying to get the url and country etc). I think since there a few variables about what we have (url, country codes) vs not it may be good to do that in one spot.

Lash-L: And as extra clarification, none of the future steps will work without the url cycling. It isn't just the country code and country that we are needing, we also need to make sure we are on the right url.

allenporter: I'm wondering if in the caller having all this together might be helpful:

```
base_url = self.base_url
if not base_url:   # This check is no longer in the function
    response = None
    for iot_url in (....):
        try:
            response = await self._get_base_url(iot_url)
        execept RoborockUrlException:
            continue
    if response is None:
        raise RoborockNoResponseFromBaseURL(...)
    base_url = response.base_url
    # Overwrite caller provided params?
    country_code = response.country_code
   ....
```

Lash-L: hmm. I feel like that may be a lot to put in the caller, especially considering we call get_base_url() in basically every web api call. Lots of duplicated logic.

So long as I am understanding correctly

Lash-L: Just made a change, let me know if you like this any better @allenporter

allenporter: Do we need to cycle through the URLS in the other scenarios? it seems like in those cases we'll have the url on the base object passed into the constructor.

Generally, I was wondering in the longer term if this api object may benefit from being split into two api objects. One for managing authentication for the first time when you have nothing and one for API stuff after authentication and initial information.

allenporter: (I do think what you have is an improvement)

Lash-L: Splitting this api may make sense, I'd have to think about it some more. 

In most cases, base url is not passed into the constructor, that's more of an edge case. In the past you could hit the api for any region with your email and you'd get the correct region app back. That is no longer the case. You have to hit the right region endpoint and the only way to tell if you're correct is to see if you got country info back. Does that make sense?

Sorry typed from my phone

allenporter: Should we always be passing it in once we discover the iot url so we don't have to guess every time?

Lash-L: Like caching to a disk? That could make sense. 

I worry a little about the api changing
i.e. right now old accounts still work with any url, buy theoretically they could br changed to only work with their specific region like the nee accounts, so if we cached it and it changed, we might fail, so it just needs some logic there to dismiss a bad cache

allenporter: I just mean saving it in the config entry like in the base url code appears to do. you mentioned it as an edge case, but its part of the startup code in home assistant so i assume there is some contract already about needing to save it.

Lash-L: oh yes, that is very true. That would be a good change to make on the HA side

Lash-L: We actually already do pass in the base url in init, so this shouldn't be a problem on the HA side. I think this should be all good to go then? Once this is in and #497 is in, I will do a version bump which will close a good bit of issues

Followed by updating the code login.

So long as we get that tomorrow, I should be able to get everything done on the HA side before beta cutoff.

roborock/web_api.py
```diff
@@ -49,41 +58,75 @@ class RoborockApiClient:
...
-            url_request = PreparedRequest(self._default_url, self.session)
-            response = await url_request.request(
-                "post",
-                "/api/v1/getUrlByEmail",
-                params={"email": self._username, "needtwostepauth": "false"},
+        self._iot_login_info: IotLoginInfo | None = None
+
+    async def _get_iot_login_info(self) -> IotLoginInfo:
+        if self._iot_login_info is None:
+            valid_urls = (
```
allenporter: Extract this list to a Constant e.g. `BASE_URLS`

roborock/web_api.py
```diff
@@ -49,41 +58,75 @@ class RoborockApiClient:
...
-    async def _get_base_url(self) -> str:
-        if not self.base_url:
-            url_request = PreparedRequest(self._default_url, self.session)
-            response = await url_request.request(
-                "post",
-                "/api/v1/getUrlByEmail",
-                params={"email": self._username, "needtwostepauth": "false"},
+        self._iot_login_info: IotLoginInfo | None = None
+
+    async def _get_iot_login_info(self) -> IotLoginInfo:
```
allenporter, line 77: Have you thought about how you'd test this?

allenporter (changes requested):

roborock/web_api.py
Lash-L: @gemini-cli Suggest a test for cycling through the urls in _get_iot_login.

Make one test where the first api call returns data that fails this function and the second one succeeds

Lash-L: idk how to get gemini to work correctly. Maybe it is not possible with the free tier, I might revert

Lash-L:
@allenporter I added a test. Sorry I kind of dropped the ball on this, maybe we can try to get it in 2025.10.1

tests/test_web_api.py
```diff
@@ -71,3 +74,99 @@ async def test_code_login_v4_flow(mock_rest) -> None:
...
+        re.compile("https://usiot.roborock.com/api/v1/getUrlByEmail.*"),
+        status=200,
+        payload={
+            "code": 200,
+            "data": {"url": "https://usiot.roborock.com", "country": None, "countrycode": None},
+            "msg": "Success",
+        },
+    )
+
+    # 2. Mock EU URL to return a server-side error code
```
allenporter: This appears to be returning success. is this a bug or stale comment?

allenporter (changes requested):

tests/test_web_api.py
Lash-L: stale comment, thanks

allenporter (approved):

snagglewest:
Is there a timeline when this fix might hit a release?  We just bought a Qrevo CurvX (in part for the HA integration) and we got hit the "no account associated with the email you entered" running V2025.10.1

Lash-L:
> Is there a timeline when this fix might hit a release?  We just bought a Qrevo CurvX (in part for the HA integration) and we got hit the "no account associated with the email you entered" running V2025.10.1

10.2!

PR #491, opened 2025-09-21
fix: improve new v1 apis to use mqtt lazily and work entirely locally

allenporter:
The new v1 channel will:
- prefer local connection, using cached network information
- establish mqtt connections on demand only when needed on first subscription. we can decide to poke the mqtt session explicitly if we prefer, but it won't be done by the device manager now
- refresh network information on the first failed attempt, then once every 12 hours
- retry local connections with backoff

This makes the local background task (per device0 now work similarly to the mqtt session task.

The background logic is difficult to unit test without races, so it is not tested. I will do additional manual testing of complex scenarios for errors when this is integrated into Home Assistant.

### DIFF
--- a/roborock/devices/v1_channel.py
+++ b/roborock/devices/v1_channel.py
@@ -3,7 +3,8 @@
 This module provides a unified channel interface for V1 protocol devices,
 handling both MQTT and local connections with automatic fallback.
 """
-
+import asyncio
+import datetime
 import logging
 from collections.abc import Callable
 from typing import TypeVar
@@ -22,7 +23,12 @@ from .cache import Cache
 from .channel import Channel
 from .local_channel import LocalChannel, LocalSession, create_local_session
 from .mqtt_channel import MqttChannel
-from .v1_rpc_channel import PickFirstAvailable, V1RpcChannel, create_local_rpc_channel, create_mqtt_rpc_channel
+from .v1_rpc_channel import (
+    PickFirstAvailable,
+    V1RpcChannel,
+    create_local_rpc_channel,
+    create_mqtt_rpc_channel,
+)
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


roborock/devices/v1_channel.py
```diff
@@ -148,24 +176,75 @@ async def _get_networking_info(self) -> NetworkInfo:
...
+                local_connect_failures += 1
+                if local_connect_failures > 1:
+                    await asyncio.sleep(reconnect_backoff.total_seconds())
+                    reconnect_backoff = min(reconnect_backoff * RECONNECT_MULTIPLIER, MAX_RECONNECT_INTERVAL)
+
+                # First failure refreshes cache. Subsequent failures use the cache
+                # until the refresh interval expires.
+                use_cache = True
+                if local_connect_failures == 1:
+                    use_cache = False
```
Lash-L: could simplify by setting use_cache = True inside the if statement for >1. One fewer if statement

Lash-L: although i guess you'll have to move that elif below.

You don't have to do this if you don't want to. If you think it hurts readability or understandability, just ignore

Lash-L (approved):
Honestly looks good, one small nitpick, feel free to ignore.

> establish mqtt connections on demand only when needed on first subscription. we can decide to poke the mqtt session explicitly if we prefer, but it won't be done by the device manager now

We may want to poke it explicitly in HA for the sake of getting protocol updates, but I think this good library wise and we can figure out protocol updates later.

roborock/devices/v1_channel.py
allenporter: Yeah i looked at doing that before, but its like this because there are a couple scenarios for deciding if cache should be used that it makes sense to do them at once.  So, I decided to instead extract to a helper function so at least this flow is a little easier to read.

Issue #485, opened 2025-09-19
`RoborockApiClient` cannot find newly created account – Error from Home Assistant Integration


deconstructionalism:
Yesterday, I created a new Roborock account (**new household account**) via the app and have transferred our QV 35S to it. I am doing this to consolidate various smart device apps onto a single household email account for ease of portability, instead of having them linked to my personal email on a previous account (**personal email account**). The **personal email account** was set up months ago and allows me to connect it to Home Assistant, which [uses `python-roborock` to connect to Roborock devices ](https://github.com/home-assistant/core/blob/dev/homeassistant/components/roborock/config_flow.py).

The **personal email account** allows the logic below to work, effectively sending a code to the account email. The **new household** account gives an error (below). Both accounts allow login using the app as well as the commercial Roborock site, but the **new household account** fails to authenticate via this package.

## Code
```python
import asyncio

from roborock.web_api import RoborockApiClient

async def main():
    web_api = RoborockApiClient(username="<acount_email>")
    await web_api.request_code()
    code = input("What is the code?")
    user_data = await web_api.code_login(code)

asyncio.run(main())
```

## Error using **new household account**

```
Traceback (most recent call last):
  File "/Users/simulacrean/test.py", line 12, in <module>
    asyncio.run(main())
  File "/Users/simulacrean/.pyenv/versions/3.10.9/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/simulacrean/.pyenv/versions/3.10.9/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/Users/simulacrean/test.py", line 8, in main
    await web_api.request_code()
  File "/Users/simulacrean/.pyenv/versions/3.10.9/lib/python3.10/site-packages/roborock/web_api.py", line 105, in request_code
    raise RoborockAccountDoesNotExist("Account does not exist - check your login and try again.")
roborock.exceptions.RoborockAccountDoesNotExist: Account does not exist - check your login and try again.
```

deconstructionalism:
> @Lash-L The PR #486  does not solve my issue as I have documented in the PR's comments, however I was able to get to the bottom of it. 

## Summary

Newer US based Roborock accounts return the wrong EU base URL from the `https://euiot.roborock.com/api/v1/getUrlByEmail` API endpoint which the , leading to a failure of the auth at the login the `/api/v4/auth/email/login/code` step.

## Details

[...]

PR #492, opened 2025-09-21
feat: add method for returning country and country code

deconstructionalism:
See [here](https://github.com/Python-roborock/python-roborock/issues/485#issuecomment-3316285251) for explanation of utlity.


Lash-L:
Hi @deconstructionalism! Thanks for the PR! But I unfortunately have a decent bit of changes in #490 that will break this logic.

Lash-L:
Feel free to remake it based off of that with a cli command!

PR #493, opened 2025-09-22
chore: add gemini default ci actions

Lash-L:
These are the default ones they recommend setting up. I disabled scheduled triage as I am just using my free api key, so I don't want it to burn through limits

Followed guide here: https://github.com/google-github-actions/run-gemini-cli#quick-start


Lash-L:
> Can you add a link in the PR description to where this comes from? Its a lot to review so mostly want to see if this is copied from somewhere or automatically created.

Done!

allenporter (approved):
Thanks, i did not review in detail, but generally happy to try it

PR #497, opened 2025-09-22
fix: broken current map logic

Lash-L:
Fixing this logic so it can be used correctly in core


allenporter (dismissed):
A test that reproduces the issue would be ideal as well.

tests/test_containers.py
```diff
@@ -481,3 +482,14 @@ def test_multi_maps_list_info(snapshot: SnapshotAssertion) -> None:
...
+
+def test_accurate_map_flag() -> None:
+    """Test that we parse the map flag accurately."""
+    s = S7MaxVStatus.from_dict(STATUS)
+    assert s.current_map == 0
+    status = copy.deepcopy(STATUS)
+    # 252 is a code for no map, it should end up being 63.
+    status["map_status"] = 252
+    s = S7MaxVStatus.from_dict(status)
+    assert s.current_map == 63
```
allenporter: How about we:
1) Define `NO_MAP = 63`
2) Return `None` when `current_map` is `NO_MAP`

(Exposing 63 seems like too low level of a detail)

allenporter (changes requested):
Thank you, test looks great.

tests/test_containers.py
```diff
@@ -481,3 +482,14 @@ def test_multi_maps_list_info(snapshot: SnapshotAssertion) -> None:
...
     assert deserialized == snapshot
+
+
+def test_accurate_map_flag() -> None:
+    """Test that we parse the map flag accurately."""
+    s = S7MaxVStatus.from_dict(STATUS)
+    assert s.current_map == 0
+    status = copy.deepcopy(STATUS)
+    # 252 is a code for no map, it should end up being 63.
+    status["map_status"] = 252
```
allenporter:
```suggestion
    status = {
        **STATUS,
        "map_status": 252,  # Code for no map
    }
```

tests/test_containers.py
```diff
@@ -481,3 +482,14 @@ def test_multi_maps_list_info(snapshot: SnapshotAssertion) -> None:
...
+    s = S7MaxVStatus.from_dict(status)
```
allenporter:
```suggestion
    s = S7MaxVStatus.from_dict({
	        **STATUS,
	        "map_status": 252,  # Code for no map
	})
```

allenporter (approved):

PR #501, opened 2025-09-28
feat: add pdoc for leveraging python docstrings for documentation

allenporter:
Add pdoc, which is similar to the existing documentation, but uses existing python docstrings as the basis for documentation, then can also include markdown files. This PR creates a pdoc documentation site within the existing documentation directory for initial testing. 

Feel free to reject if this is not useful given we already have some documentation.  This may be considered just personal preference.

Example site: https://allenporter.github.io/ical/ical.html
See https://pdoc.dev/docs/pdoc.html for more background on how pdoc works.


Lash-L (approved):
I think this definitely makes more sense than what we have now. It is the best chance of docs actually getting updated and helps make better documentation in the code.

Are we able to include something like the RoborockCommand enum? Can we even get more granular and attempt to doc each entry in a enum? Or is that above what pydoc is capable of

PR #502, opened 2025-09-28
feat: add support for getting and reseting consumables

allenporter:
Add a trait that reads and writes consumables. Add tests for the protocol parsing snapshot tests and some of the trait tests. Add the commands to the CLI.

### DIFF
--- a/roborock/cli.py
+++ b/roborock/cli.py
@@ -21,6 +21,7 @@ roborock> list-devices
 roborock> status --device_id <device_id>
 ```
 """
+
 import asyncio
 import datetime
 import functools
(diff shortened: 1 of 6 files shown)


Lash-L (commented):
Different devices have different consumables - how can we adapt this as needed?

Theoretically, when you get the result from the get consumable status you know which ones can be reset. Not sure if it is a needed check or not, just something I wanted to pose

Lash-L:
> > Got it, so they're either present or absent once this has been refreshed?
> > Is that the only place this is revealed from features?
> 
> I'm not sure, I would have to look through the code and see

filter_work_time, main_brush_work_time,side_brush_work_time, sensor_dirty_time are enabled for all devices


filter_element_work_time is enabled if product nickname is Tanos.
mopSwabSupplies is enabled if product nickname is not tanosSE, RubySE, or RubyPlus
strainer_work_times, cleaning_brush_work_times, dust_collection_work_times,  dust_bag_work_times, floor_cleaning_fluid is enbled if the dock is enabled based on a property of the dock. I don't have this logic anywhere.

allenporter:
OK. I've added a comment to the class saying that once refreshed you can tell what is supported by checking not none.


From a client perspective, from home assistant pov: I'm not sure how annoying this will be in the integration, but it may be ok to just include this on the first refresh of the coordinator, then continue to check for none when setting up sensors.

Lash-L (approved):
I'll do a follow up expanding this with some more . But it looks good for now

PR #503, opened 2025-09-28
fix: Update containers that __post_init__ to use properties

allenporter:
Update all computed properties to use properties rather than setting during `__post_init__`
This is done since with the new v1 api we reuse attributes and mutate them on update, so this allows the properties to not get stale with values from a previous post init.


Lash-L (dismissed):
Should properties be included in the snapshot? Genuine question, no real opinion either way

allenporter:
> Should properties be included in the snapshot? Genuine question, no real opinion either way

Good idea. Updated to rewrite repr to include the attributes as well.

roborock/containers.py
```diff
@@ -193,6 +207,9 @@ def end_time(self) -> datetime.time | None:
             else None
         )
 
+    def __repr__(self) -> str:
+        return _attr_repr(self, ["start_time", "end_time"])
```
Lash-L: Would this be simpler/ avoid maintenance if we just added a iterator in the roborockbase, checking if each element is a property and if so we add it to _attr_repr?

Lash-L: More computational overhead though

allenporter: I tried to combine with the existing logic so it won't be more complex overall. The only thing i can't seem to make better is just sticking this in the base class without additional copies in the subclasses. dir in the parent doesn't find the child properties for some reason, but maybe something we can figure out in a future pr

roborock/containers.py
```diff
@@ -107,22 +107,32 @@ def _decamelize(s: str):
     return re.sub("([A-Z]+)", "_\\1", s).lower()
 
 
-def _attr_repr(obj: Any, attrs: list[str]) -> str:
-    """Return a string representation of the object including specified attributes."""
+def _attr_repr(obj: Any) -> str:
```
Lash-L, line 110: could this just become the default __repr__ for RoborockBase?

allenporter: I tried but it seemed to not see child attributes for some reason. (see comment above)

allenporter: (It could be that it is specific to the traits with mixins since those are the ones I am testing, but other way, this did seem to address that)

Lash-L: Ah I see. Thanks, I see your other comment now. Yep we can revisit in the future if needed.

Lash-L:
Already started doing a version bump on core so i can fix the url issue. But if you want this included, lmk and i can add it. Or we can do a follow up

PR #511, opened 2025-10-02
feat: add dynamic changing status

Lash-L:
This is still a WIP. But there are a few things that I need to figure out and I thought it would be good to get in front of you

### DIFF
--- a/roborock/containers.py
+++ b/roborock/containers.py
@@ -10,6 +10,15 @@ from enum import Enum
 from functools import cached_property
 from typing import Any, NamedTuple, get_args, get_origin
 
+from .clean_modes import (
+    CleanRoutes,
+    VacuumModes,
+    VacuumModesOld,
+    WaterModes,
+    get_clean_modes,
+    get_clean_routes,
+    get_water_modes,
+)
 from .code_mappings import (
     SHORT_MODEL_TO_ENUM,
     RoborockCategory,
@@ -19,7 +28,6 @@ from .code_mappings import (
     RoborockDockTypeCode,
     RoborockDockWashTowelModeCode,
     RoborockErrorCode,
-    RoborockFanPowerCode,
     RoborockFanSpeedP10,
     RoborockFanSpeedQ7Max,
     RoborockFanSpeedQRevoCurv,
@@ -34,7 +42,6 @@ from .code_mappings import (
     RoborockFinishReason,
     RoborockInCleaning,
     RoborockModeEnum,
-    RoborockMopIntensityCode,
     RoborockMopIntensityP10,
     RoborockMopIntensityQ7Max,
     RoborockMopIntensityQRevoCurv,
@@ -46,7 +53,6 @@ from .code_mappings import (
     RoborockMopIntensityS8MaxVUltra,
     RoborockMopIntensitySaros10,
     RoborockMopIntensitySaros10R,
-    RoborockMopModeCode,
     RoborockMopModeQRevoCurv,
     RoborockMopModeQRevoMaster,
     RoborockMopModeQRevoMaxV,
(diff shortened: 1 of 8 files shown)


roborock/containers.py
```diff
@@ -574,6 +557,34 @@ class Saros10Status(Status):
...
 
+def get_custom_status(features: DeviceFeatures, region: str) -> "DeviceStatus":
+    _available_fan_speeds = get_clean_modes(features)
+    _available_fan_speed_mapping = {fan.code: fan.name for fan in _available_fan_speeds}
+    _available_water_modes = get_water_modes(features)
+    _available_water_modes_mapping = {mop.code: mop.name for mop in _available_water_modes}
+    _available_mop_routes = get_clean_routes(features, region)
+    _available_mop_routes_mapping = {route.code: route.name for route in _available_mop_routes}
+
+    class DeviceStatus(Status):
```
Lash-L, line 568: I think this needs to be a protocol or something? Was hoping you may have some ideas on how to improve it

roborock/devices/traits/v1/status.py
```diff
@@ -13,12 +13,12 @@ class StatusTrait(Status, common.V1TraitMixin):
     def __init__(self, product_info: HomeDataProduct) -> None:
         """Initialize the StatusTrait."""
         self._product_info = product_info
+        self._status_type = get_custom_status(self.device_info.device_features, self.device_info.region)
```
Lash-L, line 16: This is where it gets weird - To determine the status type, we need the device features, which we get from init status, which we need to be connected for.

Hoping you may have some ideas here

roborock/containers.py
allenporter: I can see why this is set up this way. 

Does it make sense for DeviceStatus to be a top level object with a DeviceFeatures + region as input to constructor?

roborock/devices/traits/v1/status.py
allenporter: Can device features be provided similar to how HomeDataProduct or network info are supplied? Do it up front, and we provide an interface for caching it.

roborock/containers.py
Lash-L: So the issue is right now, every instance of status that we do we use from_dict()

So we'd need to add it to from_dict() and constructor. Theoretically fine, but potentially bloats from_dict() calls for status so i wasn't sure.

allenporter: This seems similar to the traits that need to parse room names. We basically have a hacky dance to make this work:

- RoomsTrait is constructed with home data passed to constructor and no values initially
- On update, it cals _parse_data which parses the response object and merges with the home data
- Parent class uses setattr to update the object

Perhaps this could just be added to `StatusTrait`?

roborock/devices/traits/v1/status.py
Lash-L: That's what i thought about doing but I don't know if it is possible. The Trait objects are created synchronously before we are subscribed and we could send a message to get the init status.

allenporter: Maybe making a separate trait for now can work, then once its loaded it can be used? This could override `refresh` to refresh it first before status is loaded the first time.

Separately: Is making it part of `Status` required?

Lash-L: > Separately: Is making it part of Status required?

Not explicitly, no, but i figured it would make sense. but open to doing it differently.

I could make a device features trait that gets the init status/whatever else it needs and other things rely on it? As we will need it to determine what is supported when we start doing more complicated traits? Thoughts on that?

allenporter: Seems worth a try? That way we can get it working then next figure out how to best use it?

this thing of combining traits together with dependencies is kinda weird but kind of simple too but not sure if it will bite us later, but may be an easy way to string things together.

PR #516, opened 2025-10-04
feat: Add v1 rooms support to the device traits API

allenporter:
Add a trait that can return rooms and handles all the mapping from iot id to room name.
Later, we'll add an interface that helps with connecting maps and rooms together.


roborock/devices/traits/v1/rooms.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def room_map(self) -> dict[int, NamedRoomMapping]:
+        """Returns a mapping of segment_id to NamedRoomMapping."""
+        if self.rooms is None:
+            return {}
+        return {room.segment_id: room for room in self.rooms}
+
+
+class RoomsTrait(Rooms, common.V1TraitMixin):
```
allenporter, line 42: This is currently using the local API but let me know if I missed that this needs to use cloud.

Lash-L: Local is good

roborock/devices/traits/v1/rooms.py
```diff
@@ -0,0 +1,93 @@
...
+
+@dataclass
+class NamedRoomMapping(RoomMapping):
+    """Dataclass representing a mapping of a room segment to a name.
+
+    The name information is not provided by the device directly, but is provided
+    from the HomeData based on the iot_id from the room.
+    """
+
+    name: str | None = None
```
Lash-L: We can just make this str right? Avoid having to do future type checking since we set it to Unknown by default

roborock/devices/traits/v1/rooms.py
```diff
@@ -0,0 +1,93 @@
...
+def _extract_segment_pairs(response: list) -> list[tuple[int, str]]:
+    """Extract segment_id and iot_id pairs from the response.
+
+    The response format can be either a flat list of [segment_id, iot_id] or a
+    list of lists, where each inner list is a pair of [segment_id, iot_id]. This
+    function normalizes the response into a list of (segment_id, iot_id) tuples
+
+    NOTE: We currently only have samples of the list of lists format in
+    tests/protocols/testdata so improving test coverage with samples from a real
+    device with this format would be helpful.
```
Lash-L: [[16, '2537178', 6], [17, '2537175', 14], [18, '2537174', 13], [19, '2537176', 14], [20, '10655627', 12], [21, '2537145', 2], [22, '2537147', 12]]

Here is an example

allenporter: Thanks, that caught a bug.  I've added this to the test then made up what I think the raw format probably looks like given that.

allenporter:
Whoops, accidentally merged this with the other outstanding PR

Lash-L (approved):

PR #524, opened 2025-10-10
chore: add python 3.14 for tests

Lash-L:
Remove python 3.12 and 3.13 and add python 3.14

The idea here is that we should make sure we can still run on our lowest possible version and our max possible version.

I believe we are doing things that require 3.11.

If it works on 3.11 and it works on 3.14, it is highly unlikely it would not work on 3.12 or 3.13

By reducing the test count, hopefully it will be less likely for us to have a random timeout test.

### DIFF
--- a/tests/test_queue.py
+++ b/tests/test_queue.py
@@ -6,7 +6,7 @@ from roborock.exceptions import VacuumError
 from roborock.roborock_future import RoborockFuture
 
 
-def test_can_create():
+async def test_can_create():
     RoborockFuture(1)
 
 


allenporter (approved):

.github/workflows/ci.yml
```diff
@@ -36,8 +36,7 @@ jobs:
       matrix:
         python-version:
           - "3.11"
-          - "3.12"
-          - "3.13"
+          - "3.14"
```
allenporter, line 39: Should we leave 3.12 and 3.13? drop 3.11? seems wrong to drop 3.12/3.13

Lash-L: My thought is that as long as we can support the lowest possible version, we should and we should know when we have done something that breaks support. However, we should also know if we need to do something to add support for the latest version.


3.12/3.13 server the same purpose as the newest/oldest. If something was removed, it would be removed in 3.12, it is removed in 3.14

If we are using a new feature that was added in 3.14, it wouldn't be possible in 3.13 or 3.11

I can add them back, but this was my thought process, what are your thoughts?

allenporter: ok i can buy that, just haven't ever seen it done that way.

Lash-L: Me neither lol - I just didn't want to drastically increase our test count while we have a sometimes flaky test. Thought through it and figured this made enough sense at least for now. We can always add some them back.

PR #526, opened 2025-10-10
Add a Home trait that is responsible for caching information about maps and rooms

allenporter:
This handles "discovery" of doing a pass through all the maps for initial population. After that, it will only update information for the active maps. This is largely pulling existing home assistant logic into this library, but will improve metadata caching so we can avoid doing discovery on every startup (a feature we've wanted for awhile).

This will be updated to also handle map content storage in a followup.

### DIFF
--- a/roborock/devices/traits/v1/home.py
+++ b/roborock/devices/traits/v1/home.py
@@ -13,14 +13,16 @@ import asyncio
 import logging
 from typing import Self
 
+from roborock.code_mappings import RoborockStateCode
 from roborock.containers import CombinedMapInfo, RoborockBase
 from roborock.devices.cache import Cache
 from roborock.devices.traits.v1 import common
-from roborock.exceptions import RoborockException
+from roborock.exceptions import RoborockDeviceBusy, RoborockException
 from roborock.roborock_typing import RoborockCommand
 
 from .maps import MapsTrait
 from .rooms import RoomsTrait
+from .status import StatusTrait
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -32,7 +34,13 @@ class HomeTrait(RoborockBase, common.V1TraitMixin):
 
     command = RoborockCommand.GET_MAP_V1  # This is not used
 
-    def __init__(self, maps_trait: MapsTrait, rooms_trait: RoomsTrait, cache: Cache) -> None:
+    def __init__(
+        self,
+        status_trait: StatusTrait,
+        maps_trait: MapsTrait,
+        rooms_trait: RoomsTrait,
+        cache: Cache,
+    ) -> None:
         """Initialize the HomeTrait.
 
         We keep track of the MapsTrait and RoomsTrait to provide a comprehensive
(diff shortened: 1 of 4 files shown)


roborock/devices/traits/v1/home.py
```diff
@@ -0,0 +1,157 @@
...
+            self._maps_trait.map_info or [],
+            key=lambda mi: mi.map_flag == self._maps_trait.current_map,
+            reverse=False,
+        )
+        _LOGGER.debug("Building home cache for maps: %s", [mi.map_flag for mi in sorted_map_infos])
+        for map_info in sorted_map_infos:
+            # We need to load each map to get its room data
+            if len(sorted_map_infos) > 1:
+                _LOGGER.debug("Loading map %s", map_info.map_flag)
+                await self._maps_trait.set_current_map(map_info.map_flag)
```
Lash-L, line 120: So, i'm not sure where it should live, but either here or in the trait, we need to check if it is in cleaning.

roborock/devices/traits/v1/home.py
```diff
@@ -0,0 +1,157 @@
...
+
+        await self._maps_trait.refresh()
+        if self._maps_trait.current_map_info is None:
+            raise RoborockException("Cannot perform home discovery without current map info")
+
+        home_cache = await self._build_home_cache()
+        _LOGGER.debug("Home discovery complete, caching data for %d maps", len(home_cache))
+        cache_data = await self._cache.get()
+        cache_data.home_cache = home_cache
+        await self._cache.set(cache_data)
```
Lash-L: We do this a couple of times, does a self._cache_set_home_cache(home_cache) make sense?

As well how often would the home cache get out of sync? Do we need to do a load and a set every time?

roborock/devices/traits/v1/home.py
```diff
@@ -0,0 +1,157 @@
...
+    def home_cache(self) -> dict[int, CombinedMapInfo] | None:
+        """Returns the map information for all cached maps."""
+        return self._home_cache
+
+    @property
+    def current_map_data(self) -> CombinedMapInfo | None:
+        """Returns the map data for the current map."""
+        current_map_flag = self._maps_trait.current_map
+        if current_map_flag is None or self._home_cache is None:
+            return None
```
Lash-L, line 163: might be good to do some explicit if current_map_flag is 63 checking. But maybe not worth the hastle as it theoretically should be covered.

roborock/devices/traits/v1/home.py
allenporter: Done.

roborock/devices/traits/v1/home.py
allenporter: Updated to raise an exception when cleaning. i added pydoc so the caller can know to catch it then decide what to do (e.g. try again later, do nothing, etc)

roborock/devices/traits/v1/home.py
allenporter: I think this is checked by status trait given your previous change and will return `None`? if that is not the case let me know.

PR #538, opened 2025-10-17
feat: Add various optional features with support for checking device features

allenporter:
This PR adds:
- Some assorted optional features, mostly associated with switches
- Ability to do device features enforcement with a new decorator (part of v1 properties api)
- New CLI for each

My devices do not have support for all of these features, so they were not all manually tested.

### DIFF
--- a/tests/devices/traits/v1/test_dnd.py
+++ b/tests/devices/traits/v1/test_dnd.py
@@ -4,13 +4,10 @@ from unittest.mock import AsyncMock
 
 import pytest
 
-from roborock.containers import AppInitStatus, DnDTimer, HomeDataProduct
-from roborock.device_features import DeviceFeatures
-from roborock.devices.cache import Cache
+from roborock.containers import DnDTimer
 from roborock.devices.device import RoborockDevice
 from roborock.devices.traits.v1.do_not_disturb import DoNotDisturbTrait
 from roborock.roborock_typing import RoborockCommand
-from tests import mock_data
 
 
 @pytest.fixture
@@ -19,8 +16,6 @@ async def dnd_trait(device: RoborockDevice) -> DoNotDisturbTrait:
     assert device.v1_properties
     assert device.v1_properties.dnd
     return device.v1_properties.dnd
-
-
 
 
 @pytest.fixture


roborock/cli.py
```diff
@@ -532,6 +540,116 @@ async def reset_consumable(ctx, device_id: str, consumable: str):
...
+@click.pass_context
+@async_command
+async def child_lock(ctx, device_id: str, enabled: bool | None):
+    """Get device child lock status."""
+    context: RoborockContext = ctx.obj
+    try:
+        trait = await _v1_trait(context, device_id, lambda v1: v1.child_lock)
+    except RoborockUnsupportedFeature:
+        click.echo("Feature not supported by device")
+        return
```
Lash-L, line 555: should this logic go inside _v1_trait so you don't have to repeat it?

Lash-L: oh although i guess you stlil need a try except here to return opposed to erroring out. Or check for None, so maybe it isn't worth it

roborock/devices/traits/v1/child_lock.py
```diff
@@ -0,0 +1,20 @@
+from roborock.containers import ChildLockStatus
+from roborock.devices.traits.v1 import common
+from roborock.roborock_typing import RoborockCommand
+
+_STATUS_PARAM = "lock_status"
+
+
+@common.requires_feature("is_set_child_supported")
```
Lash-L: So I think there are cases where multiple features need to exist for it to be supported. My gut feeling is that we should just make a custom feature that does that logic and then use that custom feature here instead of making this decorator more complicated - do you share that opinion?

roborock/devices/traits/v1/common.py
```diff
@@ -143,3 +147,29 @@ def wrapper(*args, **kwargs):
 
     cls.map_rpc_channel = True  # type: ignore[attr-defined]
     return wrapper
+
+
+def requires_feature(feature_attribute_name: str):
```
Lash-L: What is the benefit of a decorator opposed to say a class variable that defaults to None?

Not against it, just thinking about it

roborock/devices/traits/v1/device_features.py
```diff
@@ -9,9 +12,26 @@ class DeviceFeaturesTrait(DeviceFeatures, common.V1TraitMixin):
...
+    def __init__(self, product_nickname: RoborockProductNickname, cache: Cache) -> None:
         """Initialize MapContentTrait."""
         self._nickname = product_nickname
+        self._cache = cache
+
+    async def refresh(self) -> Self:
+        """Refresh the contents of this trait.
+
+        This will use cached device features if available since they do not
+        change often and this avoids unnecessary RPC calls.
```
Lash-L: Maybe specifically outline here that they would only change from a firmware update.

roborock/devices/traits/v1/device_features.py
```diff
@@ -9,9 +12,26 @@ class DeviceFeaturesTrait(DeviceFeatures, common.V1TraitMixin):
...
+        """
+        cache_data = await self._cache.get()
+        if cache_data.device_features is not None:
+            self._update_trait_values(cache_data.device_features)
+            return cache_data.device_features  # type: ignore[return-value]
```
Lash-L: I have admittedly not touched the -> Self return type. But it does feel a bit weird that we have to tell the system to ignore the typing. Is everything right here?

roborock/devices/traits/v1/child_lock.py
allenporter: That sounds like a good idea to me. The device features object can expose whatever is helpful to make the trait code simple.

roborock/devices/traits/v1/common.py
allenporter: I don't see any difference -- given this decorator is literally just setting a class variable based on the argument? In the other cases the decorator is doing something dynamic but here it is not. How about we go with the class variable since its strictly simpler and may also help with the typing below.

roborock/cli.py
allenporter: A totally reasonable option is to just ignore this and let it print the exception stack trace. We don't really have a consistent interface for this CLI e.g. how is data printed, how are errors handled, etc so we can do anything here. Maybe i'll just drop it but let me know if you have a preference.

Lash-L: Don't have a preference, was just thinking about the repeated code, but it doesn't really matter either way.

roborock/devices/traits/v1/device_features.py
allenporter: This was wrong. `_update_trait_values` merges the features into the trait and this should be returning the trait -- though really `refresh` probably should have a `None` return type.

PR #550, opened 2025-10-19
feat: Add clean record and dock related traits

allenporter:
This creates a new dock type required feature. This requires the status is fetched first initially. The dock type is added to the cache since it is not expected to change once discovered.

The clean record trait depends on the clean summary and fetches the last clean record from the summary.

Tests are added for clean record and dock summary. The dock summary tests also exercise the feature discovery and underlying modes.


roborock/devices/traits/v1/__init__.py
```diff
@@ -103,7 +154,7 @@ def __init__(
                 # We exclude optional features and them via discover_features
                 if (union_args := get_args(item.type)) is None or len(union_args) > 0:
                     continue
-                _LOGGER.debug("Initializing trait %s", item.name)
+                _LOGGER.debug("Trait '%s' is supported, initializing", item.name)
```
allenporter, line 153: Note: Updated the debug logging to be more consistent and readable across the different cases

roborock/devices/traits/v1/__init__.py
```diff
@@ -136,22 +191,79 @@ async def discover_features(self) -> None:
...
+            return dock_type in requires_dock_type
+
+        if (feature_name := getattr(trait, "requires_feature", None)) is None:
+            _LOGGER.debug("Optional trait missing 'requires_feature' attribute %s, skipping", name)
+            return False
+        if (is_supported := getattr(self.device_features, feature_name)) is None:
+            raise ValueError(f"Device feature '{feature_name}' on trait '{name}' is unknown")
+        return is_supported
+
+    async def _dock_type(self) -> RoborockDockTypeCode:
```
allenporter, line 211: Note: New logic for extracting the dock type from the cached status. I didn't see other attributes in status that should also be cached for feature discovery, but let me know if there are more you know of.

roborock/devices/traits/v1/clean_record.py
```diff
@@ -0,0 +1,69 @@
...
+            _LOGGER.debug("No clean records available in clean summary.")
+            return self
+        last_record_id = self._clean_summary_trait.records[-1]
+        response = await self.rpc_channel.send_command(self.command, params=[last_record_id])
+        new_self = self._parse_response(response)
+        self._update_trait_values(new_self)
+        return self
+
+    @classmethod
+    def _parse_type_response(cls, response: common.V1ResponseData) -> CleanRecord:
```
allenporter, line 41: Note: This is basically copied from the other code

roborock/devices/traits/v1/dock_summary.py
```diff
@@ -0,0 +1,43 @@
...
+
+from . import common
+from .dust_collection_mode import DustCollectionModeTrait
+from .smart_wash_params import SmartWashParamsTrait
+from .wash_towel_mode import WashTowelModeTrait
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class DockSummaryTrait(DockSummary, common.V1TraitMixin):
```
Lash-L: So I know this is a holdover from the old system, but I am not sure if I like the idea of a dock summary. Maybe it makes sense to keep things seperated dock vs not, but I think it adds more confusion than it is worth. I know that is more changes in the HA side though. I'm not completely sold on this and we can always remove it in the futuer though if you want to punt that conversation to the future.

roborock/devices/traits/v1/dock_summary.py
```diff
@@ -0,0 +1,43 @@
...
+        tasks: list[asyncio.Task] = [asyncio.create_task(self._dust_collection_mode.refresh())]
+        if self._wash_towel_mode is not None:
+            tasks.append(asyncio.create_task(self._wash_towel_mode.refresh()))
+        if self._smart_wash_params is not None:
+            tasks.append(asyncio.create_task(self._smart_wash_params.refresh()))
+        _LOGGER.debug("Waiting for %d Dock Summary task(s) to complete ", len(tasks))
+        await asyncio.gather(*tasks)
+        summary = DockSummary(self._dust_collection_mode, self._wash_towel_mode, self._smart_wash_params)
+        self._update_trait_values(summary)
+        return self
```
allenporter: Note: This can probably be simplified across traits in a future PR since i think we have this pattern a few times, as well as changing the return value as it could just be `None`

roborock/devices/traits/v1/__init__.py
Lash-L: that should be all we need

roborock/devices/traits/v1/dust_collection_mode.py
```diff
@@ -0,0 +1,11 @@
+"""Trait for dust collection mode."""
+
+from roborock.containers import DustCollectionMode
+from roborock.devices.traits.v1 import common
+from roborock.roborock_typing import RoborockCommand
+
+
+class DustCollectionModeTrait(DustCollectionMode, common.V1TraitMixin):
```
Lash-L, line 9: This needs some kind of way to make sure it isn't used for devices that don't have a dock I believe.

There might be a feature i can find too if that is best

roborock/devices/traits/v1/dock_summary.py
allenporter: Sounds good, removed `DockSummary`

roborock/devices/traits/v1/dust_collection_mode.py
allenporter: OK, i updated this to use a function that checks for a valid dock mode.

Lash-L:
Make sure commitlint is happy when you merge - need to see if release cycle is fixed

Lash-L (approved):

PR #551, opened 2025-10-20
feat: add data subpackage

Lash-L:
Giant change...

I was getting frustrated with the size and chaos that was .containers and .code_mappings. Let me know your thoughts here, this is a breaking change, but it is very easy to fix. Replace instances of roborock.containers or roborock.code_mappings with roborock.data

### DIFF
--- a/roborock/data/v1/v1_clean_modes.py
+++ b/roborock/data/v1/v1_clean_modes.py
@@ -1,8 +1,11 @@
 from __future__ import annotations
 
-from roborock.device_features import DeviceFeatures
+import typing
 
 from ..code_mappings import RoborockModeEnum
+
+if typing.TYPE_CHECKING:
+    from roborock.device_features import DeviceFeatures
 
 
 class VacuumModes(RoborockModeEnum):


roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -0,0 +1,53 @@
+from ..containers import RoborockBase
+
+
+class dpCleanRecord(RoborockBase):
```
allenporter, line 4: Interesting naming style here. Should start with capital?

roborock/data/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""This module is meant to hold dataclasses and codemappings for various devices and protocols."""
+
+from .b01_q7 import *
```
allenporter, line 3: These will all fit in the same namespace w/o any overlap? I'm a little unsure when we should import from `.data` vs importing from subdirectories.

Lash-L: My thought was you should typically import from the subdirectories. But that felt like a very big breaking change that we could maybe make later. For now, none of them overlap namespace wise but definitely a good future concern. Because they have shared a file in the past it hasn't been a problem as I would name them differently (I.e. ZyadStatus vs Status) but now that they are in different files it would be an easier mistake to make. 

I think removing the * import would be a good future step. And encouraging specific imports from the subdirectories. I think for now we have a lot of moving parts as is, so I wanted to leave it simple and similar to how it is now

roborock/data/b01_q10/b01_q10_containers.py
Lash-L: Yes- this is how it is named in the decompiled source code so I just matched. I can do typically Pythonic naming conventions if you think that is better. I went back and forth

Lash-L: Will follow up based on what you say - for now will merge as this isn't actually utilized anywhere yet

allenporter: Yeah i'd say do the pythonic naming in a followup.

roborock/data/__init__.py
allenporter: Sounds good to me.

PR #559, opened 2025-10-24
feat: add l01 to the new device format

Lash-L:
Rob will test this for me, if you're okay with this approach and it works for Rob, i'll update this to include tests as well, didn't want to write tests before the code was in the format that you think bests works with your design

### DIFF
--- a/roborock/devices/local_channel.py
+++ b/roborock/devices/local_channel.py
@@ -58,8 +58,9 @@ class LocalChannel(Channel):
         self._subscribers: CallbackList[RoborockMessage] = CallbackList(_LOGGER)
         self._is_connected = False
         self._local_protocol_version: LocalProtocolVersion | None = None
-        self._params = LocalChannelParams(local_key=local_key, connect_nonce=get_next_int(10000, 32767), ack_nonce=None)
-        self._update_encoder_decoder(self._params)
+        self._update_encoder_decoder(
+            LocalChannelParams(local_key=local_key, connect_nonce=get_next_int(10000, 32767), ack_nonce=None)
+        )
 
     def _update_encoder_decoder(self, params: LocalChannelParams):
         self._params = params


roborock/data/v1/v1_containers.py
```diff
@@ -578,7 +578,7 @@ class AppInitStatus(RoborockBase):
     new_feature_info_str: str
     new_feature_info_2: int | None = None
     carriage_type: int | None = None
-    dsp_version: int | None = None
```
Lash-L, line 587: unrelated change, but i threw it in as it was basic

roborock/mqtt/roborock_session.py
```diff
@@ -106,6 +106,7 @@ async def _run_task(self, start_future: asyncio.Future[None] | None) -> None:
                     # Reset backoff once we've successfully connected
                     self._backoff = MIN_BACKOFF_INTERVAL
                     self._healthy = True
+                    _LOGGER.info("MQTT Session connected.")
```
Lash-L, line 109: unrelated change, but when the system disconnected, you didn't get a clear message that it reconnected.

roborock/devices/local_channel.py
```diff
@@ -39,18 +42,83 @@ class LocalChannel(Channel):
...
+        except RoborockException as e:
+            _LOGGER.debug(
+                "Client %s did not respond or does not speak the %s protocol. %s",
+                self._host,
+                local_protocol_version,
+                e,
+            )
+            return False
+
+    async def hello(self):
```
allenporter: I think this can be a loop over protocols, sorted by whether or not they equal `self._local_protocol_version` if not none

e.g. 

```
    attempt_versions = [LocalProtocolVersion.V1, LocalProtocolVersion.L01]
    if self._local_protocol_version:
        attempt_versions.sort(key=lambda v: v == self._local_protocol_version). # Not sure if this works or needs to be reversed
    for version in attempt_versions:
        if await self._do_hello(version):
            return
    raise RoborockException("Failed to connect to device with any known protocol")
```

roborock/devices/v1_channel.py
```diff
@@ -212,6 +212,7 @@ async def _local_connect(self, *, use_cache: bool = True) -> None:
             raise RoborockException(f"Error connecting to local device {self._device_uid}: {e}") from e
         # Wire up the new channel
         self._local_channel = local_channel
+        await self._local_channel.hello()
```
allenporter: I don't think we need to expose this to the caller, it seems like an implementation detail of `connect`.  If anything fails during this period we don't want to assign the local channel and leave it marked as connected.

roborock/devices/local_channel.py
```diff
@@ -39,18 +42,83 @@ class LocalChannel(Channel):
     format most parsing to higher-level components.
     """
 
-    def __init__(self, host: str, local_key: str):
+    def __init__(self, host: str, local_key: str, local_protocol_version: LocalProtocolVersion | None = None):
```
allenporter: I'm not seeing that this is ever set in the constructor, can this be removed?

roborock/devices/local_channel.py
```diff
@@ -39,18 +42,83 @@ class LocalChannel(Channel):
...
-        self._decoder: Decoder = create_local_decoder(local_key)
-        self._encoder: Encoder = create_local_encoder(local_key)
+        self._local_key = local_key
+        self._local_protocol_version = local_protocol_version
+        self._connect_nonce = get_next_int(10000, 32767)
+        self._ack_nonce: int | None = None
+        self._update_encoder_decoder()
+
+    def _update_encoder_decoder(self):
+        self._encoder = create_local_encoder(
```
allenporter, line 67: I think it's time to convert these 3 arguments (local key, connect nonce, ack nonce) into a dataclasss that are something like `LocalChannelParams`. I think it would be nice to set these 3 values in one single operation rather than having the 3 member variables set each time. They can either be arguments to `create_local_encoder` + `create_local_decoder` or just assigned in the constructor or after successful hello. 

Doing this means there is a more clear contract of when the local channel connection is reset and what the arguments are.

Perhaps these 3 arguments via dataclass can be passed to `_update_encoder_decoder` as well, rather than set on the object.

roborock/devices/local_channel.py
```diff
@@ -39,18 +42,83 @@ class LocalChannel(Channel):
...
+            local_key=self._local_key, connect_nonce=self._connect_nonce, ack_nonce=self._ack_nonce
+        )
+        self._decoder = create_local_decoder(
+            local_key=self._local_key, connect_nonce=self._connect_nonce, ack_nonce=self._ack_nonce
+        )
         # Callback to decode messages and dispatch to subscribers
         self._data_received: Callable[[bytes], None] = decoder_callback(self._decoder, self._subscribers, _LOGGER)
 
+    async def _do_hello(self, local_protocol_version: LocalProtocolVersion) -> bool:
```
allenporter: It seems like it would be nice to remove side effects from this method. Can this be just about sending the hello, then the caller is response for converting the response to updated encoder params? Right now there are multiple methods with side effects (_do_hello, _update_encoder_decoder) and i would like to have fewer in the stack.

psuedocode but more something like this pattern:
```
async def hello()
    result = self.do_hello()
    self.update_encoder(params(result))
```

This gets easier to do given my suggestions above (args to update encoder) and below (reducing number of cases of calling _do_hello to 1) so make that more possible.

Lash-L:
Rob tested and it did work. (Pre my changes, but they shouldn't impact anything)

roborock/devices/local_channel.py
```diff
@@ -143,6 +154,9 @@ async def connect(self) -> None:
         except OSError as e:
             raise RoborockConnectionException(f"Failed to connect to {self._host}:{_PORT}") from e
 
+        # Perform protocol negotiation
+        await self._hello()
```
allenporter: I think we may need to manage `self._is_connected` here in case this fails. I guess the behavior is undefined for reconnecting if this fails, but still seems worth being consistent. (e.g. equivalent to calling `close()`

allenporter (changes requested):
Looks great, definitely this looks good to make tests now for sure. just one small comment about failure handling.

roborock/devices/local_channel.py
```diff
@@ -113,6 +200,29 @@ async def publish(self, message: RoborockMessage) -> None:
             logging.exception("Uncaught error sending command")
             raise RoborockException(f"Failed to send message: {message}") from err
 
+    async def send_message(
```
allenporter: Can this be private?

In the future i'd like to find a way to reuse an rpc channel for this, but doesn't need to happen here.ing) but we can take a look in a follow up CL)

roborock/devices/local_channel.py
allenporter: can `_ack_nonce` and and `_local_protocol_version` be replaced with a `LocalChannelParams` to further reduce the number of mutable attributes? (I'm not sure if its best to update in `_update_encoder_decoder` or pass in as an argument.)

roborock/devices/local_channel.py
```diff
@@ -45,12 +57,79 @@ def __init__(self, host: str, local_key: str):
...
+            )
+            _LOGGER.debug(
+                "Client %s speaks the %s protocol.",
+                self._host,
+                local_protocol_version,
+            )
+            return LocalChannelParams(
+                local_key=self._local_key, connect_nonce=self._connect_nonce, ack_nonce=response.random
+            )
+        except RoborockException as e:
```
allenporter, line 103: For a follow up: Is there a very specific  specific error code or raised when the protocol is not supported? Would be nice if we could narrow down to a specific exception as right now the transport code is simply catching `except Exception as err` which is probably overly broad.

Lash-L: No response from the vac when protocol is not supported. So it just gives a timeout

roborock/devices/local_channel.py
```diff
@@ -45,12 +57,77 @@ def __init__(self, host: str, local_key: str):
         self._protocol: _LocalProtocol | None = None
         self._subscribers: CallbackList[RoborockMessage] = CallbackList(_LOGGER)
         self._is_connected = False
-
-        self._decoder: Decoder = create_local_decoder(local_key)
-        self._encoder: Encoder = create_local_encoder(local_key)
+        self._local_key = local_key
```
allenporter: I think we can drop `_local_key` and `_connect_nonce` too? Either `set _params` here or pass to `_update_encoder_decoder` explcitly

roborock/devices/local_channel.py
```diff
@@ -45,12 +57,75 @@ def __init__(self, host: str, local_key: str):
...
-        self._decoder: Decoder = create_local_decoder(local_key)
-        self._encoder: Encoder = create_local_encoder(local_key)
+        self._local_protocol_version: LocalProtocolVersion | None = None
+        self._params = LocalChannelParams(local_key=local_key, connect_nonce=get_next_int(10000, 32767), ack_nonce=None)
+        self._update_encoder_decoder(self._params)
+
+    def _update_encoder_decoder(self, params: LocalChannelParams):
+        if params is None:
+            params = self._params
+        self._params = params
```
allenporter: Can remove the `None` case now
```suggestion
        self._params = params
```

roborock/devices/local_channel.py
```diff
@@ -45,12 +57,75 @@ def __init__(self, host: str, local_key: str):
         self._protocol: _LocalProtocol | None = None
         self._subscribers: CallbackList[RoborockMessage] = CallbackList(_LOGGER)
         self._is_connected = False
-
-        self._decoder: Decoder = create_local_decoder(local_key)
-        self._encoder: Encoder = create_local_encoder(local_key)
+        self._local_protocol_version: LocalProtocolVersion | None = None
+        self._params = LocalChannelParams(local_key=local_key, connect_nonce=get_next_int(10000, 32767), ack_nonce=None)
+        self._update_encoder_decoder(self._params)
```
allenporter: Either set `_params` in caller or `_update_encoder_decoder` but not both.

allenporter (dismissed):

PR #560, opened 2025-10-24
Add diagnostic information to the device

allenporter:
This encapsulated diagnostic information entirely within the device since the device has all of the local information such as traits, device information, product information, etc. The motivation is to facilitate local testing rather than testing via Home Assistant and dealing with mocks and it could be helpful for other users of the library as well.

The diagnostic redaction is copied from [nest](https://github.com/allenporter/python-google-nest-sdm/blob/main/google_nest_sdm/diagnostics.py).

### DIFF
--- a/tests/devices/__snapshots__/test_v1_device.ambr
+++ b/tests/devices/__snapshots__/test_v1_device.ambr
@@ -32,7 +32,7 @@
       'roomId': 2362003,
       'share': False,
       'silentOtaSwitch': True,
-      'sn': 'abc123',
+      'sn': '**REDACTED**',
       'timeZoneId': 'America/Los_Angeles',
       'tuyaMigrated': False,
     }),
@@ -369,7 +369,7 @@
         'isWorkdayHolidaySupported': False,
       }),
       'network_info': dict({
-        'ip': '',
+        'ip': '**REDACTED**',
       }),
     }),
   })
@@ -407,7 +407,7 @@
       'roomId': 2362003,
       'share': False,
       'silentOtaSwitch': True,
-      'sn': 'abc123',
+      'sn': '**REDACTED**',
       'timeZoneId': 'America/Los_Angeles',
       'tuyaMigrated': False,
     }),
@@ -724,7 +724,7 @@
         'startMinute': 0,
       }),
       'network_info': dict({
-        'ip': '',
+        'ip': '**REDACTED**',
       }),
     }),
   })
@@ -762,7 +762,7 @@
       'roomId': 2362003,
       'share': False,
       'silentOtaSwitch': True,
-      'sn': 'abc123',
+      'sn': '**REDACTED**',
       'timeZoneId': 'America/Los_Angeles',
       'tuyaMigrated': False,
     }),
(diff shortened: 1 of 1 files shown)


tests/devices/__snapshots__/test_v1_device.ambr
```diff
@@ -1,13 +1,1520 @@
...
+      'network_info': dict({
+        'ip': '',
+      }),
+      'rooms': dict({
+      }),
+      'sound_volume': dict({
+      }),
+      'status': dict({
+      }),
+    }),
```
Lash-L: Are these supposed to be blank?

Lash-L: oh just kidding - isee each one has a different one filled out.

roborock/devices/device.py
```diff
@@ -113,3 +114,41 @@ async def close(self) -> None:
...
+        return {
+            "device": _redact_data(self.device_info.as_dict()),
+            "product": _redact_data(self.product.as_dict()),
+            **extra,
+        }
+
+
+T = TypeVar("T")
+
+REDACT_KEYS = {"duid", "localKey", "mac", "bssid"}
```
Lash-L: We should probably also redact sn

roborock/devices/traits/v1/network_info.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, device_uid: str, cache: Cache) -> None:
         """Initialize the trait."""
         self._device_uid = device_uid
         self._cache = cache
+        self.ip = ""
```
Lash-L, line 32: What is the variable for? The NetworkInfo object should have ip on it, right?

tests/devices/__snapshots__/test_v1_device.ambr
allenporter: How about we omit unset values? Let me know what you think of that change. (pushed)

roborock/devices/device.py
allenporter: added ip too?

roborock/devices/traits/v1/network_info.py
allenporter: This is here because network info can't be inspected if this is not set since it is not an optional field. I ended up just hacking it here because i expect `refresh()` to be called on a trait to populate this. Another option is to make it nullable?

roborock/devices/device.py
```diff
@@ -113,3 +114,41 @@ async def close(self) -> None:
...
+        return {
+            "device": _redact_data(self.device_info.as_dict()),
+            "product": _redact_data(self.product.as_dict()),
+            **extra,
+        }
+
+
+T = TypeVar("T")
+
+REDACT_KEYS = {"duid", "localKey", "mac", "bssid", "sn", "ip"}
```
Lash-L, line 132: i'm not sure if we need to redact ip as it is the local ip and that shows up in the logs. but probably not worth the change for now, just going to approve but maybe if we touch this file again, it could be worth removing?

PR #562, opened 2025-10-25
chore: get uv release to work properly

Lash-L:
Theoretically fix python-semantic release...

Was a bad time for a breaking change in python-semantic release + migrating to uv lol. Lots of small things

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,14 @@ build-backend = "hatchling.build"
 
 [tool.semantic_release]
 branch = "main"
-branches = ["main", "temp-main-branch"]
+
+[tool.semantic_release.branches.main]
+match = "main"
+prerelease = false
+
+[tool.semantic_release.branches.temp-main-branch]
+match = "temp-main-branch"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -49,10 +49,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: "3.11"
+          activate-environment: true
+      - name: Build package
+        run: uv build
```
Lash-L, line 64: this is here mainly as a test, if we ever do anything to break building, this should catch it.

.github/workflows/ci.yml
```diff
@@ -49,10 +49,60 @@ jobs:
...
+
+  # Test semantic-release configuration on PR branches
+  test-release:
```
Lash-L, line 67: Another test that should hopefully help us find things before they break

PR #572, opened 2025-10-26
feat: Add map content to the Home trait

allenporter:
This manages caching map content for the whole home, including performing discovery similar to metadata. This caches the raw map bytes API response and parses it on startup for each map.

### DIFF
--- a/roborock/devices/traits/v1/home.py
+++ b/roborock/devices/traits/v1/home.py
@@ -86,7 +86,7 @@ class HomeTrait(RoborockBase, common.V1TraitMixin):
                 self._home_map_content = {
                     k: self._map_content.parse_map_content(v) for k, v in cache_data.home_map_content.items()
                 }
-            except RoborockException as ex:
+            except (ValueError, RoborockException) as ex:
                 _LOGGER.warning("Failed to parse cached home map content, will re-discover: %s", ex)
                 self._home_map_content = {}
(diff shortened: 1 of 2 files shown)


roborock/devices/traits/v1/home.py
```diff
@@ -141,13 +166,11 @@ async def refresh(self) -> Self:
...
+        new_map_content = await self._refresh_map_content()
         # Refresh the current map's room data
-        current_map_data = self._home_map_info.get(map_flag)
-        if current_map_data:
-            map_data = await self._refresh_map_info(current_map_info)
-            if map_data != current_map_data:
-                await self._update_home_map_info({**self._home_map_info, map_flag: map_data})
-
+        combined_map_info = await self._refresh_map_info(current_map_info)
+        await self._update_current_map_cache(map_flag, combined_map_info, new_map_content)
```
Lash-L, line 173: So I guess in our HA implementation, this is going to be checked on the HA side if we are in cleaning/ how frequently we update - that is not a responsibility here, right?

allenporter: Yes, the caller is responsible for frequency of calling `refresh`. One difference is we're combing map info and map content and room information, rather than treating just the map bytes/image as a separate call. If that turns out to be a problem for some reason, we could separate them.

PR #578, opened 2025-11-07
Update supported devices with Roborock Saros 10R

thromera:
Added Roborock Saros 10R to the supported devices list.

I have one and can issue basic commands with app_send_command at least.


allenporter:
Thank you for the contribution. Closing since this documentation has been removed.

PR #580, opened 2025-11-08
fix: Improve Home trait discovery process.

allenporter:
This updates the Home trait to allow the current map to be refreshed even if initial discovery fails. The caller can invoke discovery at startup, but then after just invoke refresh and the right thing should happen.

We now track discovery explicitly and only persist the cache when we have completed discovery to avoid persisting an incomplete view of all available maps.

### DIFF
--- a/tests/devices/traits/v1/test_home.py
+++ b/tests/devices/traits/v1/test_home.py
@@ -500,9 +500,6 @@ async def test_discover_home_device_busy_cleaning(
 
     # Verify we now have all of the information populated from discovery
     assert home_trait.home_map_info is not None
-    assert len(home_trait.home_map_info) == 2
-
-    assert home_trait.home_map_info is not None
     assert home_trait.home_map_info.keys() == {0, 123}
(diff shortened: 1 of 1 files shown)


roborock/devices/traits/v1/home.py
```diff
@@ -156,8 +165,15 @@ async def refresh(self) -> Self:
...
         """
-        if self._home_map_info is None:
-            raise RoborockException("Cannot refresh home data without home cache, did you call discover_home()?")
+        if not self._discovery_completed:
+            # Running initial discovery also populates all of the same information
+            # as below so we can just call that method. If the device is busy
+            # then we'll fall through below to refresh the current map only.
+            try:
+                await self.discover_home()
+                return self
```
Lash-L, line 174: not against it but do we utilize anywhere the fact that this is returning self?

allenporter: Yes, agreed, we never use it and it only adds typing pain. I'll remove it in a followup PR.

PR #586, opened 2025-11-11
Update trait `refresh` method to return `None`

allenporter:
The return value is currently not used and just causes typing issues.  This is followup from discussion in https://github.com/Python-roborock/python-roborock/pull/580

### DIFF
--- a/roborock/devices/traits/v1/clean_summary.py
+++ b/roborock/devices/traits/v1/clean_summary.py
@@ -77,7 +77,6 @@ class CleanSummaryTrait(CleanSummaryWithDetail, common.V1TraitMixin):
                 except Exception:
                     # Return final record when an exception occurred
                     return final_record
-
             # There are still a few unknown variables in this.
             begin, end, duration, area = unpack_list(response, 4)
             return CleanRecord(begin=begin, end=end, duration=duration, area=area)


roborock/devices/traits/v1/clean_summary.py
```diff
@@ -64,21 +63,15 @@ def _parse_clean_record_response(cls, response: common.V1ResponseData) -> CleanR
...
                     final_record.start_type = records[0].start_type
                     for rec in records[0:-1]:
                         final_record.duration = (final_record.duration or 0) + (rec.duration or 0)
                         final_record.area = (final_record.area or 0) + (rec.area or 0)
                         final_record.avoid_count = (final_record.avoid_count or 0) + (rec.avoid_count or 0)
                         final_record.wash_count = (final_record.wash_count or 0) + (rec.wash_count or 0)
-                        final_record.square_meter_area = (final_record.square_meter_area or 0) + (
-                            rec.square_meter_area or 0
-                        )
-                    return final_record
```
Lash-L, line 77: I think this is a unrelated/accidental change?

allenporter: Yes, thank you, fixed.

PR #587, opened 2025-11-11
Add device info for roborock.vacuum.s5e

allenporter:
Update device info for my vacuums. Generated with these commands:
```
$ roborock get-device-info
$ roborock update-docs --data-file=./device_info.yaml --output-file=./SUPPORTED_FEATURES.md
```


Lash-L (approved):
Nice - glad it worked. Realize at some point I need to store the firmware version. If a user runs it and they have a device with a *newer* firmware version, we should overwrite it. But that's a future problem.

And storing protocol version (local and cloud) would be nice

PR #588, opened 2025-11-11
feat: Connect to devices asynchronously

allenporter:
Update the device manager to not block on connecting to devices immediately, but instead connect to them asynchronously in the background. Once they connect, the devices will reconnect on their own (e.g. using the mqtt session or local channel behavior).

Add a test that exercises the behavior of an unsuccessful connect with retries in the background.

### DIFF
--- a/roborock/devices/device.py
+++ b/roborock/devices/device.py
@@ -127,7 +127,9 @@ class RoborockDevice(ABC, TraitsMixin):
                         return
                     except RoborockException as e:
                         _LOGGER.info("Failed to connect to device %s: %s", self.name, e)
-                        _LOGGER.info("Retrying connection to device %s in %s seconds", self.name, backoff.total_seconds())
+                        _LOGGER.info(
+                            "Retrying connection to device %s in %s seconds", self.name, backoff.total_seconds()
+                        )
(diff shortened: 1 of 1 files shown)


roborock/devices/device.py
```diff
@@ -107,6 +148,12 @@ async def connect(self) -> None:
 
     async def close(self) -> None:
         """Close all connections to the device."""
+        if self._connect_task:
+            self._connect_task.cancel()
+            try:
+                await self._connect_task
```
Lash-L, line 154: Maybe I'm sleep deprived but can you double check this is the right functionality? Won't this reopen the connection on close()

allenporter: The `self._connect_task` is a task started in `start_connect`.  That start event happens once there, and gets kicked off in the background. Then in close, we cancel the task. This line will wait for the task to complete -- either because it already completed or because its waiting for cancellation to finish.

This has an example https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel and what we're doing here is similar to how the `main` example looks.

Lash-L: Ah yes that makes sense. Looking at it again. Thank you for the explanation.

Lash-L (approved):

PR #591, opened 2025-11-14
fix: switches commands were incorrect

Lash-L:
The structure for the timers was incorrect, this is now working in my tests


commitlint.config.mjs
```diff
@@ -1,6 +1,9 @@
 export default {
     extends: ["@commitlint/config-conventional"],
-    ignores: [(msg) => /Signed-off-by: dependabot\[bot]/m.test(msg)],
+    ignores: [
```
Lash-L, line 3: unrelated change, but i think worthwhile

PR #593, opened 2025-11-15
fix: fix L01 encoding and decoding

Lash-L:
Fix L01 on the new devices api

### DIFF
--- a/roborock/devices/local_channel.py
+++ b/roborock/devices/local_channel.py
@@ -10,8 +10,6 @@ from roborock.exceptions import RoborockConnectionException, RoborockException
 from roborock.protocol import create_local_decoder, create_local_encoder
 from roborock.roborock_message import RoborockMessage, RoborockMessageProtocol
 
-
-
 from ..protocols.v1_protocol import LocalProtocolVersion
 from ..util import get_next_int
 from .channel import Channel
@@ -53,7 +51,6 @@ class LocalChannel(Channel):
     format most parsing to higher-level components.
     """
 
-
     def __init__(self, host: str, local_key: str):
         self._host = host
         self._transport: asyncio.Transport | None = None
@@ -67,7 +64,7 @@ class LocalChannel(Channel):
 
     def _update_encoder_decoder(self, params: LocalChannelParams) -> None:
         """Update the encoder and decoder with new parameters.
-        
+
         This is invoked once with an initial set of values used for protocol
         negotiation. Once negotiation completes, it is updated again to set the
         correct nonces for the follow up communications and updates the encoder


roborock/devices/local_channel.py
```diff
@@ -51,13 +51,15 @@ class LocalChannel(Channel):
     format most parsing to higher-level components.
     """
 
+    _protocol_cache: dict[str, LocalProtocolVersion] = {}
```
Lash-L: I can move the responsibility for this further upstream if you'd like.

roborock/devices/local_channel.py
```diff
@@ -72,6 +74,8 @@ def _update_encoder_decoder(self, params: LocalChannelParams):
         )
         # Callback to decode messages and dispatch to subscribers
         self._data_received: Callable[[bytes], None] = decoder_callback(self._decoder, self._subscribers, _LOGGER)
+        if self._protocol:
+            self._protocol.messages_cb = self._data_received
```
allenporter: Instead of mutating the object, lets just give it a different callback to this class that will invoke the right thing. (change the constructor value of _LocalProtocol(...) to point to something else.

roborock/devices/local_channel.py
allenporter: i generally get the idea of caching things, but i'm having a hard time seeing how often this will get used. Are we creating local channels repeatedly? i would have thought once per session given once connected it will stay connected.

allenporter: (I would cache this like or in `NetworkInfo`, but i realize it could change if firmware version changes so i casee thats difficult)

Lash-L: Right now, seemingly we are. On every reconnect. Rob was noticing that it took longer for his system to reconnect on disconnects

allenporter: Are we reconnecting so often that this matters in practice? Ideally we reconnect at most once per day so presumably reconnect is not the primary cost. I can see wanting to optimize startup time, though, which i don't think this does. I guess I think if we really want to cache this to improve performance then we should also cache the same way we cache network info to avoid the extra round trip on startup too?

roborock/devices/local_channel.py
allenporter: Updated.

roborock/devices/local_channel.py
allenporter: I think we should come back to the protocol cache in a separate PR, so its reverted for now.

allenporter:
Offline Lash-L also reviewed and gave an LG -- just for audit log :) (not sure how approval works on two party CL)

PR #605, opened 2025-11-29
fix: add a health manager for restarting unhealthy mqtt connections

allenporter:
This adds the ability to restart an mqtt session and adds a callback based health manager that handles restarting the session when the number of repeated timeouts for the mqtt channel reaches a threshold.

### DIFF
--- a/roborock/mqtt/roborock_session.py
+++ b/roborock/mqtt/roborock_session.py
@@ -86,12 +86,13 @@ class RoborockMqttSession(MqttSession):
     async def close(self) -> None:
         """Cancels the MQTT loop and shutdown the client library."""
         self._stop = True
-        if self._reconnect_task:
-            self._reconnect_task.cancel()
-            try:
-                await self._reconnect_task
-            except asyncio.CancelledError:
-                pass
+        tasks = [task for task in [self._connection_task, self._reconnect_task] if task]
+        for task in tasks:
+            task.cancel()
+        try:
+            await asyncio.gather(*tasks)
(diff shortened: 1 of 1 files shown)


roborock/mqtt/health_manager.py
```diff
@@ -0,0 +1,51 @@
...
+We observe a problem where sometimes the MQTT connection appears to be alive but
+no messages are being received. To mitigate this, we track consecutive timeouts
+and restart the connection if too many timeouts occur in succession.
+"""
+
+import datetime
+from collections.abc import Awaitable, Callable
+
+# Number of consecutive timeouts before considering the connection unhealthy.
+TIMEOUT_THRESHOLD = 3
```
Lash-L, line 12: This would technically trigger immediately with gathered function calls. But anything more than this is probably too much complexity.

Lash-L: potential follow up could be to keep track of the last timeout. if the timeout is more than say 15 seconds ago, it increases the increment. Could be a follow up PR though, I don't want to slow this one down as we are on a time crunch.

Lash-L (approved):

roborock/mqtt/health_manager.py
allenporter: I think we're currently sending most commands serially, but yeah this kind of heuristic is hard. i was also considering if we could do it entirely in the mqtt session but hard when you can't correlate incoming and outgoing messages to know if something really did timeout.

Not sure if i get the timeout point you're making but interested in following up.

PR #606, opened 2025-11-29
fix: fallback to the cached network information on failure

allenporter:
Use the cached network info when looking up the latest network information fails.


Lash-L (dismissed):
good outside copilot comments. can re-approve

Lash-L (approved):

PR #607, opened 2025-11-29
chore: add design documentation and copilot instructions

allenporter:
I would like to do some further simplification of the device channels since there are currently too many unnecessary layers of abstraction, but wanted to first document what currently exists.

These are AI generated with co-pilot and Gemini 3 Pro.

### DIFF
--- a/roborock/devices/README.md
+++ b/roborock/devices/README.md
@@ -13,36 +13,36 @@ graph TB
     subgraph "Application Layer"
         User[Application Code]
     end
-    
+
     subgraph "Device Management Layer"
         DM[DeviceManager]
         Traits[Device Traits]
     end
-    
+
     subgraph "Channel Layer"
         V1C[V1Channel]
         RPC[RpcChannel]
         MC[MqttChannel]
         LC[LocalChannel]
     end
-    
+
     subgraph "Session Layer"
         MS[MqttSession]
         LS[LocalSession Factory]
     end
-    
+
     subgraph "Protocol Layer"
         Proto[Protocol Encoders/Decoders]
         V1P[V1 Protocol]
         A01P[A01 Protocol]
         B01P[B01 Protocol]
     end
-    
+
     subgraph "Transport Layer"
         MQTT[MQTT Broker]
         TCP[TCP Socket]
     end
-    
+
     User --> DM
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,117 @@
+# GitHub Copilot Instructions for python-roborock
```
Lash-L: I wish a standard existed for this for all AI tools...

.github/copilot-instructions.md
```diff
@@ -0,0 +1,117 @@
...
+*   **Encryption**: `pycryptodome`
+*   **Testing**: `pytest`, `pytest-asyncio`
+*   **Linting/Formatting**: `ruff`
+
+## Coding Standards
+
+### 1. Typing
+*   **Strict Typing**: All functions and methods must have type hints.
+*   **Generics**: Use `list[str]` instead of `List[str]`, `dict[str, Any]` instead of `Dict[str, Any]`, etc.
+*   **Optional**: Use `str | None` instead of `Optional[str]`.
```
Lash-L: for what it is worth, i prefer this to be the case

.github/copilot-instructions.md
allenporter: The standard I am aware of is to have them all reference each others files.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,117 @@
...
+
+### 2. Asynchronous Programming
+*   **Async/Await**: Use `async def` and `await` for all I/O bound operations.
+*   **Context Managers**: Use `async with` for managing resources like network sessions and locks.
+*   **Concurrency**: Use `asyncio.gather` for concurrent operations. Avoid blocking calls in async functions.
+
+### 3. Documentation
+*   **Docstrings**: Use Google-style docstrings for all modules, classes, and functions.
+*   **Comments**: Comment complex logic, especially protocol parsing and encryption details.
```
Lash-L: Worth stating that we use pydoc?

.github/copilot-instructions.md
Lash-L: I think stating that we use semantic commit tags would be helpful as well.

PR #611, opened 2025-11-29
feat: make status dynamic

Lash-L:
This makes status dynamically determined based on the features on the device instead of relying on user provided examples. A PR will follow this one to delete the old status

### DIFF
--- a/roborock/devices/traits/v1/status.py
+++ b/roborock/devices/traits/v1/status.py
@@ -21,6 +21,16 @@ class StatusTrait(StatusV2, common.V1TraitMixin):
         the latest data from the device. You must pass in the device feature trait
         to this trait so that the dynamic attributes can be pre-determined.
 
+    The current dynamic attributes are:
+    - Fan Speed
+    - Water Mode
+    - Mop Route
+
+    You should call the _options() version of the attribute to know which are supported for your device
+    (i.e. fan_speed_options())
+    Then you can call the _mapping to convert an int value to the actual Enum. (i.e. fan_speed_mapping())
+    You can call the _name property to get the str value of the enum. (i.e. fan_speed_name)
+
     """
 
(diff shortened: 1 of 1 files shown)


roborock/data/code_mappings.py
```diff
@@ -104,6 +104,13 @@ def keys(cls) -> list[str]:
         """Returns a list of all member values."""
         return [member.value for member in cls]
 
+    def __eq__(self, other: Any) -> bool:
```
Lash-L, line 108: this is a very vague eq, but I think it makes sense, let me know if you think otherwise though @allenporter

allenporter: What did you think about the copilot comment on hash?

Lash-L: The comment makes sense - I don't necessarily see us often mixing the types in the same section where they would be hashed together. My opinion is that the __eq__ logic is helpful enough that it makes the warning on __hash__ worthwhile. Open to other opinions though.

I added a docstring to hash for now.

roborock/devices/traits/v1/status.py
```diff
@@ -1,24 +1,82 @@
...
+
+class StatusTrait(StatusV2, common.V1TraitMixin):
+    """Trait for managing the status of Roborock devices.
+
+    The StatusTrait gives you the access to the state of a Roborock vacuum.
+    The various attribute options on state change per each device.
+    Values like fan speed, mop mode, etc. have different options for every device
+    and are dynamically determined.
+
+    Usage:
```
allenporter, line 19: is there anything callers need to do about knowing what features are supported or not? Maybe i'm mixing up what this will do.

Lash-L: How are those extra docs?

allenporter (approved):

PR #613, opened 2025-12-02
feat: Make CacheData serializable

allenporter:
Update the CacheData class with serialization methods so that clients can use it for persistence. This provides an API for Home Assistant to use for serialization so it doesn't need to make its own.

Fixing https://github.com/home-assistant/core/issues/157663

### DIFF
--- a/roborock/devices/file_cache.py
+++ b/roborock/devices/file_cache.py
@@ -6,7 +6,6 @@ persistent caching of device data across application restarts.
 """
 
 import asyncio
-import logging
 import pathlib
 import pickle
 from collections.abc import Callable


roborock/devices/file_cache.py
```diff
@@ -17,18 +18,28 @@
...
+        self._serialize_fn = serialize_fn
+        self._deserialize_fn = deserialize_fn
 
     async def get(self) -> CacheData:
         """Get cached value."""
         if self._cache_data is not None:
             return self._cache_data
-
-        data = await load_value(self._file_path)
+        logging.info("What's the deserialize_fn? %s", self._deserialize_fn)
```
Lash-L: stale log?

roborock/devices/file_cache.py
```diff
@@ -17,18 +18,28 @@
...
+        init_fn: Callable[[], CacheData] = CacheData,
+        serialize_fn: Callable[[Any], bytes] = pickle.dumps,
+        deserialize_fn: Callable[[bytes], Any] = pickle.loads,
+    ) -> None:
         """Initialize the file cache with the given file path."""
         self._init_fn = init_fn
         self._file_path = file_path
         self._cache_data: CacheData | None = None
+        logging.info("What's the serialize_fn? %s", serialize_fn)
+        logging.info("What's the deserialize_fn? %s", deserialize_fn)
```
Lash-L: same thing here?

roborock/devices/file_cache.py
```diff
@@ -17,18 +18,28 @@
 class FileCache(Cache):
     """File backed cache implementation."""
 
-    def __init__(self, file_path: pathlib.Path, init_fn: Callable[[], CacheData] = CacheData) -> None:
+    def __init__(
+        self,
+        file_path: pathlib.Path,
+        init_fn: Callable[[], CacheData] = CacheData,
+        serialize_fn: Callable[[Any], bytes] = pickle.dumps,
+        deserialize_fn: Callable[[bytes], Any] = pickle.loads,
```
Lash-L, line 25: does it make sense to set the dict serialization on this side rather than on the HA side? as a separate premade serialization function?

allenporter: Home Assistant is really not using FileCache at all, and has its own json store. I mostly used this for testing of the approach.

So the real value in this CL is  making this library handle the "to_dict" / "from_dict" and these are just tests making sure that can be used for various serialization strategies.

allenporter: We could make file cache use json by default instead of pickle? that is not a bad idea -- we could use it in the CLI or the example that way too. But just saying Home Assistant won't use it as is.

roborock/devices/file_cache.py
allenporter: Oops, removed.

roborock/devices/file_cache.py
allenporter: Removed.

roborock/devices/file_cache.py
Lash-L: i'm just thinking that if we have a custom serialization function and deserialization function on the HA side that may get annoying. i.e. if we update something on CacheData, we then have to do a version bump on the HA side + change the serialization functions, rather than having the serialization happen on the library side and HA just utilizing them for it's cache

allenporter: The function ha will call is `CacheData.as_dict` and `.from_dict`. it just does it's own json loads etc.

allenporter: I don't think we can get a large change to this API done by release time.

allenporter: I don't think we can get a large change to this API done by release time.

allenporter: I don't think we can get a large change to this API done by release time.

PR #614, opened 2025-12-02
fix: Fix issues with the cache clobbering information for each device

allenporter:
Update cache interface to add a device specific cache.

This fixes invalid use today where the device data is not stored based on device id in all cases.

Related to https://github.com/Python-roborock/python-roborock/pull/615

### DIFF
--- a/roborock/devices/cache.py
+++ b/roborock/devices/cache.py
@@ -9,7 +9,6 @@ from dataclasses import dataclass, field
 from typing import Any, Protocol
 
 from roborock.data import CombinedMapInfo, HomeData, NetworkInfo, RoborockBase
-from roborock.data.containers import RoborockBase
 from roborock.device_features import DeviceFeatures
 
 


roborock/devices/traits/v1/home.py
```diff
@@ -247,21 +241,17 @@ async def _update_current_map(
...
+            device_cache_data.home_map_info[map_flag] = map_info
             if map_content.raw_api_response:
-                if cache_data.home_map_content_base64 is None:
-                    cache_data.home_map_content_base64 = {}
-                cache_data.home_map_content_base64[map_flag] = base64.b64encode(map_content.raw_api_response).decode(
-                    "utf-8"
-                )
-            await self._cache.set(cache_data)
+                if device_cache_data.home_map_content_base64 is None:
+                    device_cache_data.home_map_content_base64 = {}
```
Lash-L, line 250: may be worth setting home_map_content_base64 to a default factory so we never have to worry about it being none.

roborock/devices/cache.py
```diff
@@ -9,36 +9,75 @@
...
+    """Home map information indexed by map_flag.
+
+    This is deprecated. Use the per-device `home_map_info` field instead.
+    """
 
     home_map_content: dict[int, bytes] = field(default_factory=dict)
     """Home cache content for each map data indexed by map_flag.
 
-    This is deprecated in favor of `home_map_content_base64`.
+    This is deprecated. Use the per-device `home_map_content_base64` field instead.
```
Lash-L, line 60: do we need to hold onto the deprecated value anymore?

Lash-L (dismissed):
few small comments - I do not think this PR should be held up for these - maybe in a follow up later.

As well, what will happen right now for users who have a CacheData that is stored but now are have completely different typing? I don't think we have a versioning system or anything in place but maybe i have missed it

But that's more on the HA side, but just something I want to make sure we have covered

roborock/devices/cache.py
allenporter: I think i want to just let it die and we can drop the old field name in the next major release as a breaking change.

roborock/devices/traits/v1/home.py
allenporter: I realize we set the default value to dict and this can never be none. I'll drop this in a followup PR.

Lash-L:
> My plan was to bump the store version and blow away the old cache entirely.

Perfect!

Lash-L (approved):

PR #624, opened 2025-12-04
fix: Update logging to be more conservative

allenporter:
We currently log warning logs on local disconnects. This is moved to log info and instead we should let the caller log warnings when RPCs fail etc. This will log at the info level only after failures to reconnect, but then logs again when reconnect happens.

### DIFF
--- a/roborock/devices/v1_channel.py
+++ b/roborock/devices/v1_channel.py
@@ -371,8 +371,15 @@ class V1Channel(Channel):
 
                 # Not connected, so wait with backoff before trying to connect.
                 # The first time through, we don't sleep, we just try to connect.
+                # We also only log after the first retry to avoid spamming logs.
                 local_connect_failures += 1
                 if local_connect_failures > 1:
+                    if local_connect_failures == 2:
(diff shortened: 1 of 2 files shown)


roborock/devices/v1_channel.py
```diff
@@ -371,15 +371,24 @@ async def _background_reconnect(self) -> None:
 
                 # Not connected, so wait with backoff before trying to connect.
                 # The first time through, we don't sleep, we just try to connect.
+                # We also only log after the first retry to avoid spamming logs.
                 local_connect_failures += 1
                 if local_connect_failures > 1:
+                    if local_connect_failures == 2:
+                        _LOGGER.info(
```
Lash-L, line 378: I worry someone may get confused by their logs if it stops showing it retrying. What if we did this every n reconnect failures?

Lash-L: Or maybe just "will continue retrying" rather than giving a timeframe

allenporter:
Will let this get replaced by #637

PR #627, opened 2025-12-04
feat: Add pinging to local client

Lash-L:
Tested twice:

Without pinging:
9 disconnects in 5 minutes

With pinging:
0 disconnects in 5 minutes

### DIFF
--- a/tests/devices/test_local_channel.py
+++ b/tests/devices/test_local_channel.py
@@ -366,3 +366,112 @@ async def test_hello_preferred_protocol_version_ordering(mock_loop: Mock, mock_t
     assert channel._params is not None
     assert channel._params.ack_nonce == 11111
     assert channel._is_connected is True
+
+
+async def test_keep_alive_task_created_on_connect(local_channel: LocalChannel, mock_loop: Mock) -> None:
+    """Test that _keep_alive_task is created when connect() is called."""
+    # Before connecting, task should be None
+    assert local_channel._keep_alive_task is None
+
+    await local_channel.connect()
+
+    # After connecting, task should be created and not done
+    assert local_channel._keep_alive_task is not None
+    assert isinstance(local_channel._keep_alive_task, asyncio.Task)
+    assert not local_channel._keep_alive_task.done()
+
+
+async def test_keep_alive_task_canceled_on_close(local_channel: LocalChannel, mock_loop: Mock) -> None:
+    """Test that the keep-alive task is properly canceled when close() is called."""
+    await local_channel.connect()
+
+    # Verify task exists
+    task = local_channel._keep_alive_task
+    assert task is not None
+    assert not task.done()
+
+    # Close the connection
+    local_channel.close()
+
+    # Give the task a moment to be cancelled
+    await asyncio.sleep(0.01)
+
+    # Task should be canceled and reset to None
+    assert task.cancelled() or task.done()
+    assert local_channel._keep_alive_task is None
+
+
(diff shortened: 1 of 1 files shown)


roborock/devices/local_channel.py
```diff
@@ -132,6 +134,28 @@ async def _hello(self):
...
+        while self._is_connected:
+            try:
+                await asyncio.sleep(_PING_INTERVAL)
+                if self._is_connected:
+                    await self._ping()
+            except asyncio.CancelledError:
+                break
+            except Exception:
+                _LOGGER.debug("Keep-alive ping failed", exc_info=True)
+                # Retry next interval
```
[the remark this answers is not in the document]
Lash-L: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Python-roborock/python-roborock/pull/627#discussion_r2589458996)

roborock/devices/local_channel.py
```diff
@@ -166,6 +190,7 @@ async def connect(self) -> None:
         # Perform protocol negotiation
         try:
             await self._hello()
+            self._keep_alive_task = asyncio.create_task(self._keep_alive_loop())
```
allenporter, line 193: Move this outside the scope of the try/catch. It won't throw RoborockException

roborock/devices/local_channel.py
```diff
@@ -132,6 +134,28 @@ async def _hello(self):
...
+
+    async def _keep_alive_loop(self) -> None:
+        while self._is_connected:
+            try:
+                await asyncio.sleep(_PING_INTERVAL)
+                if self._is_connected:
+                    await self._ping()
+            except asyncio.CancelledError:
+                break
+            except Exception:
```
allenporter, line 155: Let's catch RoborockException (expected) separate from uncaught exceptions so we can log them separately e.g. "ping failed" vs "Uncaught exception" implying a bug/shouldn't happen case we need to fix.  Similar to `_background_reconnect` in v1 channel?

roborock/devices/local_channel.py
```diff
@@ -17,6 +17,7 @@
 _LOGGER = logging.getLogger(__name__)
 _PORT = 58867
 _TIMEOUT = 5.0
+_PING_INTERVAL = 10
```
allenporter, line 20: Can you use a timedelta here to be more explicit on the unit?

```suggestion
_PING_INTERVAL = datetime.timedelta(seconds=10)
```

roborock/devices/local_channel.py
```diff
@@ -132,6 +134,28 @@ async def _hello(self):
...
+        await self._send_message(
+            roborock_message=ping_message,
+            request_id=ping_message.seq,
+            response_protocol=RoborockMessageProtocol.PING_RESPONSE,
+        )
+
+    async def _keep_alive_loop(self) -> None:
+        while self._is_connected:
+            try:
+                await asyncio.sleep(_PING_INTERVAL)
```
allenporter, line 150: Assuming you accept the above suggestion
```suggestion
                await asyncio.sleep(_PING_INTERVAL.total_seconds())
```

tests/devices/test_local_channel.py
```diff
@@ -366,3 +366,112 @@ async def mock_do_hello(local_protocol_version: LocalProtocolVersion) -> LocalCh
     assert channel._params is not None
     assert channel._params.ack_nonce == 11111
     assert channel._is_connected is True
+
+
+async def test_keep_alive_task_created_on_connect(local_channel: LocalChannel, mock_loop: Mock) -> None:
```
allenporter, line 371: We should avoid these tests of internal state like this generally, i'm not sure its worhtwhile to keep?

PR #631, opened 2025-12-05
chore: Add tests that reproduce key parsing bugs

allenporter:
Add tests for  #630. I first wrote the tests and made sure they failed without that PR, then confirm they pass with those changes.

The testing approach is:
(1) Have a test with a generic data type reproducing the issue
(2) Test the caching library with the specific objects causing the problem


Lash-L (dismissed):

Lash-L:
Oh gotta fix long before this one is merged

PR #634, opened 2025-12-05
fix: Throw MQTT authentication errors as authentication related exceptions

allenporter:
Update the MQTT exception handling to carve out a different exception type for authorization errors.  Today callers do not have the ability to handle special authentication errors like this:
```
2025-12-05 13:42:39.684 INFO (MainThread) [roborock.mqtt.roborock_session] MQTT error starting session: [code:135] Not authorized
```
Now this will be thrown with `RoborockInvalidCredentials`.  Callers can catch this  to know when they need to re-authenticate a device and obtain new mqtt params.

### DIFF
--- a/roborock/mqtt/roborock_session.py
+++ b/roborock/mqtt/roborock_session.py
@@ -18,9 +18,8 @@ import aiomqtt
 from aiomqtt import MqttCodeError, MqttError, TLSParameters
 
 from roborock.callbacks import CallbackMap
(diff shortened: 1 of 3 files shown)


Lash-L (approved):
this is good! Only note: Unathorized does not only mean they need new credentials, it could be that the specific userdata is rate limited for <=24 hours. 

Getting new credentials will fix the problem, waiting will also fix the problem (if it is a rate limit). Just important we make that clear.

Issue #639, opened 2025-12-06
Widespread issues in python-roborock in 3.10.x


allenporter:
We're seeing widespread issues introduced resulting in timeouts, rate limiting. We'll use this issue to track the diagnosis and improvements.

We can also use this to discuss local testing with new library versions inside of Home Assistant for users who are comfortable with making local python package updates.

Lash-L:
We should also look through #632 and double check all the logic there.

allenporter:
In summary we believe the two issues are and solutions are:
- Local connections were skipped entirely, causing large load on MQTT servers. This is the opposite intended consequence of the overhaul with a bug intended to reduce duplicate code that introduced it without sufficient tests. We now have tests in place for this behavior.
- Local connections were being skipped on the first RPC. The check for "can i use the local connection?" would skip since it ran before we actually attempted to create the local connection. This was introduced by a chance that tried to speed up startup time by creating devices without blocking on establishing a connection. Now we block on establishing a connection for X seconds, but don't fail if it fails or times out. This will slow down startup time, but avoids a cloud dependency for the first RPC.

allenporter:
My original thinking is (1) local connection should break, then get re-established by the background thread. mqtt connection should be watched by health manager (2) mqtt connection should be alive thus we can say the device is "available" still. Howevr, commands during this period will either fail or run over MQTT.

However, one caveat to that is given the topic keepalive we may not consider the connection healthy. I may want to change the MQTT channel health to be based on the mqtt session instead of the channel (which needs a topic)

However, if mqtt connection is not alive during this time period then it will still be considered not available. in that case, availability should be determined by RPCs failing.

Lash-L:
Slightly related #643

Will not help with reliability but should help with all entities being available.

Lash-L:
![image](https://github.com/user-attachments/assets/a385af6a-65ef-4dc5-bfcc-622a5a6177e0)

Tested with the latest version of the library overnight. Behavior looks expected

Will check logs in a bit

PR #641, opened 2025-12-07
fix: Revert CLIENT_KEEPALIVE back to 60

allenporter:
This reverts back to the old default since there is no specific reason to change it. This reverts part of https://github.com/Python-roborock/python-roborock/pull/632

Issue #639

### DIFF
--- a/roborock/mqtt/roborock_session.py
+++ b/roborock/mqtt/roborock_session.py
@@ -24,7 +24,7 @@ from .session import MqttParams, MqttSession, MqttSessionException
 _LOGGER = logging.getLogger(__name__)
 _MQTT_LOGGER = logging.getLogger(f"{__name__}.aiomqtt")
 
-CLIENT_KEEPALIVE = datetime.timedelta(seconds=120)
+CLIENT_KEEPALIVE = datetime.timedelta(seconds=60)
 TOPIC_KEEPALIVE = datetime.timedelta(seconds=60)
 
 # Exponential backoff parameters


roborock/mqtt/roborock_session.py
```diff
@@ -24,7 +24,7 @@
 _LOGGER = logging.getLogger(__name__)
 _MQTT_LOGGER = logging.getLogger(f"{__name__}.aiomqtt")
 
-CLIENT_KEEPALIVE = datetime.timedelta(seconds=120)
+CLIENT_KEEPALIVE = datetime.timedelta(seconds=60)
 TOPIC_KEEPALIVE = datetime.timedelta(seconds=60)
```
Lash-L, line 28: Good with the change as is. But do we think the topic keepalive logic is okay? How often are we sending to a mqtt  topic? Although, I guess otherthings should come in naturally that we didn't send, so maybe it's fine? I'll try to test

Lash-L (approved):

roborock/mqtt/roborock_session.py
allenporter: I think what we want is to establish the topic connection and not have it constantly refresh, but mostly stay around. Before it was doing it on every RPC, which is obviously bad sow we want to find the right interval.

I assume what we want is something like this:
- have some common polling interval
- set the keepalive to be 1.5-2x higher than the polling interval.

So maybe we set this based on common polling interval for map data (alt: make configurable, pass in via callers so home assistant can set it)

allenporter: So i think the rational is:
- we don't normally poll cloud at all
- for map content, which relies on cloud, we use the V1_CLOUD_IN_CLEANING_INTERVAL only during cleaning. That  is 30 seconds, then this keeping topics around for 60 seconds is 2x that and within that rule.

Lash-L: Yes in an ideal setup, we should just be using cloud for map. 30 seconds seems correct for the interval at which we update. 

What classifies as a success for rhe topic keepalive? A message we requested or any message?

allenporter: Any time a message is requested it will reset the timer.

So, a caveat is: We do /not/ have a notion of topic health. That said, i don't think opening and closing a topic subscription helps with health and the main connection issues we see are at the session level. So, our MQTT health manager should manage looking for lots of failed RPCs when things are wedged and should reconnect the client.

Just to reset, i think we have two things we care about:
(1) Priority 0: Do not get "stuck" at 3am
(2) Priority 1: Do not cause lots of noise / unavailability at 3am and paper over any reconnects etc. (If someoen tries to start cleaning, too bad, but at least it's not causing disconnect noise if its idle).

I don't think we're done with #2 yet and may need client (home assistant) solutions to ignore refresh errors. e.g. (a) fail 3 times before marking unavailable always (b) ignore errors for ~n times between 3am and 3:10am (c) do not even try to send RPCs at 3am to 3:10am)

PR #643, opened 2025-12-07
fix: handle base64 serializing wrong

Lash-L:
Basically, the serialization and deserialization logic was changing base64 to base_64 because of camelization and decamelization. I could modify the regex, but I was a bit worried that something might be relying on it that I'm not thinking of. I figured it would be best to make a targeted fix and revisit this once everything is stable again.


roborock/data/containers.py
```diff
@@ -32,7 +32,10 @@ def _decamelize(s: str):
     s = re.sub(r"(?<=[a-z0-9])([A-Z])", r"_\1", s)
     s = re.sub(r"([A-Z]+)([A-Z][a-z])", r"\1_\2", s)  # Split acronyms followed by normal camelCase
     s = re.sub(r"([a-zA-Z])([0-9]+)", r"\1_\2", s)
-    return s.lower()
+    s = s.lower()
+    # Temporary fix to avoid breaking any serialization.
+    s = s.replace("base_64", "base64")
+    return s
```
Lash-L, line 38: Very targetted and not ideal

Issue to track here: #644

allenporter: The current set of tests should be exhaustive, so you can break anything that is not tested.

allenporter: This regexp was modified to match the names in device features so if we just want to change field names in device features it should not be a problem to change. Agree with this change that making serialization work is more important in the short term.

Lash-L: Will reference this in issue for tracking. Will merge for now

Issue #623, opened 2025-12-04
Add conversions for DYAD_PROTOCOL_ENTRIES


allenporter:
We should be doing these conversions to get readable values rather than passing around ints. This may require a breaking change in the API. We also need to add test coverage for this (and consider in upstream products too)

PR #645, opened 2025-12-07
fix: Encode a01 values as json strings

allenporter:
Update the a01 internal values to be json strings inside the dictionary. This matches the old API behavior.

Issue #623


allenporter:
I want to come back to this with better tests

PR #650, opened 2025-12-07
fix: handle different query types

Lash-L:
This fixes A01 devices. Open to tweaks as needed. I believe it will always be a string now? But a check is probably the safest thing rn.

### DIFF
--- a/roborock/devices/a01_channel.py
+++ b/roborock/devices/a01_channel.py
@@ -1,6 +1,7 @@
 """Thin wrapper around the MQTT channel for Roborock A01 devices."""
 
 import asyncio
+import json
 import logging
 from typing import Any, overload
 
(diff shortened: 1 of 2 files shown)


roborock/devices/a01_channel.py
```diff
@@ -54,6 +55,13 @@ async def send_decoded_command(
         await mqtt_channel.publish(roborock_message)
         return {}
 
+    if isinstance(query_values, str):
+        try:
+            query_values = json.loads(query_values)
```
allenporter, line 60: So this is then undoing the str from `ZeoApi.query_values`. That means we probably need to (1) revert that change in the trait, pass in the raw values here, then go back to encoding that inside of`encode_mqtt_payload` and (2) update the tests to verify at the mqtt channel level, decoding the `Roborockmessage` itself? The other caller of `encode_mqtt_payload` should probably push its logic down or we need to make it a flag or something to have that behavior.

tests/test_a01_api.py
```diff
@@ -307,3 +307,38 @@ async def test_future_timeout(
     with patch("roborock.roborock_future.asyncio.timeout", side_effect=asyncio.TimeoutError):
         data = await connected_a01_mqtt_client.update_values([RoborockZeoProtocol.STATE])
     assert data.get(RoborockZeoProtocol.STATE) is None
+
+
+async def test_send_decoded_command_handles_stringified_query() -> None:
```
allenporter, line 312: This should be somewhere under `tests/devices/` instead?

allenporter (changes requested):

allenporter:
I sent #651 which is partial rollback of the PR that introduced the bug, doing it another way

PR #656, opened 2025-12-09
fix: handle random length bytes before version bytes

Lash-L:
So our expected behavior is that if it does not start with the verison bytes, it should start with 4 bytes that give the payload length.

That is not always happening - likely due to the buffer, but it's hard to say. This will fix the problem while we can explore more in depth about the buffer

Partly related to https://github.com/home-assistant/core/issues/157599 ?

Talks about the decoding issue there

### DIFF
--- a/roborock/protocol.py
+++ b/roborock/protocol.py
@@ -537,7 +537,8 @@ def create_local_decoder(local_key: str, connect_nonce: int | None = None, ack_n
         parsed_messages, remaining = MessageParser.parse(
             buffer, local_key=local_key, connect_nonce=connect_nonce, ack_nonce=ack_nonce
         )
-        _LOGGER.debug("Found %d extra bytes: %s", len(remaining), remaining)
+        if remaining:
+            _LOGGER.debug("Found %d extra bytes: %s", len(remaining), remaining)
         buffer = remaining
         return parsed_messages
 


tests/devices/test_local_channel.py
```diff
@@ -147,9 +147,9 @@ async def test_message_decode_error(local_channel: LocalChannel, caplog: pytest.
     local_channel._data_received(b"invalid_payload")
     await asyncio.sleep(0.01)  # yield
 
-    assert len(caplog.records) == 1
-    assert caplog.records[0].levelname == "WARNING"
-    assert "Failed to decode message" in caplog.records[0].message
+    warning_records = [record for record in caplog.records if record.levelname == "WARNING"]
```
allenporter, line 150: I think theres something you can do like `with caplog.at_level(logging.WARNING):` as well surrounding this, here and below

tests/devices/test_local_decoder_padding.py
```diff
@@ -0,0 +1,89 @@
+from roborock.protocol import create_local_decoder, create_local_encoder
```
allenporter, line 1: We currently have the pattern that each test module corresponds to an actual model. I know AI likes to make up random test files depending on the subject, but i'd rather see this correspond to a module.

Should this be in `tests/test_protocol.py`?

tests/devices/test_local_decoder_padding.py
```diff
@@ -0,0 +1,89 @@
...
+        payload=b"test_payload",
+        version=b"1.0",
+        seq=1,
+        random=123,
+    )
+    encoded = encoder(msg)
+
+    decoded = decoder(encoded)
+    assert len(decoded) == 1
+    assert decoded[0].payload == b"test_payload"
```
allenporter, line 22: I believe you can write the first three tests as one test like this:

```suggestion
@pytest.mark.parametrize(
    ("garbage"),
    [
        b"",
        b"\x00\x00\x05\xa1"
        b"\x00\x00\x05\xa1\xff\xff"
    ],
)
def test_decoder_clean_message(garbage: bytes):
    encoder = create_local_encoder(TEST_LOCAL_KEY)
    decoder = create_local_decoder(TEST_LOCAL_KEY)

    msg = RoborockMessage(
        protocol=RoborockMessageProtocol.RPC_REQUEST,
        payload=b"test_payload",
        version=b"1.0",
        seq=1,
        random=123,
    )
    encoded = encoder(msg)

    decoded = decoder(garbage + encoded)
    assert len(decoded) == 1
    assert decoded[0].payload == b"test_payload"
```

PR #658, opened 2025-12-10
feat: log when we see a new key we have never seen before

Lash-L:
This will help with future reverse engineering, we only log once


allenporter (commented):
Could also consider storing this in diagnostic data for the traits to dump if thats also helpful.

Lash-L:
> Could also consider storing this in diagnostic data for the traits to dump if thats also helpful.

A good follow up, I'll explore doing that as well

PR #662, opened 2025-12-13
feat: add b01 Q7 basic getter support

Lash-L:


roborock/devices/b01_channel.py
```diff
@@ -2,26 +2,76 @@
...
+
+    def find_response(response_message: RoborockMessage) -> None:
+        """Handle incoming messages and resolve the future."""
+        try:
+            decoded_dps = decode_rpc_response(response_message)
+        except RoborockException as ex:
+            _LOGGER.info("Failed to decode b01 message: %s: %s", response_message, ex)
+            return
+
+        for _, dps_value in decoded_dps.items():
```
allenporter:
```suggestion
        for dps_value in decoded_dps.values():
```

roborock/devices/b01_channel.py
```diff
@@ -2,26 +2,76 @@
...
+                    continue
+
+                if isinstance(inner, dict) and inner.get("msgId") == msg_id:
+                    _LOGGER.debug("Received query response: %s", inner)
+                    data = inner.get("data")
+                    if isinstance(data, dict):
+                        result.update(data)
+                    finished.set()
+            else:
+                _LOGGER.debug("Received unexpected response: %s", dps_value)
```
allenporter: Avoid the extra indent with early exit

```suggestion
            if not isinstance(dps_value, str):
                _LOGGER.debug("Received unexpected response: %s", dps_value)
                continue

            try:
                inner = json.loads(dps_value)
            except (json.JSONDecodeError, TypeError):
                _LOGGER.debug("Received unexpected response: %s", dps_value)
                continue

            if isinstance(inner, dict) and inner.get("msgId") == msg_id:
                _LOGGER.debug("Received query response: %s", inner)
                data = inner.get("data")
                if isinstance(data, dict):
                    result.update(data)
                finished.set()
```

roborock/devices/b01_channel.py
```diff
@@ -2,26 +2,76 @@
...
     params: ParamsType,
-) -> None:
+) -> dict[str, Any]:
     """Send a command on the MQTT channel and get a decoded response."""
     _LOGGER.debug("Sending MQTT command: %s", params)
-    roborock_message = encode_mqtt_payload(dps, command, params)
-    await mqtt_channel.publish(roborock_message)
+    msg_id = str(get_next_int(100000000000, 999999999999))
+    roborock_message = encode_mqtt_payload(dps, command, params, msg_id)
+    finished = asyncio.Event()
```
allenporter: `asyncio.Future` seems more appropriate here than  event + separate result object.

roborock/protocols/b01_protocol.py
```diff
@@ -22,13 +22,13 @@
...
 
-def encode_mqtt_payload(dps: int, command: CommandType, params: ParamsType) -> RoborockMessage:
+def encode_mqtt_payload(dps: int, command: CommandType, params: ParamsType, msg_id: str) -> RoborockMessage:
     """Encode payload for B01 commands over MQTT."""
     dps_data = {
         "dps": {
             dps: {
                 "method": str(command),
-                "msgId": str(get_next_int(100000000000, 999999999999)),
+                "msgId": msg_id or str(get_next_int(100000000000, 999999999999)),
```
allenporter: Why have this empty msg_id behavior?

tests/devices/traits/b01/test_init.py
```diff
@@ -0,0 +1,130 @@
+import json
```
allenporter, line 1: test_init.py?

Lash-L: Pytest was getting mad when it had a matching name to a01 test init for some reason

roborock/devices/b01_channel.py
```diff
@@ -2,26 +2,76 @@
...
+                    inner = json.loads(dps_value)
+                except (json.JSONDecodeError, TypeError):
+                    _LOGGER.debug("Received unexpected response: %s", dps_value)
+                    continue
+
+                if isinstance(inner, dict) and inner.get("msgId") == msg_id:
+                    _LOGGER.debug("Received query response: %s", inner)
+                    data = inner.get("data")
+                    if isinstance(data, dict):
+                        result.update(data)
```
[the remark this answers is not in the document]
allenporter: I think this is a valid issue. if we match the msg_id then we have the response and we should reply -- even with an exception -- rather than timeout.

tests/devices/traits/b01/test_init.py
Lash-L: see here: https://github.com/Python-roborock/python-roborock/actions/runs/20200816620/job/57991726603?pr=662#step:5:41

allenporter: Ah i think i have seen this before and it can be resolved by adding an `__init__.py` file in the pytest directory.

Alternatively, we can stop using `__init_.py` in the traits since it may be hard to find the actual content in these subdirectories.

Lash-L: for now i have added a init as the alternative is a larger change

roborock/devices/b01_channel.py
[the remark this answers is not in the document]
Lash-L: raise an exception to clarify what is going wrong

allenporter (approved):

roborock/devices/b01_channel.py
```diff
@@ -2,26 +2,76 @@
...
+
+    def find_response(response_message: RoborockMessage) -> None:
+        """Handle incoming messages and resolve the future."""
+        try:
+            decoded_dps = decode_rpc_response(response_message)
+        except RoborockException as ex:
+            _LOGGER.info("Failed to decode b01 message: %s: %s", response_message, ex)
+            return
+
+        for dps_value in decoded_dps.values():
```
allenporter, line 46: For a future PR consider if `decode_rpc_response` should return a dict of parsed values similar to [a01 decode](https://github.com/Python-roborock/python-roborock/blob/4b4e5aa48fcc2ca2d091e3a2c3bc737b121cb878/roborock/protocols/a01_protocol.py#L54) then unit tests for corner cases can be more narrow without the full test setup.

PR #672, opened 2025-12-14
fix: Share a HealthManager instance across all mqtt channels

allenporter:
The motiviation is to better isolate single device timeouts from overall session health. I have not seen this as an explicit issue, but seems worth being defensive against. This will ensure that a single inactive device will not unnecessarily restart the MQTT session. Any successful RPCs by other devices will reset the counter.

Adds explicit logging when a session is reset by the health manager.

### DIFF
--- a/roborock/devices/v1_channel.py
+++ b/roborock/devices/v1_channel.py
@@ -181,7 +181,6 @@ class V1Channel(Channel):
         self._logger = RoborockLoggerAdapter(device_uid, _LOGGER)
         self._security_data = security_data
         self._mqtt_channel = mqtt_channel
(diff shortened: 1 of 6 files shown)


roborock/devices/v1_channel.py
```diff
@@ -272,7 +271,7 @@ def _create_mqtt_rpc_strategy(self, decoder: Callable[[RoborockMessage], Any] =
                 security_data=self._security_data,
             ),
             decoder=decoder,
-            health_manager=self._mqtt_health_manager,
+            health_manager=self._mqtt_channel.health_manager,
```
allenporter, line 274: Note to reviewer: The key change is here where the health manager will be shared across devices using the common mqtt session health manger.

Issue #346, opened 2025-03-13
Add Switches for scheduled cleaning


matthabjan:
Hi everyone

I am using the Roborock HA integration and it would be really cool if I could de-/activate schedules for cleaning.

I have a few schedules for cleaning for different areas and would like to simply switch them on and off from the dashboard.

Any chance of adding support for this?

cheers,
Matthias

allenporter:
I think a next step here is to figure out what the API is for getting schedules.

PR #675, opened 2025-12-14
feat: add basic schedule getting

Lash-L:
This adds some basic support towards #346

Updating is slightly more complicated, as you don't just send enabled/disabled but the entire data again. so I did not want to spend the time on it, but this should be a good basis for someone else to work off of or for me to come back to in the future.

### DIFF
--- a/roborock/web_api.py
+++ b/roborock/web_api.py
@@ -628,7 +628,7 @@ class RoborockApiClient:
         if isinstance(schedules, list):
             return [HomeDataSchedule.from_dict(schedule) for schedule in schedules]
         else:
-            raise RoborockException(f"scene_response result was an unexpected type: {result}")
+            raise RoborockException(f"schedule_response result was an unexpected type: {schedules}")
 
     async def get_products(self, user_data: UserData) -> ProductResponse:
         """Gets all products and their schemas, good for determining status codes and model numbers."""


roborock/web_api.py
```diff
@@ -607,6 +608,28 @@ async def execute_scene(self, user_data: UserData, scene_id: int) -> None:
...
+        rriot = user_data.rriot
+        if rriot is None:
+            raise RoborockException("rriot is none")
+        if rriot.r.a is None:
+            raise RoborockException("Missing field 'a' in rriot reference")
+        schedules_request = PreparedRequest(
+            rriot.r.a,
+            self.session,
+            {
+                "Authorization": _get_hawk_authentication(rriot, f"/user/devices/{str(device_id)}/jobs"),
```
allenporter: The `str` should not be needed because (a) device_id is already str and (b) the string formatting is equivalent to `str(x)` anyway:

```suggestion
                "Authorization": _get_hawk_authentication(rriot, f"/user/devices/{device_id}/jobs"),
```

roborock/web_api.py
```diff
@@ -607,6 +608,28 @@ async def execute_scene(self, user_data: UserData, scene_id: int) -> None:
...
+            },
+        )
+        schedules_response = await schedules_request.request("get", f"/user/devices/{str(device_id)}/jobs")
+        if not schedules_response.get("success"):
+            raise RoborockException(schedules_response)
+        schedules = schedules_response.get("result")
+        if isinstance(schedules, list):
+            return [HomeDataSchedule.from_dict(schedule) for schedule in schedules]
+        else:
+            raise RoborockException("scene_response result was an unexpected type")
```
allenporter: Maybe include the bad data or its type
```suggestion
            raise RoborockException(f"scene_response result was an unexpected type: {result}")
```

allenporter (approved):
Nice!

PR #676, opened 2025-12-14
chore: add snapshot tests for device payloads

allenporter:
This adds a fixture that captures the requests and responses in the test so they can be snapshotted. I will use this to build similar test functionality for the new API and assert on the same test contents.

Tests are printed with a hex dump in a tcpdump style. The capture fixture logs messages in order, then the test itself decides when to snapshot the request log (e.g. before racing shutdown messages)

This appears to be stable but if it is flaky in CI for the old framework we can disable it and only enable with a flag for local testing, and use it in the new API only.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -65,8 +65,6 @@ class CapturedRequestLog:
 
         This makes the packets easier to read and compare in test snapshots.
 
-
-
         Args:
             data: The bytes object to print.
             bytes_per_line: The number of bytes to display per line (default is 16).


roborock/roborock_message.py
```diff
@@ -245,4 +243,4 @@ class RoborockMessage:
     seq: int = field(default_factory=lambda: get_next_int(100000, 999999))
     version: bytes = b"1.0"
     random: int = field(default_factory=lambda: get_next_int(10000, 99999))
-    timestamp: int = field(default_factory=lambda: math.floor(time.time()))
+    timestamp: int = field(default_factory=lambda: get_timestamp())
```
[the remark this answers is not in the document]
allenporter: This is needed for the tests.

tests/conftest.py
```diff
@@ -27,6 +27,130 @@
...
+            lines.extend(self._hexdump(data))
+        return "\n".join(lines)
+
+    def _hexdump(self, data: bytes, bytes_per_line: int = 16) -> list[str]:
+        """Print a hexdump of the given bytes object in a tcpdump/hexdump -C style.
+
+        This makes the packets easier to read and compare in test snapshots.
+
+
+
```
Lash-L: not a big deal but a lot of spacing here

tests/conftest.py
```diff
@@ -27,6 +27,130 @@
...
+    def get_token_bytes(n: int) -> bytes:
+        nonlocal token_chr
+        result = token_chr.encode() * n
+        # Cycle to the next character
+        token_chr = chr(ord(token_chr) + 1)
+        if token_chr > "Z":
+            token_chr = "A"
+        return result
+
+    with (
```
Lash-L, line 134: would patching globally not work here? i.e. roborock.util.get_next_int

Genuine question, I know patching can be weird

Lash-L (approved):
Cool! LGTM

tests/conftest.py
allenporter: It can work, yes. I think generally it is preferred to patch at the call site, to make sure not to have unintended side effects on other things, but here given the number of calls you could make an argument its not worth it or could be brittle. There are a couple calls in the web API i didn't necessarily need to patch so figured leaving them out is helpful for now.  I also considered that we might not want so many calls sites (since i realize they are both on two versions of the request message, which is probably error prone) so enumerating them would be helpful to reduce the number.

PR #688, opened 2025-12-17
fix: Handle AppInitStatus with omitted new_feature_info_str

allenporter:
The approach taken here is:
(1) add an example of this response type to the json examples to exercise the decoder. (This succeeds already, but is just used to record the payload) (2) Add a test for passing the decoded data to the dataclass.

In the future this can be improved with trait tests that use these payloads, but this is a way to do it incrementally for now.

Issue: https://github.com/home-assistant/core/issues/158049

### DIFF
--- a/tests/test_containers.py
+++ b/tests/test_containers.py
@@ -646,3 +646,4 @@ def test_partial_app_init_status() -> None:
     )
     assert app_init_status.local_info.name == "custom_A.03.0096_FCC"
     assert app_init_status.feature_info == [111, 112, 113, 114, 115, 116, 117, 118, 119, 120, 121, 122, 123, 124, 125]
+    assert app_init_status.new_feature_info_str == ""


roborock/data/v1/v1_containers.py
```diff
@@ -585,7 +585,7 @@ class AppInitStatus(RoborockBase):
     local_info: AppInitStatusLocalInfo
     feature_info: list[int]
     new_feature_info: int
-    new_feature_info_str: str
+    new_feature_info_str: str | None = None
```
Lash-L: maybe it makes sense to default ot "" here rather than adding a type check later?

allenporter: Updated.

tests/test_containers.py
```diff
@@ -600,3 +600,49 @@ def test_offline_device() -> None:
...
+                "camera_status": 3495,
+                "is_exploring": 0,
+                "home_sec_status": 0,
+                "home_sec_enable_password": 1,
+                "adbumper_status": [0, 0, 0],
+            },
+        }
+    )
+    assert app_init_status.local_info.name == "custom_A.03.0096_FCC"
+    assert app_init_status.feature_info == [111, 112, 113, 114, 115, 116, 117, 118, 119, 120, 121, 122, 123, 124, 125]
```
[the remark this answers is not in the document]
allenporter: Did this suggestion.

PR #689, opened 2025-12-18
Add 2 new states for ZeoState in zeo_code_mappings

NOisi-x:
Add 2 new states for ZeoState in zeo_code_mappings.
aftercare = 12
waitingforaftercare = 13

### DIFF
--- a/roborock/data/zeo/zeo_code_mappings.py
+++ b/roborock/data/zeo/zeo_code_mappings.py
@@ -21,6 +21,7 @@ class ZeoState(RoborockEnum):
     aftercare = 12
     waiting_for_aftercare = 13
 
+
 class ZeoProgram(RoborockEnum):
     standard = 1
     quick = 2


roborock/data/zeo/zeo_code_mappings.py
```diff
@@ -18,7 +18,8 @@ class ZeoState(RoborockEnum):
     cooling = 8
     under_delay_start = 9
     done = 10
-
+    aftercare = 12
+    waitingforaftercare = 13
```
allenporter: Should the value have underscores between words similar to `under_delay_start`?

(I assuming this is parsed by number and not string name)

NOisi-x: It is indeed parsed by number and the string name is defined, so it can be changed if there are any formatting requirements for the string name.

allenporter: OK follow the existing pattern please where underscores are used between words

NOisi-x: Done. Thank you.

roborock/data/zeo/zeo_code_mappings.py
```diff
@@ -18,7 +18,8 @@ class ZeoState(RoborockEnum):
     cooling = 8
     under_delay_start = 9
     done = 10
-
+    aftercare = 12
+    waiting_for_aftercare = 13
```
Lash-L, line 23:
```suggestion
    waiting_for_aftercare = 13


```

PR #690, opened 2025-12-19
feat: add some basic setters for q7

Lash-L:
This should be all we need for the vacuum entity.

### DIFF
--- a/roborock/devices/b01_channel.py
+++ b/roborock/devices/b01_channel.py
@@ -28,7 +28,7 @@ async def send_decoded_command(
     dps: int,
     command: CommandType,
     params: ParamsType,
-) -> dict | None:
+) -> dict[str, Any] | None:
     """Send a command on the MQTT channel and get a decoded response."""
     msg_id = str(get_next_int(100000000000, 999999999999))
     _LOGGER.debug(
@@ -104,6 +104,17 @@ async def send_decoded_command(
         raise RoborockException(
             f"B01 command timed out after {_TIMEOUT}s (method={command}, msg_id={msg_id}, dps={dps}, params={params})"
         ) from ex
+    except RoborockException as ex:
+        _LOGGER.warning(
+            "Error sending B01 decoded command (method=%s msg_id=%s dps=%s params=%s): %s",
+            command,
+            msg_id,
+            dps,
+            params,
+            ex,
+        )
+        raise
+
     except Exception as ex:
         _LOGGER.exception(
             "Error sending B01 decoded command (method=%s msg_id=%s dps=%s params=%s): %s",
--- a/roborock/devices/traits/b01/q7/__init__.py
+++ b/roborock/devices/traits/b01/q7/__init__.py
@@ -38,24 +38,24 @@ class Q7PropertiesApi(Trait):
             raise TypeError(f"Unexpected response type for GET_PROP: {type(result).__name__}: {result!r}")
         return B01Props.from_dict(result)
 
-    async def set_prop(self, prop: RoborockB01Props, value: Any) -> Any:
+    async def set_prop(self, prop: RoborockB01Props, value: Any) -> None:
         """Set a property on the device."""
         await self.send(
             command=RoborockB01Q7Methods.SET_PROP,
             params={prop: value},
         )
 
-    async def set_fan_speed(self, fan_speed: SCWindMapping) -> Any:
+    async def set_fan_speed(self, fan_speed: SCWindMapping) -> None:
         """Set the fan speed (wind)."""
-        return await self.set_prop(RoborockB01Props.WIND, fan_speed.code)
+        await self.set_prop(RoborockB01Props.WIND, fan_speed.code)
 
-    async def set_water_level(self, water_level: WaterLevelMapping) -> Any:
+    async def set_water_level(self, water_level: WaterLevelMapping) -> None:
         """Set the water level (water)."""
-        return await self.set_prop(RoborockB01Props.WATER, water_level.code)
+        await self.set_prop(RoborockB01Props.WATER, water_level.code)
 
-    async def start_clean(self) -> Any:
+    async def start_clean(self) -> None:
         """Start cleaning."""
-        return await self.send(
+        await self.send(
             command=RoborockB01Q7Methods.SET_ROOM_CLEAN,
             params={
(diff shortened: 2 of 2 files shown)


roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -28,23 +28,27 @@ def __init__(self, channel: MqttChannel) -> None:
         """Initialize the B01Props API."""
         self._channel = channel
 
-    async def query_values(self, props: list[RoborockB01Props]) -> B01Props | None:
-        """Query the device for the values of the given Q7 properties."""
-        result = await send_decoded_command(
+    async def send(self, command: RoborockB01Q7Methods, params: dict) -> Any:
```
allenporter: Add typing

```suggestion
    async def send(self, command: RoborockB01Q7Methods, params: dict[str, Any]) -> Any:
```

allenporter: Is this intended to be publicly called by clients? If no, then `_send`. If yes then add a pydoc string.

Should this be the first function the user reads about in this file or should we put it below all the higher level API calls?

roborock/devices/b01_channel.py
```diff
@@ -28,12 +28,15 @@ async def send_decoded_command(
     dps: int,
     command: CommandType,
     params: ParamsType,
-) -> dict[str, Any]:
-    """Send a command on the MQTT channel and get a decoded response."""
+) -> Any:
+    """Send a command on the MQTT channel and get a decoded response.
+
+    Note: B01 "set" commands may return a scalar (e.g. 0/"ok") rather than a dict.
```
allenporter: Do we need to expose that past this function though? That seems fine to just return either `None` an empty dict in that case rather than making the caller figure it out.

roborock/devices/b01_channel.py
```diff
@@ -58,20 +61,24 @@ def find_response(response_message: RoborockMessage) -> None:
...
         await mqtt_channel.publish(roborock_message)
-        try:
-            return await asyncio.wait_for(future, timeout=_TIMEOUT)
-        except TimeoutError as ex:
-            raise RoborockException(f"Command timed out after {_TIMEOUT}s") from ex
+        return await asyncio.wait_for(future, timeout=_TIMEOUT)
+    except TimeoutError as ex:
+        raise RoborockException(f"Command timed out after {_TIMEOUT}s") from ex
+    except Exception as ex:
+        _LOGGER.exception("Error sending decoded command: %s", ex)
```
allenporter: Catch `RoborockException` and raise? Catching bare exception and logging exception should be for unexpected cases only

roborock/devices/b01_channel.py
allenporter: (What do they return on failure?)

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -20,13 +28,86 @@ def __init__(self, channel: MqttChannel) -> None:
...
         return B01Props.from_dict(result)
 
+    async def set_prop(self, prop: RoborockB01Props, value: Any) -> Any:
+        """Set a property on the device."""
+        await self.send(
+            command=RoborockB01Q7Methods.SET_PROP,
+            params={prop: value},
+        )
+
+    async def set_fan_speed(self, fan_speed: SCWindMapping) -> Any:
```
allenporter: If all of these are returning 'ok/...` I would say don't return that here and below.

allenporter (changes requested):

roborock/devices/traits/b01/q7/__init__.py
Lash-L: well we at the very least will need it for send_command() but i can separate that out into a separate trait like we do for V1

allenporter: I don't have a preference. Right now this has a single API for the entire protocol which is fine with me if you wan tto keep it here. Then just add the pydoc string, and i'd still suggest moving it lower in the file.

roborock/devices/b01_channel.py
Lash-L: the code will be non-zero on error. I have updated to add that case

roborock/devices/b01_channel.py
```diff
@@ -28,19 +28,31 @@ async def send_decoded_command(
     dps: int,
     command: CommandType,
     params: ParamsType,
-) -> dict[str, Any]:
+) -> dict | None:
```
allenporter: Is this still `dict[str, Any]`?

roborock/devices/b01_channel.py
```diff
@@ -57,21 +69,50 @@ def find_response(response_message: RoborockMessage) -> None:
...
-        try:
-            return await asyncio.wait_for(future, timeout=_TIMEOUT)
-        except TimeoutError as ex:
-            raise RoborockException(f"Command timed out after {_TIMEOUT}s") from ex
+        return await asyncio.wait_for(future, timeout=_TIMEOUT)
+    except TimeoutError as ex:
+        raise RoborockException(
+            f"B01 command timed out after {_TIMEOUT}s (method={command}, msg_id={msg_id}, dps={dps}, params={params})"
+        ) from ex
+    except Exception as ex:
```
allenporter, line 118: Add separate logging for `RoborockException` and `Exception`. You can make `Exception` for unexpected uncaught exceptions but `RoborockException` probably should not be `logger.exception` but `warning`?

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -22,11 +30,85 @@ def __init__(self, channel: MqttChannel) -> None:
...
-            self._channel, dps=10000, command=RoborockB01Q7Methods.GET_PROP, params={"property": props}
+        result = await self.send(
+            RoborockB01Q7Methods.GET_PROP,
+            {"property": props},
         )
+        if not isinstance(result, dict):
+            raise TypeError(f"Unexpected response type for GET_PROP: {type(result).__name__}: {result!r}")
         return B01Props.from_dict(result)
 
+    async def set_prop(self, prop: RoborockB01Props, value: Any) -> Any:
```
allenporter: Make return type None for these setters here and below.

```suggestion
    async def set_prop(self, prop: RoborockB01Props, value: Any) -> None:
```

PR #692, opened 2025-12-19
feat: add some basic support for q10

Lash-L:
Accidentally based on top of the Q7 work. Will finish that and then rebase this

### DIFF
--- a/roborock/devices/traits/b01/q7/__init__.py
+++ b/roborock/devices/traits/b01/q7/__init__.py
@@ -109,63 +109,6 @@ class Q7PropertiesApi(Trait):
             params=params,
         )
 
-    async def start_clean(self) -> dict[str, Any]:
-        """Start cleaning."""
-        return await send_decoded_command(
-            self._channel,
-            dps=10000,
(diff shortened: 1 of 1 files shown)


roborock/devices/device.py
```diff
@@ -214,10 +215,24 @@ async def close(self) -> None:
         if self._unsub:
             self._unsub()
             self._unsub = None
+        close_trait = getattr(self._trait, "close", None)
```
allenporter, line 219: I would say add this to the trait protoocl and make it a no-op for the others.

roborock/devices/device.py
```diff
@@ -214,10 +215,24 @@ async def close(self) -> None:
...
+                result = close_trait()
+                if asyncio.iscoroutine(result):
+                    await result
+            except Exception as ex:  # noqa: BLE001
+                self._logger.debug("Error closing trait: %s", ex)
 
     def _on_message(self, message: RoborockMessage) -> None:
         """Handle incoming messages from the device."""
         self._logger.debug("Received message from device: %s", message)
+        on_message = getattr(self._trait, "on_message", None)
```
allenporter, line 231: Let the trait add their own callback.

PR #699, opened 2025-12-22
Organize test fixtures to prepare for more end to end tests

allenporter:
Rewrite the test fixtures to have a more clear split between local and mqtt fixtures. This is in prepration for running both at once for e2 tests for device manager.

All fixtures are moved into a fixtures subdirectory. The helper classes that are imported into other tests are added in separate files for importing, and to avoid import warnings from pytests.

This renames all the fixtures to have mqtt prefixed names and local fixtures to have local prefixed names. There is one minor change to make the local asyncio tests uses asyncio Queues rather than blocking queues.

### DIFF
--- a/tests/fixtures/logging_fixtures.py
+++ b/tests/fixtures/logging_fixtures.py
@@ -3,67 +3,10 @@ from unittest.mock import patch
 
 import pytest
 
+from .logging import CapturedRequestLog
+
 # Fixed timestamp for deterministic tests for asserting on message contents
 FAKE_TIMESTAMP = 1755750946.721395
-
-
-class CapturedRequestLog:
-    """Log of requests and responses for snapshot assertions.
-
-    The log captures the raw bytes of each request and response along with
-    a label indicating the direction of the message.
-    """
-
-    def __init__(self) -> None:
-        """Initialize the request log."""
-        self.entries: list[tuple[str, bytes]] = []
-
-    def add_log_entry(self, label: str, data: bytes) -> None:
-        """Add a request entry."""
-        self.entries.append((label, data))
-
-    def __repr__(self):
-        """Return a string representation of the log entries.
-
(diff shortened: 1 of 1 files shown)


tests/fixtures/local_async_fixtures.py
```diff
@@ -0,0 +1,77 @@
...
+from collections.abc import Awaitable, Callable, Generator
+from typing import Any
+from unittest.mock import Mock, patch
+
+import pytest
+
+AsyncLocalRequestHandler = Callable[[bytes], Awaitable[bytes | None]]
+
+
+@pytest.fixture(name="local_received_requests")
```
allenporter, line 15: This file was copied, then modified to use `asyncio.Queue` instead of `queue.Queue`

tests/test_local_api_v1.py
```diff
@@ -1,23 +1,122 @@
...
+pytest_plugins = [
+    "tests.fixtures.logging_fixtures",
+]
+
+QUEUE_TIMEOUT = 10
+
+LocalRequestHandler = Callable[[bytes], bytes | None]
+
+
+@pytest.fixture(name="local_received_requests")
```
allenporter, line 37: The synchronous local request handler fixtures were moved here since it is only used by this test.

tests/test_api.py
```diff
@@ -34,7 +35,33 @@
...
+    )
+    client = RoborockMqttClientV1(user_data, device_info, queue_timeout=QUEUE_TIMEOUT)
+    try:
+        yield client
+    finally:
+        if not client.is_connected():
+            try:
+                await client.async_release()
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
allenporter: Don't want to address this one.

Lash-L (approved):

PR #704, opened 2025-12-23
feat: Small tweaks to test fixtures 

allenporter:
Update the test fixtures with some cleanups pulled out of a larger PR:
- Be more careful when patching the event loop to avoid side effects in other tests given the running loop is a global variable
- Update the mqtt fixture to wait to populate the response buffer until a client is connected. This is more correct, but also helps make the test buffer output more readable so requests and responses are in order
- Add warnings for unconsumed responses and remove some unconsumed responses


allenporter:
This now contains a feat commit since I want to build a new release.

Lash-L (approved):
looks good - just make sure that when you squash you change hte commit title, i don't think the descriptions will cause a bump

PR #707, opened 2025-12-24
fix: Allow startup with unsupported devices

allenporter:
Raise and catch when finding an unsupported device. We will allow the device manager to continue with the devices that are supported, ignoring the ones that are not supoprted.

We test with an unsupported protocol version but it also applies to unsupported q10 devices.

Issue: https://github.com/home-assistant/core/issues/159656


Lash-L:
my GitHub app has been on the fritz recently, it won’t let me comment on the actual line, but while you’re there could you fix the unsupported device message? I did strip instead of split on accident

PR #708, opened 2025-12-24
fix: Improve device startup connection reliability for L01 devices

allenporter:
Increase the timeout waiting for the initial device startup to happen to wait beyond the L01 fallback behavior. Before this change, the device returns and is not yet connected locally (the test fails) forcing MQTT requests.

Currently in draft as it depends on #705

Addresses the symptoms in https://github.com/home-assistant/core/issues/159609


tests/e2e/test_device_manager.py
```diff
@@ -268,3 +267,81 @@ async def test_device_manager(
...
+    device_manager = await device_manager_factory(user_params)
+
+    # The mocked Home Data API returns a single v1 device
+    devices = await device_manager.get_devices()
+    assert len(devices) == 1
+    device = devices[0]
+    assert device.duid == "abc123"
+    assert device.name == "Roborock S7 MaxV"
+    assert device.is_connected
+    assert device.is_local_connected
```
allenporter, line 333: Note: This used to always return false here before the timeout change.

Lash-L (approved):

PR #713, opened 2025-12-27
feat(api)!: Remove original Cloud and Local APIs

allenporter:
Remove the old Cloud / Local APIs, and associated tests and shapshots.
These APIs have been replaced by the device manager API that createsa single single MQTT session shared across all devices.

See example/ for how to use the new API.
 

BREAKING CHANGE: Removes older cloud and local APIs.


roborock/version_1_apis/__init__.py
pelson, line 1: Note that currently the CLI still depends on this subpackage

allenporter: Thanks, good catch. I'll remove from there also.

allenporter: https://github.com/Python-roborock/python-roborock/pull/717

PR #714, opened 2025-12-27
chore: Split up test_containers.py into data subdirectories

allenporter:
Split up the existing containers test into tests for the protocol specific subdirectories, following the newr roborock.data module structure.

This also updates all mock_data references to reference the tests path rather than a relative path.


Lash-L (approved):
good cleanliness change

PR #709, opened 2025-12-27
feat: Add initial Q10 support

allenporter:
Add initial Q10 device support. This device works a bit different than q7 so the protocol and channel are forked.

- Add CLI support for sending commands
- Subscribe to messages with basic parsing/printing the debug log, but no capturing state
- Listen to messages with a message stream and a while loop rather than callbacks (aiomqtt like)
- Add trait commands for start clean/stop/pause/resume/return to dock/etc.

PR #718, opened 2025-12-29
fix: Add b01 q10 protocol encoding/decoding and tests

allenporter:
Pulled this out from #692 and #709 into a smaller PR.

This pushes some of the enum encoding into the protocol library since it is common across all b01 q10 requests. This handles the flattening of COMMON results into the base layer to make processing results easier in the channel / trait layer.

### DIFF
--- a/roborock/data/code_mappings.py
+++ b/roborock/data/code_mappings.py
@@ -55,6 +55,7 @@ class RoborockModeEnum(StrEnum):
     """A custom StrEnum that also stores an integer code for each member."""
 
     code: int
+    """The integer code associated with the enum member."""
 
     def __new__(cls, value: str, code: int) -> RoborockModeEnum:
         """Creates a new enum member."""
@@ -68,7 +69,18 @@ class RoborockModeEnum(StrEnum):
         for member in cls:
             if member.code == code:
                 return member
-        raise ValueError(f"{code} is not a valid code for {cls.__name__}")
+        message = f"{code} is not a valid code for {cls.__name__}"
+        if message not in completed_warnings:
+            completed_warnings.add(message)
+            _LOGGER.warning(message)
+        raise ValueError(message)
+
+    @classmethod
+    def from_code_optional(cls, code: int) -> RoborockModeEnum | None:
+        try:
+            return cls.from_code(code)
+        except ValueError:
+            return None
 
     @classmethod
     def from_value(cls, value: str) -> RoborockModeEnum:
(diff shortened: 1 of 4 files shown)


roborock/protocols/b01_q10_protocol.py
```diff
@@ -0,0 +1,91 @@
...
+    # The COMMON response contains nested datapoints need conversion. To simplify
+    # response handling at higher levels we flatten these into the main result.
+    if B01_Q10_DP.COMMON in result:
+        common_result = result.pop(B01_Q10_DP.COMMON)
+        if not isinstance(common_result, dict):
+            raise RoborockException(f"Invalid dpCommon format: expected dict, got {type(common_result).__name__}")
+        try:
+            common_dps_result = _convert_datapoints(common_result, message)
+        except ValueError as e:
+            raise RoborockException(f"Invalid dpCommon format: {e}") from e
```
Lash-L, line 85: i'm unconvinced we should error out here. i.e. a new data point is added, rather than erroring out, we should be made aware of it, but if there are other dps that we DO have, we should continue

allenporter: Done, updated to add logic into the enum parsing to handle logging unknown codes and to support optional parsing. Updated the tests to exercise this case with a mix of known and unknown DPS keys.

roborock/data/code_mappings.py
```diff
@@ -68,7 +69,18 @@ def from_code(cls, code: int) -> RoborockModeEnum:
...
+            completed_warnings.add(message)
+            _LOGGER.warning(message)
+        raise ValueError(message)
+
+    @classmethod
+    def from_code_optional(cls, code: int) -> RoborockModeEnum | None:
+        try:
+            return cls.from_code(code)
+        except ValueError:
+            return None
```
Lash-L, line 83: Fine with this - how we handled it on RoborockEnum though is that we had a missing attribute that would automatically be set when you give it a invalid value. Fine either direction though.

Lash-L (approved):

roborock/data/code_mappings.py
allenporter: I tried to add it that way but it doesn't work well with the enum constructor (it started saying code was required) and I couldn't really figure it out.

allenporter: This seems like it works based on the constructor but not sure if that works here given we're doing from_code. Happy to change it just not sure what to do exactly.

Lash-L: Not sure either.

What we did before was override _missing_ to handle if the value was missing. But to actually take advantage of it, every instance has to have a specific key added, which isn't ideal. As far as I am aware there is no way to do it from a parent class automatically as enums don't inherit members.

I have no issue with this approach, we will just have to play it by ear and see if we need to change how we handle anything.

PR #721, opened 2025-12-29
feat: Recognize Q10 devices and add a command trait 

allenporter:
This adds CLI support and end to end tests in the device manager.

This has a large test refactoring that restructure mock data and snapshot values for unrelated tests: The mock data for devices/products are moved to json files so that we can more easily manipulate the set of devices/products returned in Home Data API response and make it easier to capture for new devices going forward.

This is based on a mix of approaches from #709 and #692

### DIFF
--- a/roborock/devices/b01_q10_channel.py
+++ b/roborock/devices/b01_q10_channel.py
@@ -14,7 +14,6 @@ from roborock.protocols.b01_q10_protocol import (
 from .mqtt_channel import MqttChannel
 
 _LOGGER = logging.getLogger(__name__)
-_TIMEOUT = 10.0
 
 
 async def send_command(


roborock/devices/b01_q10_channel.py
```diff
@@ -0,0 +1,38 @@
...
+from roborock.exceptions import RoborockException
+from roborock.protocols.b01_q10_protocol import (
+    ParamsType,
+    encode_mqtt_payload,
+)
+
+from .mqtt_channel import MqttChannel
+
+_LOGGER = logging.getLogger(__name__)
+_TIMEOUT = 10.0
```
Lash-L: unused right?

allenporter: Removed.

PR #722, opened 2025-12-30
chore: Add a high level direction for improving device and feature discovery

allenporter:
This is meant to try to summarize the device features progress and potential next steps, and discovery pieces that need to be improved to get to the next level of home assistant quality.

### DIFF
--- a/docs/DYNAMIC_DEVICES.md
+++ b/docs/DYNAMIC_DEVICES.md
@@ -1,0 +1,107 @@
+# Roborock Device & Feature Discovery
+
+## Goals
+
+This document outlines the proposed improvements to the Roborock Home Assistant integration. The integration should be **fast**, **resilient**, and **dynamic**. Users should see their devices immediately upon setting up the integration, even if some devices are currently offline. The system should adapt to account changes (adding/removing devices) without requiring a restart.
+
+## Current Challenges
+
+- **Startup Latency:** The integration currently waits for *all* devices to connect and complete their initial data refresh before the integration is considered "ready". A single slow connection delays the entire platform.
+- **Resilience Issues:** If a device is offline during startup, it may not be created in Home Assistant at all, or worse, can cause the integration setup to fail when all devices are offline (temporarily or otherwise).
+- **Manual Management:** Changes to the user's account (adding a new vacuum, removing an old one) are not reflected until Home Assistant is manually restarted. There is also a bug with the initial version of the cache implementation since it never refreshes.
+- **Coupled Availability & Discovery:** Currently, the integration performs a liveness check during setup (`async_setup`). If a device is offline, setup fails because we rely on that initial successful poll to determine *which* entities to create. We cannot create "Unavailable" entities because without the poll, we don't know what features the device supports.
+- **Resource Waste:** We currently automatically connect to *every* discovered device. If a user has disabled a device in Home Assistant, we still maintain an active connection to it.
+
+## Requirements Summary
+
+| Scenario | Current Behavior | Proposed Behavior |
+| :--- | :--- | :--- |
+| **Startup** | Waits for all devices to refresh. | Starts immediately after getting device list. |
+| **Offline Device** | May cause setup failure or missing entities. | Device and entities are created as "Unavailable". |
+| **Unknown Device** | Might crash or fail integration setup. | Gracefully handled using basic generic traits. |
+| **New Device** | Requires HA Restart. | Automatically added in the background. |
+
+## Proposal Overview
+
+### 1. Fast & Resilient Startup
+
+We're moving from a model where we wait for all devices to be online to one where we load the device list and then listen for changes.
+
+#### How it works:
+1.  **Account Load (Cache First)**:
+    - On startup, the `DeviceManager` loads the device list (`HomeData`) from its local cache.
+    - **Result**: The integration setup finishes in milliseconds.
+2.  **Device Setup (Lazy Loading)**:
+    - We iterate through the cached list and create a Home Assistant device and coordinator for *every item*.
+    - Entities are created immediately. If we don't have data yet, they show as **"Unavailable"**.
+    - **Result**: The device is registered and visible in the UI immediately.
+3.  **Connection & Recovery (Background)**:
+    - In the background, `DeviceManager` works to establish MQTT/Local connections.
+    - As devices come online, the `DeviceListener` triggers updates, and entities become "Available".
+    - A background job periodically refreshes `HomeData` from the cloud to catch new devices or removals.
+#### Why this solves our problems:
+- **Zero Latency**: Startup time is constant, not proportional to device count or network speed.
+- **Offline Persistence**: Offline devices remain in the UI (as Unavailable), so users can troubleshoot instead of wondering where they went.
+- **Cloud Resilience**: If the Roborock Cloud is down, the local integration still boots using cached credentials and maps.
+
+---
+
+### 2. Entity Discovery
+
+To verify a device is supported, we'll need to move to the model of checking device features, or only relying on the product information `HomeData` instead of waiting for a live response.
+
+#### Entity Registration
+- **Current Model**: We try to communicate with the device. If it fails, we don't know what sensors to create, so we create nothing.
+- **Proposed Model**: We check the account metadata or cached device features, and create the corresponding sensors immediately.
+
+#### Implementation: Trait Mapping & Cache Bridging
+1.  **Static Mapping**: Where possible, we use existing product data (like protocol version or model) to determine features.
+2.  **Cache Bridging**: If a device is too complex for static mapping (like some V1 vacuums), we use the cache.
+    - *Run 1*: Connect live, discover features, save them to cache.
+    - *Run 2+*: Load features from cache. Setup is instant.
+3.  **Lazy Attributes**: Entities are created based on this metadata, not on whether the device has sent a value yet.
+
+Question: Can existing device features be used confidently now? We likely need to either accelerate adopting this, or create placeholder device features based on previous device state state (e.g. fetch "Device prop" traits once, then decide if a feature is supported)
+
+Alternative: As a temporary measure, we could continue to ensure v1 coordinator devices are refreshed on startup so that they have a valid `coordinator.data`. Given this is not the long term direction we're moving, I propose not to do this.
+
+### 3. Background device discovery
+
+Since we no longer block on startup, we treat account changes as a stream of events:
+- **Auto-Discovery**: New devices in the background data refresh are added dynamically.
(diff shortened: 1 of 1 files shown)


docs/DYNAMIC_DEVICES.md
```diff
@@ -0,0 +1,107 @@
...
+- **Coupled Availability & Discovery:** Currently, the integration performs a liveness check during setup (`async_setup`). If a device is offline, setup fails because we rely on that initial successful poll to determine *which* entities to create. We cannot create "Unavailable" entities because without the poll, we don't know what features the device supports.
+- **Resource Waste:** We currently automatically connect to *every* discovered device. If a user has disabled a device in Home Assistant, we still maintain an active connection to it.
+
+## Requirements Summary
+
+| Scenario | Current Behavior | Proposed Behavior |
+| :--- | :--- | :--- |
+| **Startup** | Waits for all devices to refresh. | Starts immediately after getting device list. |
+| **Offline Device** | May cause setup failure or missing entities. | Device and entities are created as "Unavailable". |
+| **Unknown Device** | Might crash or fail integration setup. | Gracefully handled using basic generic traits. |
```
Lash-L, line 21: by this i assume you mean unknown as in devices we have never seen - rather than 'new' devices i.e. the b01 devices

docs/DYNAMIC_DEVICES.md
```diff
@@ -0,0 +1,107 @@
...
+1.  **Account Load (Cache First)**:
+    - On startup, the `DeviceManager` loads the device list (`HomeData`) from its local cache.
+    - **Result**: The integration setup finishes in milliseconds.
+2.  **Device Setup (Lazy Loading)**:
+    - We iterate through the cached list and create a Home Assistant device and coordinator for *every item*.
+    - Entities are created immediately. If we don't have data yet, they show as **"Unavailable"**.
+    - **Result**: The device is registered and visible in the UI immediately.
+3.  **Connection & Recovery (Background)**:
+    - In the background, `DeviceManager` works to establish MQTT/Local connections.
+    - As devices come online, the `DeviceListener` triggers updates, and entities become "Available".
```
Lash-L, line 40: Ideally, DeviceListener utilizes the existing `RoborockProtocol` to automatically listen for devices.

Devices will ping us with their ip and their duid. We can then automatically add them.

Some instances will have these ports blocked (we should recommend opening them on the docs). If the ports are blocked, we should do a less often Home data call.

docs/DYNAMIC_DEVICES.md
```diff
@@ -0,0 +1,107 @@
...
+- **Offline Persistence**: Offline devices remain in the UI (as Unavailable), so users can troubleshoot instead of wondering where they went.
+- **Cloud Resilience**: If the Roborock Cloud is down, the local integration still boots using cached credentials and maps.
+
+---
+
+### 2. Entity Discovery
+
+To verify a device is supported, we'll need to move to the model of checking device features, or only relying on the product information `HomeData` instead of waiting for a live response.
+
+#### Entity Registration
```
Lash-L, line 53: While less sophisticated, I wonder if utilizing the HA entity manager would be easier - to determine what entities have existed before

docs/DYNAMIC_DEVICES.md
```diff
@@ -0,0 +1,107 @@
...
+- **Current Model**: We try to communicate with the device. If it fails, we don't know what sensors to create, so we create nothing.
+- **Proposed Model**: We check the account metadata or cached device features, and create the corresponding sensors immediately.
```
Lash-L, line 55: If there is a new device that we fail to communicate with - do we still create the device and just add no entities?

docs/DYNAMIC_DEVICES.md
```diff
@@ -0,0 +1,107 @@
...
+The device will manage its own connection state in the background depending on the protocol.
+
+## Work Plan: Implementation TL;DR
+
+### 1. `DeviceManager` Changes
+- Update `discover_devices` to instantiate devices but **do not await** the connection tasks.
+    - *Rationale*: A timeout or connection failure is a *device state*, not a discovery failure. Once we get the list (from cache or cloud), discovery is successful.
+- Update `DeviceManager` to optionally *not* auto-connect. The Home Assistant Coordinator will be responsible for calling `start_connect()` so that disabled devices remain disconnected.
+- Ensure `device_ready` callbacks fire once feature discovery completes even if the initial connection fails
+- Add a periodic task to re-run `discover_devices` every N minutes to fetch fresh `HomeData`.
```
Lash-L, line 99: We should be careful calling home_data often. This is a potential route to rate limiting.

Lash-L:
I wonder if we should follow this rule:

If a feature is supported on only some devices - it should be split up into a separate trait.

This may mean that things like 'Status' need to be split up.

It is a lot easier for us to check/organize the logic if it were checking one trait rather than checking the attribute of a trait. But that may have some complexity to it. What did you have in mind for this part of the system? The actual entity/feature mapping.

docs/DYNAMIC_DEVICES.md
allenporter: I meant here an unsupported device. There was a bug where it would throw an exception and fail the whole integration, even if some devices were supported. I believe this may now be fixed though.

docs/DYNAMIC_DEVICES.md
allenporter: Yeah, that would be the proposal.

docs/DYNAMIC_DEVICES.md
allenporter: Want to propose a frequency? We could also do this:
(1) Never poll in the background
(2) Fetch on startup only if over X hours since last fetch (e.g. if you reload many times in a row it does nothing)

allenporter:
> I wonder if we should follow this rule:
> 
> If a feature is supported on only some devices - it should be split up into a separate trait.
> 
> This may mean that things like 'Status' need to be split up.
> 
> It is a lot easier for us to check/organize the logic if it were checking one trait rather than checking the attribute of a trait. But that may have some complexity to it. What did you have in mind for this part of the system? The actual entity/feature mapping.

This sounds like a reasonable proposal to me.

allenporter:
> > I wonder if we should follow this rule:
> > If a feature is supported on only some devices - it should be split up into a separate trait.
> > This may mean that things like 'Status' need to be split up.
> > It is a lot easier for us to check/organize the logic if it were checking one trait rather than checking the attribute of a trait. But that may have some complexity to it. What did you have in mind for this part of the system? The actual entity/feature mapping.
> 
> This sounds like a reasonable proposal to me.

By this i mean: It sounds like it can work, i don't have any objection to it, and i don't have any other better suggestions.  I like the idea of treating "device features" as an internal implementation detail.

allenporter:
> > > > I wonder if we should follow this rule:
> > > > If a feature is supported on only some devices - it should be split up into a separate trait.
> > > > This may mean that things like 'Status' need to be split up.
> > > > It is a lot easier for us to check/organize the logic if it were checking one trait rather than checking the attribute of a trait. But that may have some complexity to it. What did you have in mind for this part of the system? The actual entity/feature mapping.
> > > 
> > > 
> > > This sounds like a reasonable proposal to me.
> > 
> > 
> > By this i mean: It sounds like it can work, i don't have any objection to it, and i don't have any other better suggestions. I like the idea of treating "device features" as an internal implementation detail.
> 
> Although - some devices have more data for things like dnd than others. and we may end up with something like DnDTrait DnDTraitv2 or DndConsumableTrait
> 
> Then consumables as well can be different per devices, meaning we may need to have a seperate trait per consumable.
> 
> I guess this may get pretty big pretty fast.
> 
> What about a custom decorator for properties?
> 
> If no decorator -> if the device ahs the trait it is supported.
> 
> If decorator -> device must have specific trait(s) for this property to be supported. We can still abstract this up somehow, but then device features stay grouped with the specific thing that they are supporting.

What would the caller API be to check if a feature is supported?

An alternative I can think of is to have an enum with supported features on the trait (an enum value per optional field) and the convention would just be that it has a similar name to the specific fields.

Lash-L:
> > > > > I wonder if we should follow this rule:
> > > > > If a feature is supported on only some devices - it should be split up into a separate trait.
> > > > > This may mean that things like 'Status' need to be split up.
> > > > > It is a lot easier for us to check/organize the logic if it were checking one trait rather than checking the attribute of a trait. But that may have some complexity to it. What did you have in mind for this part of the system? The actual entity/feature mapping.
> > > > 
> > > > 
> > > > This sounds like a reasonable proposal to me.
> > > 
> > > 
> > > By this i mean: It sounds like it can work, i don't have any objection to it, and i don't have any other better suggestions. I like the idea of treating "device features" as an internal implementation detail.
> > 
> > 
> > Although - some devices have more data for things like dnd than others. and we may end up with something like DnDTrait DnDTraitv2 or DndConsumableTrait
> > Then consumables as well can be different per devices, meaning we may need to have a seperate trait per consumable.
> > I guess this may get pretty big pretty fast.
> > What about a custom decorator for properties?
> > If no decorator -> if the device ahs the trait it is supported.
> > If decorator -> device must have specific trait(s) for this property to be supported. We can still abstract this up somehow, but then device features stay grouped with the specific thing that they are supporting.
> 
> What would the caller API be to check if a feature is supported?
> 

theoretically, it could just hit the property and get back some exception? Not sure. Or the property could dynamically pass the information upwards and create a kind of device feature manager. Or the property could set some value on the trait that we can then check.

> An alternative I can think of is to have an enum with supported features on the trait (an enum value per optional field) and the convention would just be that it has a similar name to the specific fields.

That could be fine. entities would end up having a extra field on the description of which enum it needs for that entity to exist.

How do you then map that enum to the actual device feature? i.e. how do we know that this device supports that enum value for that property?

I think device features is likely too abstract of a concept to expose to HA, so I like where we are heading with this.

allenporter:
The points I am hearing that resonate with me are, that I think can be "decisions" we agree on:
- Device features are too granular to use directly and make sense of, therefore we need a new higher level abstraction.
- We will prioritize what "features" are exposed, and they will be a subset. They can be based on any lower level signals or heuristics (could be device features, product info, or hard coded based on model info if that happens to make sense. we can leave that to the specific "feature").
- The exact mechanism has a few options, but likely needs to be able to point at specific fields. it seems like it will be too difficult to make new traits. We'll want an easy way to associate a feature with one more more fields.

if you agree with that, as a next step we can try a couple specifics like what you are proposing and see if we like them. Your proposed API makes sense, so a prototype is probably in order.  I would say we try to use this to identify the existing optional sensors / binary sensors in home assistant (mop, cleaning area, dock entity stuff) and see if we can get something to make sense.

allenporter:
I believe this has all now been implemented, but we don't need this document in the repo necessarily. Closing.

PR #723, opened 2025-12-30
chore: Add additional Home data to diagnostics 

allenporter:
Add home API data including products devices, rooms, etc and increase the number of redacted fields. Add a device manager snapshot test to verify which fields are redacted.

### DIFF
--- a/roborock/devices/device_manager.py
+++ b/roborock/devices/device_manager.py
@@ -4,7 +4,7 @@ import asyncio
 import enum
 import logging
 from collections.abc import Callable, Mapping
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass
 from typing import Any
 
 import aiohttp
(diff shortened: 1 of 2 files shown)


roborock/devices/device_manager.py
```diff
@@ -136,7 +137,11 @@ async def close(self) -> None:
 
     def diagnostic_data(self) -> Mapping[str, Any]:
         """Return diagnostics information about the device manager."""
-        return self._diagnostics.as_dict()
+        return {
+            "home_data": redact_device_data(asdict(self._home_data) if self._home_data else {}),
```
Lash-L: is using our as_dict() functionality better?

allenporter: Yes, it is, good call, fixed.

Lash-L (approved):

PR #726, opened 2025-12-30
chore: Add Saros 10R API response data

allenporter:
Add the home data API product info and device info for a Saros 10R, captured from diagnostics.

Some followup up considerations:
- This is building a parallel world to device_info.yaml so we may want to converge these (e.g. allow either diagnostic or CLI based contributions in the same format)
- The product naming is not very readable since its using the model name which may be hard to read, so we can consider renaming the files.


Lash-L (changes requested):
lint is failing

allenporter:
> lint is failing

Fixed.

PR #729, opened 2025-12-31
A couple of issues encountered with the provided example

CiaranG:
Two commits, one is "docs:" and one is "fix:" because the first is just an error in the example itself (apparently there since it was created) and the other is actually in the library code.

### DIFF
--- a/roborock/devices/traits/v1/device_features.py
+++ b/roborock/devices/traits/v1/device_features.py
@@ -34,7 +34,11 @@ class DeviceFeaturesTrait(DeviceFeatures, common.V1TraitMixin):
             return
         # Save cached device features
         await super().refresh()
-        cache_data.device_features = self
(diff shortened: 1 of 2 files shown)


roborock/devices/traits/v1/device_features.py
```diff
@@ -34,7 +34,11 @@ async def refresh(self) -> None:
             return
         # Save cached device features
         await super().refresh()
-        cache_data.device_features = self
+        # Create a pure DeviceFeatures instance without runtime objects
```
allenporter, line 37: I know this was hacky when I added it in a hurry -- Is this resulting in a bug? I'd like to improve our tests to reflect the issue here since it may apply elsewhere etc.

allenporter:
Converting to draft given there are outstanding questions.

PR #731, opened 2026-01-05
feat: add protocol updates

### LINKED ISSUE #625: Handle automatic updates via protocol data.
The Mqtt topic will send automatic updates for various data points, such as cleaning status or items being removed from the vacuum.

We can take advantage of this for instant updates rather than having to poll.

Lash-L:
Unresolved questions:

1) I am not sure if it makes sense to listen on the trait as a whole or specific properties or specific protocols. Listening on the trait is definitely the easiest from a caller perspective. But from a HA perspective, that means every entity that is connected to that trait has to update
2) We need to manage the mqtt channel to get these protocols - i'm not sure the best way to handle it
3) Need to do tests

closes #625


.gitignore
```diff
@@ -19,3 +19,7 @@ docs/_build/
 
 # GitHub App credentials
 gha-creds-*.json
+
+# pickle files
+*.p
+*.pickle
```
Lash-L, line 25: unrelated but i have a few of these in my repo and just wanted to stop them from filling up my git diff

allenporter (commented):
Great, will be useful to add this.

My initial thought is that for the logic currently in the device: This is all v1 specific logic so it should either live in the v1 channel, or in the trait, and not in the device which is more generic. It can make it's own subscriber on the MQTT channel to also avoid needing to exclude anything from local messages.

For your question about granularity i think we can separate two things:
(1) I do think having the traits have the listener makes sense.  Unclear to me that it needs to pass "self" however, i think that can be omitted and let the caller track which trait it subscribed to if needed.
(2) We can separately decide if we want the coordinator to listen to all traits and notify all entities or if we want the traits themselves to subscribe to updates from the entity. But as an example where we don't have to decide that now (We could have the coordinator listen to all updates from all traits and the notify all entities, if we wanted it to work that way.). Point being from an API perspective its not locking in anything.

roborock/devices/device.py
```diff
@@ -219,8 +225,77 @@ async def close(self) -> None:
...
+        Args:
+            protocol: The data protocol number.
+            data_point: The data value for this protocol.
+        """
+        # Handle status protocol updates
+        if protocol in ROBOROCK_DATA_STATUS_PROTOCOL and self.v1_properties and self.v1_properties.status:
+            # Update the specific field in the status trait
+            match protocol:
+                case RoborockDataProtocol.ERROR_CODE:
+                    self.v1_properties.status.error_code = RoborockErrorCode(data_point)
```
allenporter: I think the decision about which field in the trait to update should live inside the specific trait. I think we should pass in the protocol and data point and let it decide to update or not or which field to update.

Lash-L:
Closing and replaced by https://github.com/Python-roborock/python-roborock/pull/799

PR #734, opened 2026-01-06
chore: Add CONTRIBUTING.md

allenporter:
Document things like uv, pytest, where test data json files are. This can be improved (e.g. explaining how to test, use tooling, etc) but documenting what we have for now.


Lash-L:
Can you add info for activating the venv before pre commit install?

allenporter:
> Can you add info for activating the venv before pre commit install?

Done.

Lash-L (approved):

PR #740, opened 2026-01-08
Add an approach for determining if a dataclass field is supported

allenporter:
This consists of multiple concepts:
- DeviceFeaturesTrait defines a method `is_field_supported` which will tell you if an optional field on a RoborockBase class type is supported or not
- The RoborockBase field defins a metadata attribute for a product schea code which is used to check the field against some other source
- We could substitue in other values in DeviceFeatures or status in the future as needed.

This adds 6 fields that are straight forward to add.

This will be used to help us decouple discovering devices and supported entities from the initial refresh to get data values for a trait.

This is based on attempting to implement the specific proposal from @Lash-L in https://github.com/Python-roborock/python-roborock/pull/722 from Lash-L

### DIFF
--- a/roborock/data/v1/v1_containers.py
+++ b/roborock/data/v1/v1_containers.py
@@ -115,12 +115,16 @@ class StatusField(FieldNameBase):
     DRY_STATUS = "dry_status"
 
 
+def _requires_schema_code(requires_schema_code: str, default=None) -> Any:
+    return field(metadata={"requires_schema_code": requires_schema_code}, default=default)
+
+
 @dataclass
 class Status(RoborockBase):
     msg_ver: int | None = None
     msg_seq: int | None = None
-    state: RoborockStateCode | None = field(metadata={"requires_schema_code": "state"}, default=None)
-    battery: int | None = field(metadata={"requires_schema_code": "battery"}, default=None)
+    state: RoborockStateCode | None = _requires_schema_code("state", default=None)
+    battery: int | None = _requires_schema_code("battery", default=None)
     clean_time: int | None = None
     clean_area: int | None = None
     error_code: RoborockErrorCode | None = None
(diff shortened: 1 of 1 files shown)


roborock/data/v1/v1_containers.py
```diff
@@ -91,12 +92,35 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+class FieldNameBase(str, Enum):
```
Lash-L: I think we should use StrEnum

roborock/data/v1/v1_containers.py
```diff
@@ -91,12 +92,35 @@
...
+    annotated with `requires_schema_code` metadata to map the field to a schema
+    code in the product schema, which may have a different name than the field/attribute name.
+    """
+
+    STATE = "state"
+    BATTERY = "battery"
+    FAN_POWER = "fan_power"
+    WATER_BOX_MODE = "water_box_mode"
+    CHARGE_STATUS = "charge_status"
+    DRY_STATUS = "dry_status"
```
Lash-L, line 115: is this right? I see drying_status in the data

roborock/data/v1/v1_containers.py
```diff
@@ -91,12 +92,35 @@
...
+
+
 @dataclass
 class Status(RoborockBase):
     msg_ver: int | None = None
     msg_seq: int | None = None
-    state: RoborockStateCode | None = None
-    battery: int | None = None
+    state: RoborockStateCode | None = field(metadata={"requires_schema_code": "state"}, default=None)
+    battery: int | None = field(metadata={"requires_schema_code": "battery"}, default=None)
```
Lash-L: So I know this was my suggestion and very similar to how I did it in device features. I wonder if a wrapper around field though that makes it easier to set the metadata would be worthwhile as this gets more complicated.

i.e. OurSubClassField(requires_schema_code=StatusField.BATTERY) or still just "battery" if that makes more sense. May help us avoid making any mistakes with typos in the metadata field name

The init of the field could create the metadata and everything else could stay the same but would likely lower the burden of contributing for other users.

roborock/data/v1/v1_containers.py
allenporter: Changed. TIL.

roborock/data/v1/v1_containers.py
allenporter: Yes. This is the field name `dry_status`  in `Status`.  This enum is used to ask "is `dry_status` supported on this status object?

The schema code is `drying_status` set on the metadata which determines if `dry_status` is supported based on the product schema.

roborock/data/v1/v1_containers.py
allenporter: I did this though made a few alterations, PTAL.

PR #744, opened 2026-01-16
fix: handle different error format for map status

Lash-L:
Relates to #160952

error format seemingly is different for some api/ some firmware. This allows us to handle it better and successfully return a apropriate exception on the map logic.

### DIFF
--- a/roborock/protocols/v1_protocol.py
+++ b/roborock/protocols/v1_protocol.py
@@ -13,7 +13,7 @@ from enum import StrEnum
 from typing import Any, Protocol, TypeVar, overload
 
 from roborock.data import RoborockBase, RRiot
-from roborock.exceptions import RoborockException, RoborockUnsupportedFeature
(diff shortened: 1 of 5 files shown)


roborock/protocols/v1_protocol.py
```diff
@@ -154,26 +154,38 @@ def decode_rpc_response(message: RoborockMessage) -> ResponseMessage:
         ) from e
 
     request_id: int | None = data_point_response.get("id")
-    exc: RoborockException | None = None
+    api_error: RoborockException | None = None
     if error := data_point_response.get("error"):
-        exc = RoborockException(error)
+        api_error = RoborockException(error)
+
     if (result := data_point_response.get("result")) is None:
```
allenporter, line 179: Can we check `code` here and raise RoborockInvalidStatus here in this function?  Then translate that to Device busy above. I think there is a way to do that which is generic here but then specific above withotu passing the payload up to the higher leve.

allenporter (approved):

PR #745, opened 2026-01-16
feat: add clean record for Q7

### LINKED ISSUE #735: Roborock Q7B "Total cleaning time" property value in HA does not match total cleaning time reported by Roborock app
### Steps to reproduce:
1. Add Roborock Q7B to HA 2026.1.0b4
2. Examine Integrations, Roborock, Diagnostic, Total Cleaning Time property and get "92h 34m"
3. Examine Roborock App settings under Cleaning, have value "Total cleaning time" 4733 minutes (78h 53m)

### Cause:
"Total cleaning time" in HA is mapped to Roborock property real_clean_time, but this appears to be the time of the last clean (cleaning_time) in seconds, for example on this robot:

cleaning_time = 92
real_clean_time = 5554 (92.57 minutes)

The Roborock Android app appears to be getting the total cleaning time value from an MQTT method call "service.get_record_list", which returns these values as well as a list of the last 40 cleaning runs:

total_time = 283980 (4733 minutes in seconds)
total_area = 332748 total cleaned area in square cm (/100 for square metres)

These match the values reported by the Roborock app.

Here is an example response from service.get_record_list:
```
{
    "msgId": "1767747635412",
    "code": 0,
    "method": "service.get_record_list",
    "data": {
        "total_time": 283980,
        "total_area": 332748,
        "total_count": 271,
        "record_list": [
            {
                "url": "/userdata/record_map/1749517692_1749518434_1749518432_clean_map.bin",
                "detail": "{\"record_start_time\":1749517692,\"method\":0,\"record_use_time\":660,\"clean_count\":1,\"record_clean_area\":0,\"record_clean_mode\":5,\"record_clean_way\":0,\"record_task_status\":1,\"record_faultcode\":0,\"record_dust_num\":0,\"clean_current_map\":1749518432,\"record_map_url\":\"/userdata/record_map/1749517692_1749518434_1749518432_clean_map.bin\"}"
            }, ... (39 more records)```

### PR DESCRIPTION
**Pull request by Lash-L**
Would appreciate you being pretty in depth on this one as it is the second trait for Q7 so I want to identify anything now rather than later.

fixes #735

### DIFF
--- a/roborock/data/b01_q7/b01_q7_containers.py
+++ b/roborock/data/b01_q7/b01_q7_containers.py
@@ -1,6 +1,9 @@
 import datetime
+import json
 from dataclasses import dataclass, field
+from functools import cached_property
 
+from ...exceptions import RoborockException
 from ..containers import RoborockBase
 from .b01_q7_code_mappings import (
     B01Fault,
@@ -245,6 +248,17 @@ class CleanRecordListItem(RoborockBase):
 
     url: str | None = None
     detail: str | None = None
+
+    @cached_property
+    def detail_parsed(self) -> CleanRecordDetail | None:
+        """Parse and return the detail as a CleanRecordDetail object."""
+        if self.detail is None:
+            return None
+        try:
+            parsed = json.loads(self.detail)
+        except json.JSONDecodeError as ex:
+            raise RoborockException(f"Invalid B01 record detail JSON: {self.detail!r}") from ex
+        return CleanRecordDetail.from_dict(parsed)
 
 
 @dataclass
--- a/roborock/devices/traits/b01/q7/clean_summary.py
+++ b/roborock/devices/traits/b01/q7/clean_summary.py
@@ -6,7 +6,7 @@ and a `record_list` whose items contain a JSON string in `detail`.
 
 from __future__ import annotations
 
-import json
+import logging
 
 from roborock import CleanRecordDetail, CleanRecordList, CleanRecordSummary
 from roborock.devices.rpc.b01_q7_channel import send_decoded_command
@@ -19,6 +19,8 @@ from roborock.roborock_typing import RoborockB01Q7Methods
 __all__ = [
     "CleanSummaryTrait",
 ]
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class CleanSummaryTrait(CleanRecordSummary, Trait):
@@ -59,14 +61,12 @@ class CleanSummaryTrait(CleanRecordSummary, Trait):
         """Return parsed record detail objects (newest-first)."""
         details: list[CleanRecordDetail] = []
         for item in record_list.record_list:
-            if item.detail is None:
-                continue
             try:
-                parsed = json.loads(item.detail)
-            except json.JSONDecodeError as ex:
-                raise RoborockException(f"Invalid B01 record detail JSON: {item.detail!r}") from ex
-            parsed = CleanRecordDetail.from_dict(parsed)
-
+                parsed = item.detail_parsed
+            except RoborockException as ex:
+                # Rather than failing if something goes wrong here, we should fail and log to tell the user.
+                _LOGGER.debug("Failed to parse record detail: %s", ex)
+                continue
             if parsed is not None:
                 details.append(parsed)
 
--- a/tests/devices/traits/b01/q7/test_clean_summary.py
+++ b/tests/devices/traits/b01/q7/test_clean_summary.py
@@ -1,6 +1,7 @@
 """Tests for CleanSummaryTrait class for B01/Q7 devices."""
 
 import json
+import logging
 
 import pytest
 
@@ -156,8 +157,9 @@ async def test_get_clean_record_details_with_none_detail(
 
 async def test_get_clean_record_details_invalid_json(
     clean_summary_trait: CleanSummaryTrait,
+    caplog: pytest.LogCaptureFixture,
 ) -> None:
-    """Test that invalid JSON in detail raises RoborockException."""
+    """Test that invalid JSON in detail is logged and skipped."""
     response_with_invalid_json = {
         "total_time": 34980,
         "total_area": 28540,
@@ -170,7 +172,10 @@ async def test_get_clean_record_details_invalid_json(
         ],
     }
 
-    with pytest.raises(RoborockException, match="Invalid B01 record detail JSON"):
-        await clean_summary_trait._get_clean_record_details(
+    with caplog.at_level(logging.DEBUG):
+        details = await clean_summary_trait._get_clean_record_details(
             record_list=CleanRecordList.from_dict(response_with_invalid_json)
         )
+
+    assert len(details) == 0
+    assert any("Failed to parse record detail" in record.message for record in caplog.records)

### CODE REVIEW

File: roborock/devices/traits/b01/q7/clean_summary.py [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+        record_list = await self.get_record_list()
+
+        self.total_time = record_list.total_time
+        self.total_area = record_list.total_area
+        self.total_count = record_list.total_count
+
+        details = await self.get_clean_record_details(record_list=record_list)
+        self.last_record_detail = details[0] if details else None
+
+    async def get_record_list(self) -> CleanRecordList:
```
allenporter: Can this be private `_get_record_list`?

roborock/devices/traits/b01/q7/clean_summary.py
```diff
@@ -0,0 +1,82 @@
...
+            except json.JSONDecodeError as ex:
+                raise RoborockException(f"Invalid B01 record detail JSON: {detail!r}") from ex
+            if not isinstance(parsed, dict):
+                raise RoborockException(f"Unexpected B01 record detail type: {type(parsed).__name__}: {parsed!r}")
+            return CleanRecordDetail.from_dict(parsed)
+        if isinstance(detail, dict):
+            return CleanRecordDetail.from_dict(detail)
+        raise TypeError(f"Unexpected B01 record detail type: {type(detail).__name__}: {detail!r}")
+
+    async def get_clean_record_details(self, *, record_list: CleanRecordList | None = None) -> list[CleanRecordDetail]:
```
allenporter: Can this be private?

roborock/devices/traits/b01/q7/clean_summary.py
```diff
@@ -0,0 +1,82 @@
...
+        """Return parsed record detail objects (newest-first)."""
+        if record_list is None:
```
allenporter: Why is this optionally passed in? Can it be always passed in?

roborock/data/b01_q7/b01_q7_containers.py
```diff
@@ -203,3 +203,49 @@ def wind_name(self) -> str | None:
...
+    clean_current_map: int | None = None
+    record_map_url: str | None = None
+
+
+@dataclass
+class CleanRecordListItem(RoborockBase):
+    """Represents an entry in the clean record list returned by `service.get_record_list`."""
+
+    url: str | None = None
+    detail: str | dict | None = None
```
allenporter: This really can be either type?

allenporter (changes requested):

roborock/data/b01_q7/b01_q7_containers.py
Lash-L: I originally thought i saw data where it was not a str in the app code, but I looked again and I think it is always a json str.

roborock/devices/traits/b01/q7/clean_summary.py
Lash-L: sorry - some slop. fixed

roborock/data/b01_q7/b01_q7_containers.py
allenporter: So it's always a json dict? So then this should be `dict` (post json parsing)?

Lash-L: it is a dict after we parse from json - but not at the time of accessing/loading from result of api.

it is a dumped json dict within the main json dict, so it is not parsed.

We have to manually parse it before access. We don't update the value either after we parse - i guess we theoretically could?

CleanRecordList.from_dict(result) has detail as a str

then we access .detail -> parse it to a dict -> convert it to CleanRecordDetail

allenporter: I don't think we should use the same field for the string or dict value and expose it in the API. Maybe there could be a manual step to "clean" the data first, converting it from json to dict, then parse it? or something else.

Lash-L: I have a local change i'm trying to finish up - but i changed the typing to be just str. You can let me know if that's what you were thinking

tests/mock_data.py
```diff
@@ -124,8 +124,8 @@
 
 TESTDATA = pathlib.Path("tests/testdata")
 
-PRODUCTS = {file.name: json.load(file.open()) for file in TESTDATA.glob("home_data_product_*.json")}
```
Lash-L, line 127: unrelated change but windows seems to require this or tests crash. 

I have just been doing it and not including it in any PRs

tests/devices/traits/b01/q7/test_clean_summary.py
```diff
@@ -0,0 +1,176 @@
...
+                ),
+            },
+            {
+                "url": "/userdata/record_map/record2.bin",
+                "detail": None,
+            },
+        ],
+    }
+
+    details = await clean_summary_trait._get_clean_record_details(
```
allenporter, line 150: Not sure about testing private methods like this

roborock/data/b01_q7/b01_q7_containers.py
```diff
@@ -222,24 +224,45 @@ class CleanRecordDetail(RoborockBase):
...
+        return None
+
 
 @dataclass
 class CleanRecordListItem(RoborockBase):
     """Represents an entry in the clean record list returned by `service.get_record_list`."""
 
     url: str | None = None
-    detail: str | dict | None = None
+    detail: str | None = None
```
allenporter, line 250: Can you add a comment here explaining what's up with this field? It looks like its not really usable and folks are not expected to use this directly and should only access the summary for the last detail.

As an alternative you could put the json parsing here as something like `detail_parsed` in a `@cachedproperty` if you want to make it accessible and shared with the

allenporter (approved):

tests/devices/traits/b01/q7/test_clean_summary.py
Lash-L: The alternative is I do a whole wrapping around the refresh() call, but this was a bit simplier mock wise. If you want me to change it i will.

roborock/data/b01_q7/b01_q7_containers.py
Lash-L: add a cached property

roborock/devices/traits/b01/q7/clean_summary.py
```diff
@@ -59,14 +61,12 @@ async def _get_clean_record_details(self, *, record_list: CleanRecordList) -> li
...
-                continue
             try:
-                parsed = json.loads(item.detail)
-            except json.JSONDecodeError as ex:
-                raise RoborockException(f"Invalid B01 record detail JSON: {item.detail!r}") from ex
-            parsed = CleanRecordDetail.from_dict(parsed)
-
+                parsed = item.detail_parsed
+            except RoborockException as ex:
+                # Rather than failing if something goes wrong here, we should fail and log to tell the user.
```
allenporter, line 67: Given there is already an exception catch in `detail_parsed` i'd just push this in there? What I was trying to get at by suggesting moving the parsing into the dataclas is that this whole function feels like it can be a one-liner:
```
return reverse([
    item.detail_parsed
    for item in record_list.record_list
    if item.detail_parsed is not None
])
```
While typically i think it is good to not swallow exceptions down deep, i'm just thinking since in practice you're ignoring parse errors then just make it less code.

roborock/devices/traits/b01/q7/clean_summary.py
allenporter, line 60: I don't think this needs to be async

tests/devices/traits/b01/q7/test_clean_summary.py
allenporter: I guess i assume this can be tested with either the response builder for a more end to end test, or with putting all the parsing inside the dataclass object and testing it there.  I would probably just make `last_record_detail` also cacheable and test it entirely inside the dataclass, so everything is encapsulated.

allenporter (approved):
Fine to fix in follow up PRs since its just style

PR #746, opened 2026-01-16
Add StatusField support for DeviceFeatures

allenporter:
Add support for checking if a field is supported based on the device features status. This adds support for clean percent, though we don't currently have a device info that supports it.


Lash-L:
Like the approach - lets wait until https://github.com/Python-roborock/python-roborock/pull/747 is merged in though for device info and then we can use real test.

PR #747, opened 2026-01-16
feat: improve device_info

Lash-L:
- Add some more device_info from my device and I asked rob to do it for his devices as well.
- expanded the cli to give the home data product information as we have had some interest in that
- Add the feature information to the device features as that is not currently in the diagnostic data
- Added support for loading from diagnostics (once users have this new info in their diagnostics)

### DIFF
--- a/roborock/cli.py
+++ b/roborock/cli.py
@@ -44,7 +44,7 @@ from pyshark.packet.packet import Packet  # type: ignore
 from roborock import RoborockCommand
 from roborock.data import RoborockBase, UserData
 from roborock.data.b01_q10.b01_q10_code_mappings import B01_Q10_DP
-from roborock.data.code_mappings import SHORT_MODEL_TO_ENUM, RoborockProductNickname
+from roborock.data.code_mappings import SHORT_MODEL_TO_ENUM
 from roborock.device_features import DeviceFeatures
 from roborock.devices.cache import Cache, CacheData
 from roborock.devices.device import RoborockDevice
@@ -861,7 +861,7 @@ def _parse_diagnostic_file(diagnostic_path: Path) -> dict[str, dict[str, Any]]:
     if not coordinators:
         return all_products_data
 
-    for coordinator_id, coordinator_data in coordinators.items():
+    for coordinator_data in coordinators.values():
         device_data = coordinator_data.get("device", {})
         product_data = coordinator_data.get("product", {})
 
@@ -870,11 +870,11 @@ def _parse_diagnostic_file(diagnostic_path: Path) -> dict[str, dict[str, Any]]:
             continue
         # Derive product nickname from model
         short_model = model.split(".")[-1]
-        product_nickname = SHORT_MODEL_TO_ENUM.get(short_model, RoborockProductNickname.PEARLPLUS)
+        product_nickname = SHORT_MODEL_TO_ENUM.get(short_model)
 
         current_product_data: dict[str, Any] = {
-            "Protocol Version": device_data.get("pv"),
-            "Product Nickname": product_nickname.name,
+            "protocol_version": device_data.get("pv"),
+            "product_nickname": product_nickname.name if product_nickname else "Unknown",
         }
 
         # Get feature info from the device_features trait (preferred location)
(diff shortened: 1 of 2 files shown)


roborock/device_features.py
```diff
@@ -571,9 +576,17 @@ def from_feature_flags(
         # RobotNewFeatures = new_feature_info
         # newFeatureInfoStr = new_feature_info_str
         # feature_info =robotFeatures
-        kwargs: dict[str, Any] = {}
+        kwargs: dict[str, Any] = {
+            # Store raw feature info for diagnostics
```
allenporter, line 580: Thank you for this!!!

roborock/cli.py
```diff
@@ -834,59 +842,219 @@ def on_package(packet: Packet):
...
 
 
+def _parse_diagnostic_file(diagnostic_path: Path) -> dict[str, dict[str, Any]]:
+    """Parse device info from a Home Assistant diagnostic file.
+
+    Args:
+        diagnostic_path: Path to the diagnostic JSON file.
+
+    Returns:
+        A dictionary mapping model names to device info dictionaries.
```
allenporter, line 852: It could be helpful to define a dataclass for this (in future)

allenporter: I think we may want to move to names like "protocol_version" instead of "Protocol Version"

roborock/cli.py
```diff
@@ -834,59 +842,219 @@ def on_package(packet: Packet):
...
+        product_data = coordinator_data.get("product", {})
+
+        model = product_data.get("model")
+        if not model:
+            continue
+
+        # Skip duplicate models
+        if model in all_products_data:
+            continue
+
```
allenporter: just collapse
```suggestion
        if not model or model in all_products_data:
            continue
```

roborock/cli.py
```diff
@@ -834,59 +842,219 @@ def on_package(packet: Packet):
...
+        # Derive product nickname from model
+        short_model = model.split(".")[-1]
+        product_nickname = SHORT_MODEL_TO_ENUM.get(short_model, RoborockProductNickname.PEARLPLUS)
+
+        current_product_data: dict[str, Any] = {
+            "Protocol Version": device_data.get("pv"),
+            "Product Nickname": product_nickname.name,
```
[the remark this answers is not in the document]
allenporter: i didnt review this entire change but i do think defaulting model name could lead to trouble

roborock/cli.py
```diff
@@ -834,59 +842,219 @@ def on_package(packet: Packet):
...
+        click.echo(f"Parsing diagnostic file: {diagnostic_path.resolve()}")
+        all_products_data = _parse_diagnostic_file(diagnostic_path)
+
+        if not all_products_data:
+            click.echo("No device data found in diagnostic file.")
+            return
+
+        click.echo(f"Found {len(all_products_data)} device(s) in diagnostic file.")
+
+        # Filter out already recorded models if recording
```
allenporter: i think it's ok to overwrite existing products. Then user can diff and decide if its got more/better info? and revert if not.

roborock/cli.py
```diff
@@ -834,59 +842,219 @@ def on_package(packet: Packet):
...
+            all_products_data = {
+                model: data for model, data in all_products_data.items() if model not in existing_device_info
+            }
+            if not all_products_data:
+                click.echo("No new device info to record (all models already exist).")
+                return
 
-    click.echo(f"Found {len(devices)} devices. Fetching data...")
+    else:
+        # Original behavior: connect to devices
```
allenporter: "Original behavior" is not a helpful comment given the code as is, just more like a record of the conversation between the user and AI. I'd say remove.

allenporter (changes requested):

roborock/cli.py
Lash-L: guilty...

roborock/cli.py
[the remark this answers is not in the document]
Lash-L: I have set it to be unknown for now so it is a bit clearer if there is an issue.

roborock/device_features.py
Lash-L: Hopefully a good way for us to get a lot of testing data!

roborock/cli.py
Lash-L: Done!

allenporter (approved):

PR #765, opened 2026-02-04
feat: add stronger mop washing support

Lash-L:
This adds the ability to set the wash towel mode (aka mop cleaning mode), start a clean and stop a clean.

The logic for grabbing out valid options mirrors this: https://github.com/Python-roborock/python-roborock/pull/611


roborock/devices/traits/v1/__init__.py
```diff
@@ -246,7 +245,10 @@ async def discover_features(self) -> None:
                 _LOGGER.debug("Trait '%s' not supported, skipping", item.name)
                 continue
             _LOGGER.debug("Trait '%s' is supported, initializing", item.name)
-            trait = item_type()
+            if item_type is WashTowelModeTrait:
```
allenporter: How about initializing this outside the loop before entering? This would be similar to how it works in the constructor where you first create the types that have special argument, then create the rest that haven't been set. This code will ignore it if its set already.

roborock/devices/traits/v1/wash_towel_mode.py
```diff
@@ -11,3 +14,30 @@ class WashTowelModeTrait(WashTowelMode, common.V1TraitMixin):
 
     command = RoborockCommand.GET_WASH_TOWEL_MODE
     requires_dock_type = is_wash_n_fill_dock
+
+    def __init__(
+        self,
+        device_feature_trait: DeviceFeaturesTrait | None = None,
```
allenporter: My understanding is this will always be set so it won't ever be None. Is that righT? (Can simplify a check below)

tests/devices/traits/v1/test_wash_towel_mode.py
```diff
@@ -65,3 +68,133 @@ async def test_unsupported_wash_towel_mode(
...
+)
+@pytest.mark.parametrize(
+    ("wash_mode"),
+    [
+        (WashTowelModes.SMART),
+        (WashTowelModes.LIGHT),
+    ],
+)
+async def test_set_wash_towel_mode(
+    wash_towel_mode: WashTowelModeTrait | None,
```
allenporter: I think this doesn't need to be `| None` here and below

tests/devices/traits/v1/test_wash_towel_mode.py
```diff
@@ -65,3 +68,133 @@ async def test_unsupported_wash_towel_mode(
...
+async def test_wash_towel_mode_options(
+    wash_towel_mode: WashTowelModeTrait | None,
+    dock_type_code: RoborockDockTypeCode,
+    is_super_deep_wash_supported: bool,
+    is_dirty_replenish_clean_supported: bool,
+    expected_modes: list[WashTowelModes],
+) -> None:
+    """Test what modes are available based on device features."""
+    assert wash_towel_mode is not None
+    # We need to clear the cached property to ensure it re-reads the features
```
allenporter: Is this really required? My impression is we're testing with a new instance of `WashTowelModeTrait` for every test and there are new arguments to the trait each time. If that is not the case then it would also be a problem in production code?

allenporter (changes requested):

roborock/devices/traits/v1/wash_towel_mode.py
Lash-L: I thought you were right but I think the actual flow is:

1. from_dict creates a temporary new instance (with device_feature_trait=None)                                                                                                      2. _update_trait_values copies only dataclass fields from that temp instance to self

roborock/devices/traits/v1/wash_towel_mode.py
```diff
@@ -11,3 +14,28 @@ class WashTowelModeTrait(WashTowelMode, common.V1TraitMixin):
 
     command = RoborockCommand.GET_WASH_TOWEL_MODE
     requires_dock_type = is_wash_n_fill_dock
+
+    def __init__(
+        self,
+        device_feature_trait: DeviceFeaturesTrait,
+        wash_mode: WashTowelModes | None = None,
```
allenporter: i think this is not used

allenporter (approved):

roborock/devices/traits/v1/wash_towel_mode.py
Lash-L: removed this and reworked some stuff so it's more similar to the status trait. by having it's own parse response we can avoid the |None check on device_feature_trait as well

Issue #767, opened 2026-02-14
Add Missing Q10 Feature support


lboue:
This issue is an omnibus of all the missing features from Q10 devices.

## Higher Priority

- [ ] Map (multi-map, restricted zones, virtual walls, room management)
- [ ] Cleaning history
- [x] Battery
- [x] Cleaning mode
- [x] Fan level

## Lower Priority
- [x] Basic vacuum control (start, pause, resume, stop, dock)
- [x] Empty dustbin
- [x] Consumable status (main brush, side brush, filter, rag life)
- [ ] Consumable Resets
- [ ] Mop state
- [ ] Carpet Cleaning Settings (carpet clean prefer, auto boost, carpet clean type)
- [ ] Floor Material / Ground Clean settings
- [ ] Robot voice volume
- [ ] DnD (Do Not Disturb)
- [ ] Valley Point Charging (Off peak)
- [ ] Child lock
- [ ] Button lights
- [ ] Remote Control
- [ ] Timers
- [ ] Unit settings (area unit, time zone)
- [ ] Clean line preference
- [ ] Self-identifying carpet
- [ ] Obstacle avoidance settings

TangerineNightmare:
for what it is worth there is a new open source project at https://codeberg.org/zcodeberg/roborock-q10-utils-gpl that implements a subset of the Q10v+ mobile app functionality as a self-hostable python web interface. It includes some API information that may or may not be of use?

https://codeberg.org/zcodeberg/roborock-q10-utils-gpl/src/branch/main/Q10_API_REFERENCE.md

hCoureau:
## Q10 map and cleaning contribution stack

Tracking the current contribution sequence here so each change can remain small and independently reviewable:

- [x] #932 — stable live-map vector calibration
- [ ] #933 — safe read-only current-map request
- [ ] [Archived map packet/parser API](https://github.com/hCoureau/python-roborock/tree/feat/q10-archive-packets) — prepared and fully validated; submit after #933
- [ ] [Saved-map and clean-history archive traits](https://github.com/hCoureau/python-roborock/tree/feat/q10-map-archives) — prepared and fully validated; submit after the parser API
- [ ] [Obstacle parsing and rendering](https://github.com/hCoureau/python-roborock/tree/feat/q10-obstacles) — prepared and fully validated; submit after archive traits
[...]

PR #769, opened 2026-02-14
feat: Add initial Q10 support for Status Trait

allenporter:
For now there is a central refresh mechanism for all traits which can be revisited in the future. The properties API listens for updates and fans them out to traits, which listen for apporopriate DPS values they are responsible for. This adds a way to map dps values to dataclass fields, and reuses the exisitng Roborock base data for parsing updates. Comments have been added to the code base to describe how additional traits can be added following this pattern.

This is modeled after the ideas we explored in #709 and #692 but has been even further simplified. 

issue #767

### DIFF
--- a/roborock/devices/rpc/b01_q10_channel.py
+++ b/roborock/devices/rpc/b01_q10_channel.py
@@ -28,7 +28,7 @@ async def stream_decoded_responses(
             decoded_dps = decode_rpc_response(response_message)
         except RoborockException as ex:
             _LOGGER.debug(
-                "Failed to decode B01 RPC response: %s: %s",
+                "Failed to decode B01 Q10 RPC response: %s: %s",
                 response_message,
                 ex,
             )
--- a/roborock/devices/traits/b01/q10/__init__.py
+++ b/roborock/devices/traits/b01/q10/__init__.py
@@ -56,11 +56,8 @@ class Q10PropertiesApi(Trait):
 
     async def refresh(self) -> None:
         """Refresh all traits."""
-        # Ask for updates to specific DPS values. Updates will be received
-        # by the subscribe loop below.
-        # For now we just ask for all DPS values that traits care about here
-        # but this could be split out to give each trait its own refresh
-        # method in the future if needed.
+        # Sending the REQUEST_DPS will cause the device to send all DPS values
+        # to the device. Updates will be received by the subscribe loop below.
         await self.command.send(B01_Q10_DP.REQUEST_DPS, params={})
 
     async def _subscribe_loop(self) -> None:


roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -1,53 +1,102 @@
...
 
 
+@dataclass
 class dpSelfIdentifyingCarpet(RoborockBase):
     op: str
     result: int
     data: str
 
 
+@dataclass
```
Lash-L, line 52: Not needed here - but I wonder if there is a way we can avoid having to do dataclass. Pydantic base models don't, but we do which is odd

roborock/devices/rpc/b01_q10_channel.py
```diff
@@ -3,18 +3,39 @@
...
+    mqtt_channel: MqttChannel,
+) -> AsyncGenerator[dict[B01_Q10_DP, Any], None]:
+    """Stream decoded DPS messages received via MQTT."""
+
+    async for response_message in mqtt_channel.subscribe_stream():
+        try:
+            decoded_dps = decode_rpc_response(response_message)
+        except RoborockException as ex:
+            _LOGGER.debug(
+                "Failed to decode B01 RPC response: %s: %s",
```
Lash-L: Let's specify Q10 here

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -1,29 +1,77 @@
...
+                pass  # ignore cancellation errors
+            self._subscribe_task = None
+
+    async def refresh(self) -> None:
+        """Refresh all traits."""
+        # Ask for updates to specific DPS values. Updates will be received
+        # by the subscribe loop below.
+        # For now we just ask for all DPS values that traits care about here
+        # but this could be split out to give each trait its own refresh
+        # method in the future if needed.
```
Lash-L: This comment is a little confusing. The ask for updates to specific seems to be in contract with the next few lines.

roborock/devices/traits/b01/q10/common.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, dps_type_map: dict[B01_Q10_DP, type], dps_field_map: dict[B01_Q10_DP, str]):
+        """Initialize the converter for a specific RoborockBase-derived class."""
+        self._dps_type_map = dps_type_map
+        self._dps_field_map = dps_field_map
+
+    @classmethod
+    def from_dataclass(cls, dataclass_type: type[RoborockBase]):
+        """Initialize the converter for a specific RoborockBase-derived class."""
+        dps_type_map: dict[B01_Q10_DP, type] = {}
```
Lash-L, line 60: Nothing wrong here - just stylistically(the naming), B01_Q10_DP feels off. I think it was me who originally added it, but maybe we should change as a follow up

roborock/devices/rpc/b01_q10_channel.py
allenporter: Done.

roborock/devices/traits/b01/q10/__init__.py
allenporter: Re-worded. I was trying to say we may add additional DPS values below, but maybe that wont be the case. Removed for now.

roborock/devices/traits/b01/q10/common.py
allenporter: I think just like `DataPoint` could work given its in a module, but yeah, happy to address in a followup.

Lash-L:
Lgtm - thanks great approach

PR #774, opened 2026-02-28
Add Q7 map_content support and room segment cleaning helper

arduano:
## Summary
- add `Q7PropertiesApi.clean_segments(room_ids)` helper for write-dependent room cleaning
- add B01/Q7 map decode pipeline and SCMap parse/render support
- add Q7 `map_content` trait for map image bytes
- update map fetch flow to read current `map_id` via `service.get_map_list` and fetch payload via `service.upload_by_mapid` (`map_id` param)

## Why
This provides a clean, typed foundation for both:
- read-only Q7 telemetry/map retrieval, and
- write-dependent segment cleaning,
without relying on HA-side decode logic.

### Unit tests (repo-local `.venv`, nix-shell Python 3.12)
- `pytest -q tests/map/test_b01_map_parser.py tests/devices/traits/b01/q7/test_init.py tests/devices/traits/b01/q7/test_clean_summary.py`
- result: `27 passed`

### Read-only live checks against Q7 (no cleaning/write actions)
- queried telemetry props successfully (`status`, `wind`, `cleaning_time`)
- fetched map payload and rendered PNG via `map_content.refresh()`
- verified PNG bytes and magic header (`\x89PNG...`)

### Lint
- `ruff check` on modified parser/trait/tests passed

### DIFF
--- a/roborock/devices/rpc/b01_q7_channel.py
+++ b/roborock/devices/rpc/b01_q7_channel.py
@@ -127,7 +127,8 @@ async def send_map_command(mqtt_channel: MqttChannel, request_message: Q7Request
                 return
 
             if response_message.protocol == RoborockMessageProtocol.MAP_RESPONSE and response_message.payload:
-                future.set_result(response_message.payload)
+                if not future.done():
+                    future.set_result(response_message.payload)
                 return
 
             try:
@@ -146,12 +147,16 @@ async def send_map_command(mqtt_channel: MqttChannel, request_message: Q7Request
                     continue
                 code = inner.get("code", 0)
                 if code != 0:
-                    future.set_exception(RoborockException(f"B01 command failed with code {code} ({request_message})"))
+                    if not future.done():
+                        future.set_exception(
+                            RoborockException(f"B01 command failed with code {code} ({request_message})")
+                        )
                     return
                 data = inner.get("data")
                 if isinstance(data, dict) and isinstance(data.get("payload"), str):
                     try:
-                        future.set_result(bytes.fromhex(data["payload"]))
+                        if not future.done():
+                            future.set_result(bytes.fromhex(data["payload"]))
                     except ValueError:
                         pass
 


arduano:
@Lash-L what do you mean by "honey Roborock code", do you mean `Homey`?

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -87,6 +87,17 @@ async def start_clean(self) -> None:
             },
         )
 
+    async def clean_segments(self, room_ids: list[int]) -> None:
+        """Start segment/room cleaning for the given room ids."""
+        await self.send(
```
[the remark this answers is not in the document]
arduano: Openclaw: Good catch. I aligned naming/docs: clean_segments now takes segment_ids, while protocol payload still maps to Q7 room_ids internally.

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -87,6 +87,17 @@ async def start_clean(self) -> None:
...
+            command=RoborockB01Q7Methods.SET_ROOM_CLEAN,
+            params={
+                "clean_type": CleanTaskTypeMapping.ROOM.code,
+                "ctrl_value": SCDeviceCleanParam.START.code,
+                "room_ids": room_ids,
+            },
+        )
```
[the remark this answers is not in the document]
arduano: Openclaw: Added coverage in tests/devices/traits/b01/q7/test_init.py (test_q7_api_clean_segments) asserting RPC method and payload for clean_segments([...]).

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,230 @@
...
+
+import base64
+import hashlib
+import io
+import zlib
+from dataclasses import dataclass
+
+from Crypto.Cipher import AES
+from Crypto.Util.Padding import pad, unpad
+from PIL import Image
```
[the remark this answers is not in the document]
arduano: Openclaw: Addressed. Pillow is now an explicit runtime dependency in pyproject.toml and uv.lock.

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -33,11 +35,13 @@ class Q7PropertiesApi(Trait):
...
     clean_summary: CleanSummaryTrait
     """Trait for clean records / clean summary (Q7 `service.get_record_list`)."""
+    map_content: Q7MapContentTrait
 
-    def __init__(self, channel: MqttChannel) -> None:
+    def __init__(self, channel: MqttChannel, *, local_key: str, serial: str, model: str) -> None:
         """Initialize the B01Props API."""
         self._channel = channel
         self.clean_summary = CleanSummaryTrait(channel)
+        self.map_content = Q7MapContentTrait(channel, local_key=local_key, serial=serial, model=model)
```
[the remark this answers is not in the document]
arduano: Openclaw: Addressed for backward compatibility. Q7PropertiesApi.__init__/create now accept optional local_key/serial/model; map support is enabled only when provided. Added a constructor-compatibility test for direct callers.

roborock/map/__init__.py
```diff
@@ -1,7 +1,13 @@
 """Module for Roborock map related data classes."""
```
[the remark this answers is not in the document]
arduano: Openclaw: Updated the module docstring to reflect broader map utilities (data classes plus parsing/rendering helpers).

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +99,52 @@ def find_response(response_message: RoborockMessage) -> None:
...
+    roborock_message = encode_mqtt_payload(request_message)
+    future: asyncio.Future[bytes] = asyncio.get_running_loop().create_future()
+
+    def find_response(response_message: RoborockMessage) -> None:
+        if future.done():
+            return
+
+        if response_message.protocol == RoborockMessageProtocol.MAP_RESPONSE and response_message.payload:
+            future.set_result(response_message.payload)
+            return
```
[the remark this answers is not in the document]
arduano: Openclaw: Addressed by serializing B01 map commands per channel (lock) and keeping msgId-aware response handling to prevent cross-wired payloads under overlap.

arduano:
Openclaw: Thanks for the review. I addressed the `future.done()` style concern in `send_map_command` by guarding all completion paths (`set_result`/`set_exception`) and pushed commit `3a0b944` to this PR branch. Also re-ran targeted Q7 map tests after the change (`4 passed`).

arduano:
I just connected it to my home assistant (manually build the hass container with this new PR as a dependency) and I can now view the map and send the vacuum to clean specific segments.

The only thing I'm unsure about is map feature rendering. Am I supposed to make this PR also render things like the vacuum location into the map png, or is that handled in hass? Right now it just sends the plain map

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +109,62 @@ def find_response(response_message: RoborockMessage) -> None:
...
+
+
+async def send_map_command(mqtt_channel: MqttChannel, request_message: Q7RequestMessage) -> bytes:
+    """Send map upload command and wait for MAP_RESPONSE payload bytes.
+
+    Map requests are serialized per channel so concurrent map calls cannot
+    cross-wire responses between callers.
+    """
+
+    async with _get_map_command_lock(mqtt_channel):
```
allenporter, line 121: Please move the `asyncio.Lock` to the trait, which is one per device. We don't need to do this indirectly via a map to the channel as we can just do it there before calling into this map function.

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +109,62 @@ def find_response(response_message: RoborockMessage) -> None:
...
+        def find_response(response_message: RoborockMessage) -> None:
+            if future.done():
+                return
+
+            if response_message.protocol == RoborockMessageProtocol.MAP_RESPONSE and response_message.payload:
+                if not future.done():
+                    future.set_result(response_message.payload)
+                return
+
+            try:
```
allenporter, line 134: Does this happen in practice that we get a normal protocol response back or is this speculative? I'm wondering if this could have a high level comment describing the scenarios in which we observe this happening. Second, i think this is complex enough that likely we'd want to reuse the common parts across both protocols, in a way that is easy to read.

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def _maybe_b64(data: bytes) -> bytes | None:
+    try:
+        return base64.b64decode(data, validate=False)
+    except Exception:
+        return None
+
+
+def decode_b01_map_payload(raw_payload: bytes, *, local_key: str, serial: str, model: str) -> bytes:
```
allenporter, line 176: Please review https://github.com/Python-roborock/python-roborock/blob/df84567fe7d5fc02409185a8e71fd3996093139e/docs/DEVICES.md#layer-responsibilities

Problem: This code is combining map decoding with protocol decoding and these responsibilities should be split. e.g. review how the v1 code separates the protocol encoding/decoding from map encoding. e.g. https://github.com/Python-roborock/python-roborock/blob/df84567fe7d5fc02409185a8e71fd3996093139e/roborock/protocols/v1_protocol.py#L220
or `create_v1_channel`.

The idea is we shouldn't need to be passing around local keys everywhere. Make a separate rpc channel object responsible for decoding the bytes. Then the map trait can get decoded bytes and pass them to a map content parser.

You can also review how existing encryption code is structured in protocol.py and protocols/

tests/fixtures/b01/raw-mqtt-map301.bin.inflated.bin
allenporter, line 1: Lets put this in `tests/map/testdata/`

roborock/devices/traits/b01/q7/map_content.py
```diff
@@ -0,0 +1,80 @@
+"""Map content trait for B01/Q7 devices."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+from roborock.devices.rpc.b01_q7_channel import send_decoded_command, send_map_command
+from roborock.devices.traits import Trait
+from roborock.devices.traits.v1.map_content import MapContent
```
allenporter, line 10: Let's not directly depend on the traits of another device type. For now, create a new object with the fields we need here rather than reusing this and we can decide to reuse in the future.  The protocols are different enough that I don't think we'll have calling code or anything that will treat these map objects the same.

roborock/devices/traits/b01/q7/map_content.py
```diff
@@ -0,0 +1,80 @@
...
+        super().__init__()
+        self._channel = channel
+        self._local_key = local_key
+        self._serial = serial
+        self._model = model
+
+    async def refresh(self) -> B01MapContent:
+        map_list_response = await send_decoded_command(
+            self._channel,
+            Q7RequestMessage(dps=10000, command=RoborockB01Q7Methods.GET_MAP_LIST, params={}),
```
allenporter, line 55: Can we extract a named constant for `10000`?

roborock/devices/traits/b01/q7/map_content.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, channel: MqttChannel, *, local_key: str, serial: str, model: str) -> None:
+        super().__init__()
+        self._channel = channel
+        self._local_key = local_key
+        self._serial = serial
+        self._model = model
+
+    async def refresh(self) -> B01MapContent:
+        map_list_response = await send_decoded_command(
```
allenporter, line 53: In v1 we have a separate trait for holding on to the map list and current map `MapsTrait`. I assume we'll want the other content here also? 

We decided to separate them in v1 just because there was a lot of existing code to unwind and introduced the `Home` object to help with caching map content across different maps, but not sure we need that here yet.

Can you review the existing MapsTrait / HomeTrait / MapContentTrait for v1 and consider what would be a reasonable first step to take with that in mind? I do worry that getting the `current_map` logic wrong is harder to fix later, but we can also do something simple for now. Basically i think it'd be good to at least keep the map list somewhere.

arduano:
@Lash-L before I do a deeper review myself, is this the kind of PR splitting you'd like?
1. https://github.com/arduano/python-roborock/pull/1
2. https://github.com/arduano/python-roborock/pull/2
3. https://github.com/arduano/python-roborock/pull/3

PR #777, opened 2026-03-06
fix: correct mop intensity code for slider

Lash-L:
Some vacs have a slider instead of the normal selection. This should fix the problem.

https://github.com/home-assistant/core/issues/164831

### DIFF
--- a/roborock/data/v1/v1_clean_modes.py
+++ b/roborock/data/v1/v1_clean_modes.py
@@ -68,6 +68,17 @@ class WashTowelModes(RoborockModeEnum):
     SUPER_DEEP = ("super_deep", 8)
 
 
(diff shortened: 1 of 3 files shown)


allenporter (approved):

BlackRockCity:
Follow-up from direct Saros 10R (`roborock.vacuum.a144`) hardware testing: the `WATER_SLIDE_MODE_MAPPING` introduced in this PR does not correctly correspond to the official Roborock water-flow slider.

I have documented the protocol captures and direct testing in #801.

All seven representative values have now been tested against the robot and verified in the official iOS app:

221 Slight
225 Low
230 Gentle
235 Medium
240 Moderate
245 High
250 Extreme

In particular, the current mapping's 245 -> moderate is actually High on the Saros 10R, and 248 -> high falls in the official Extreme range.

PR #778, opened 2026-03-07
q7/b01: command-layer segment clean + map payload retrieval helpers (split 1/3)

arduano:
Openclaw: This PR is **1/3** of a planned split of the original Q7/B01 map+segment work, to keep review size and scope tight.

## What this PR does (high level)
This first slice adds the **command-layer plumbing** needed to:
- start **room/segment cleaning** on Q7-class B01 devices, and
- reliably fetch the **current map payload bytes** (raw) for downstream decoding/parsing in follow-up PRs.

No SCMap parsing or map rendering is included here — those come in later splits.

### 1) Add `send_map_command()` helper for B01/Q7 map uploads
`roborock/devices/rpc/b01_q7_channel.py`

Openclaw: Introduces `send_map_command(mqtt_channel, request_message) -> bytes` which publishes a B01/Q7 request and waits for the corresponding map payload.

Current behavior in this PR:
- Waits for dedicated `MAP_RESPONSE` messages with raw `payload` bytes.
- Uses RPC decode only for fail-fast non-zero `code` responses correlated to the initiating `msgId` and requested `dps`.
- Returns clear timeout/error exceptions.

Note on evidence:
- Real-device sniffing on Q7T+ (`roborock.vacuum.sc05`, B01) observed map payload delivery via `MAP_RESPONSE` raw bytes.
- Did **not** observe RPC-wrapped `data.payload` map bytes.

### 2) Extend Q7 traits with segment cleaning + map retrieval
Files:
- `roborock/devices/traits/b01/q7/__init__.py`
- `roborock/devices/traits/b01/q7/map.py`

Openclaw: Adds Q7 segment-clean helper and introduces a dedicated map trait:
- `clean_segments(segment_ids: list[int])` sends `service.set_room_clean` with `ROOM` clean type (Q7 uses room IDs).
- New `MapTrait` holds map-related commands/state.
- `Q7MapList` / `Q7MapListEntry` dataclasses (RoborockBase) model map-list response.
- Map-list caching on trait (`get_map_list()` cache + `refresh_map_list()` refresh).
- `get_current_map_id()`, `get_map_payload(map_id=...)`, `get_current_map_payload()` live on the map trait.

Concurrency/safety:
- Map uploads are serialized per-device via an `asyncio.Lock` to avoid response cross-wiring when multiple map requests are in-flight.

### 3) Tests
`tests/devices/traits/b01/q7/test_init.py`

Openclaw: Adds coverage for:
- `clean_segments()` request shape
- current-map payload retrieval via `GET_MAP_LIST` + `UPLOAD_BY_MAPID`
- fallback-to-first-map behavior
- clear error when map list is unusable
- map-list dataclass + cache behavior on the new map trait

## Follow-ups (separate PRs)
Openclaw: Subsequent split PRs will build on this by:
- decoding/inflating the raw map payload into SCMap bytes,
- parsing SCMap for rooms/segments metadata, and
- wiring a `map_content` trait that exposes `rooms` and a PNG render.


roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +99,83 @@ def find_response(response_message: RoborockMessage) -> None:
...
     finally:
         unsub()
+
+
+async def send_map_command(mqtt_channel: MqttChannel, request_message: Q7RequestMessage) -> bytes:
+    """Send map upload command and wait for a map payload as bytes.
+
+    In practice, B01 map responses may arrive either as:
+    - a dedicated ``MAP_RESPONSE`` message with raw payload bytes, or
+    - a regular decoded RPC response that wraps a hex payload in ``data.payload``.
```
allenporter: How positive are you this this the case? Are both raw payloads the same in those cases? or is this halucinated?

Is there a specific `dps` int value this is sent with? or it can be any? (i notice the dps key is ignored entirely)

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +99,83 @@ def find_response(response_message: RoborockMessage) -> None:
...
+
+    roborock_message = encode_mqtt_payload(request_message)
+    future: asyncio.Future[bytes] = asyncio.get_running_loop().create_future()
+    publish_started = asyncio.Event()
+
+    def find_response(response_message: RoborockMessage) -> None:
+        if future.done():
+            return
+
+        # Avoid accepting queued/stale MAP_RESPONSE messages before we actually
```
allenporter: It this actually needed? I can't tell if this is needed because the code below doesn't work at determining which method was sent to it, or because its over optimized for performance. (if the latter, i don't think its worth it)

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -123,11 +140,61 @@ async def find_me(self) -> None:
             params={},
         )
 
+    async def get_map_list(self) -> dict[str, Any] | None:
```
allenporter: Can we please use a dataclass / roborock base dataclass to hold this response?

allenporter: Map related commands seem like they should be on a separate trait, similar to how the v1 API looks. You can see the clean summary as a nexample.

That will also let us keep state on the trait about the latest map list, so you don't need to ref-etch the map list to get the current map id. it can be cached on the trait.

roborock/devices/rpc/b01_q7_channel.py
arduano: This was openclaw picking up on a copilot review comment, should I remove it? I can't really tell either

roborock/devices/traits/b01/q7/__init__.py
arduano: Yeah sorry I'm not familiar enough with python to have caught this one, I come from typescript
Will adjust

arduano: AI updated it for me now but again I'm unsure about best practices with traits

roborock/devices/rpc/b01_q7_channel.py
arduano: AI did calls to the real API just then to sanity check and decided that yeah it's hallucinated. MAP_RESPONSE is the only case

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +99,60 @@ def find_response(response_message: RoborockMessage) -> None:
...
         unsub()
+
+
+async def send_map_command(mqtt_channel: MqttChannel, request_message: Q7RequestMessage) -> bytes:
+    """Send map upload command and wait for the map payload bytes.
+
+    On a real Q7 B01 device, map uploads arrive as a dedicated
+    ``MAP_RESPONSE`` message with raw payload bytes.
+
+    We still decode RPC responses so we can fail fast on non-zero ``code``
```
allenporter: What causes this to happen? It seems speculative, and is a lot of code.

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,99 @@
...
+class MapTrait(Trait):
+    """Map retrieval + map metadata helpers for Q7 devices."""
+
+    def __init__(self, channel: MqttChannel) -> None:
+        self._channel = channel
+        # Map uploads are serialized per-device to avoid response cross-wiring.
+        self._map_command_lock = asyncio.Lock()
+        self._map_list_cache: Q7MapList | None = None
+
+    async def refresh_map_list(self) -> Q7MapList:
```
allenporter: to elaborate from my examples: The idea for these traits is they have a single `refresh` method that refreshes the content. You can see the v1 device has separate traits for getting the map list vs caching the map content.

roborock/devices/rpc/b01_q7_channel.py
allenporter: openclaw, please remove it :)

roborock/devices/traits/b01/q7/__init__.py
allenporter: yeah we kind of just made up a pattern.  

OpenClaw, think about it form the callers perspective:

(1) You want an easy way to refresh all the content
https://github.com/home-assistant/core/blob/a3d8d766782e075ee8d23e4328666ca1e15807bd/homeassistant/components/roborock/coordinator.py#L338
(2) Anytime you want to get the current value you can just get it directly from the trait without any other i/o e.g. https://github.com/home-assistant/core/blob/a3d8d766782e075ee8d23e4328666ca1e15807bd/homeassistant/components/roborock/sensor.py#L96


You can see these for some examples of traits in the code base:

https://github.com/Python-roborock/python-roborock/blob/main/roborock/devices/traits/v1/home.py
https://github.com/Python-roborock/python-roborock/blob/main/roborock/devices/traits/v1/map_content.py

roborock/devices/traits/b01/q7/map.py
allenporter: `refresh`?

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,99 @@
...
+
+    async def get_current_map_id(self, *, refresh: bool = False) -> int:
+        """Resolve and return the currently active map id from map list metadata."""
+        map_list = await self.get_map_list(refresh=refresh)
+        map_id = self._extract_current_map_id(map_list)
+        if map_id is None:
+            raise RoborockException(f"Unable to determine map_id from map list response: {map_list!r}")
+        return map_id
+
+    async def get_map_payload(self, *, map_id: int) -> bytes:
```
allenporter: I don't think this needs to be exposed publicly so prefix with a `_`

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -99,3 +99,30 @@ def find_response(response_message: RoborockMessage) -> None:
         raise
     finally:
         unsub()
+
+
+async def send_map_command(mqtt_channel: MqttChannel, request_message: Q7RequestMessage) -> bytes:
```
Lash-L, line 129: Why can't this use the send_decoded_command() in b01_q7_channel?

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -87,6 +97,17 @@ async def start_clean(self) -> None:
             },
         )
 
+    async def clean_segments(self, segment_ids: list[int]) -> None:
```
Lash-L, line 99: I think this can live on the map trait. 

that way the interaction is .map.clean_segments(), but i'm not sold. Thoughts @allenporter ?

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,101 @@
...
+
+from roborock.data import RoborockBase
+from roborock.devices.rpc.b01_q7_channel import send_decoded_command, send_map_command
+from roborock.devices.traits import Trait
+from roborock.devices.transport.mqtt_channel import MqttChannel
+from roborock.exceptions import RoborockException
+from roborock.protocols.b01_q7_protocol import Q7RequestMessage
+from roborock.roborock_typing import RoborockB01Q7Methods
+
+_Q7_DPS = 10000
```
Lash-L: I think we don't need to define this constant multiple times if it is always the same

roborock/devices/traits/b01/q7/__init__.py
allenporter: It seems like a "vacuum" command since its asking it to clean? But i get the point that segments are part of a map.

arduano:
Openclaw: Pushed a follow-up review pass addressing the latest code-shape feedback.

What changed:

1) Shared B01 command send path
- Refactored `b01_q7_channel.py` so `send_decoded_command()` and `send_map_command()` now share a private `_send_command()` helper for the common subscribe/publish/wait/timeout flow.
- Kept `send_map_command()` separate because the map upload still resolves from raw `MAP_RESPONSE` bytes rather than decoded RPC `data`.

2) Removed repeated Q7 DPS literals
- Added `B01_Q7_DPS = 10000` in the protocol module and reused it across the Q7 traits / helpers instead of redefining or hardcoding the same value in multiple places.

3) Small cleanup pass
- Simplified the map helper to compare against `B01_VERSION` directly rather than re-encoding the request only to read `.version`.
- Cleaned up `map.py` formatting/imports.

Validation:
- `ruff check` on the touched files passes
- `pytest -q tests/devices/traits/b01/q7 tests/devices/rpc` passes locally (42 passed)

I left `clean_segments()` where it is for now since that part still looked like an API preference discussion rather than a settled request. If you want it moved onto the map trait as well, happy to do that in a follow-up.

arduano:
hey hey, I got openclaw (gpt-5.4 this time) to think about this from a "matainer's perspective" and address the comments, plus anything else it sees. First commit addresses the commends, second commit was a sanity check pass

If it ever gets too sloppy, I'm happy for you to tell me to properly deal with the code myself, I just prefer not to deal with python repos unless I have to

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -5,19 +5,60 @@
...
+    request_message: Q7RequestMessage,
+    *,
+    response_matcher: Callable[[RoborockMessage], _T | None],
+    timeout_error: str,
+) -> _T:
+    """Publish a B01 command and resolve on the first matching response."""
+    roborock_message = encode_mqtt_payload(request_message)
+    future: asyncio.Future[_T] = asyncio.get_running_loop().create_future()
+
+    def find_response(response_message: RoborockMessage) -> None:
```
allenporter: I think the return type is not correct here. Should be `_T | None`?

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -26,10 +67,8 @@ async def send_decoded_command(
 ) -> dict[str, Any] | None:
```
allenporter: I believe this will never return `None` now since the response matcher will never return None.

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -60,28 +99,21 @@ def find_response(response_message: RoborockMessage) -> None:
...
     try:
-        await mqtt_channel.publish(roborock_message)
-        return await asyncio.wait_for(future, timeout=_TIMEOUT)
-    except TimeoutError as ex:
-        raise RoborockException(f"B01 command timed out after {_TIMEOUT}s ({request_message})") from ex
+        return await _send_command(
+            mqtt_channel,
+            request_message,
+            response_matcher=find_response,
+            timeout_error=f"B01 command timed out after {_TIMEOUT}s ({request_message})",
```
allenporter: Lets not pass in a string here for exception handling. If we want to change the exception message based on the context, then make it so the timeout exception is caught here (and the map command caller) and it can re-throw with a different error message.   Or don't do it if its not needed.

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,99 @@
...
+            Q7RequestMessage(dps=B01_Q7_DPS, command=RoborockB01Q7Methods.GET_MAP_LIST, params={}),
+        )
+        if not isinstance(response, dict):
+            raise TypeError(f"Unexpected response type for GET_MAP_LIST: {type(response).__name__}: {response!r}")
+
+        parsed = Q7MapList.from_dict(response)
+        if parsed is None:
+            raise TypeError(f"Failed to decode map list response: {response!r}")
+
+        self._map_list = parsed
```
allenporter: I think callers won't necessarily except a TypeError here so we could rewrite as this:

```suggestion
        if (parsed := Q7MapList.from_dict(response)) is None:
            raise RoborockException(f"Failed to decode map list response: {response!r}")
        self._map_list = parsed
```

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@dataclass
+class Q7MapList(RoborockBase):
+    """Map list response returned by `service.get_map_list`."""
+
+    map_list: list[Q7MapListEntry] = field(default_factory=list)
+
+
+class MapTrait(Trait):
```
allenporter: Consider making this subclass `Q7MapList` so we can remove the extra level of indirection of `map_list` which is similar to how the v1 traits work.

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,99 @@
...
+            return None
+
+        for entry in map_list:
+            if entry.cur and isinstance(entry.id, int):
+                return entry.id
+
+        first = map_list[0]
+        if isinstance(first.id, int):
+            return first.id
+        return None
```
allenporter: I think we can simplify this to work like this:

```suggestion
  @staticmethod
  def _extract_current_map_id(map_list_response: Q7MapList) -> int | None:
      """Extract current map id, defaulting to the first map if none marked current."""
      if not map_list_response.map_list:
          return None
      # Sort so that cur=True comes first, then return the first id found
      ordered = sorted(map_list_response.map_list, key=lambda x: x.cur or False, reverse=True)
      return next(iter(ordered)).id
```

I would also say add `current_map_id` as a property on `Q7MapList` would also be a nice improvement, rather than having it in the trait.

tests/devices/traits/b01/q7/test_init.py
```diff
@@ -257,3 +257,108 @@ async def test_q7_api_find_me(q7_api: Q7PropertiesApi, fake_channel: FakeChannel
...
+    payload_data = json.loads(unpad(message.payload, AES.block_size))
+    assert payload_data["dps"]["10000"]["method"] == "service.set_room_clean"
+    assert payload_data["dps"]["10000"]["params"] == {
+        "clean_type": CleanTaskTypeMapping.ROOM.code,
+        "ctrl_value": SCDeviceCleanParam.START.code,
+        "room_ids": [10, 11],
+    }
+
+
+async def test_q7_api_get_current_map_payload(
```
allenporter: These can live in `test_map.py` since the are primarily for the map trait.

arduano:
Openclaw: pushed another follow-up pass addressing the latest review comments.

What changed in this update:

1) B01 command helper cleanup
- Kept the shared `_send_command()` helper focused on subscribe/publish/wait only.
- Moved timeout-specific exception text back to `send_decoded_command()` / `send_map_command()` instead of threading a timeout string through the shared helper.
- Tightened `send_decoded_command()` so the returned decoded response is typed as a dict rather than `dict | None`.

2) Map trait / map-list cleanup
- Switched map-list decode failures in `MapTrait.refresh()` from `TypeError` to `RoborockException`.
- Made `MapTrait` inherit from `Q7MapList` directly so the cached `map_list` data lives on the trait itself.
- Added `Q7MapList.current_map_id` and used that instead of keeping the current-map extraction helper on the trait.

3) Test organization
- Moved the map-trait-focused tests out of `test_init.py` into a new `test_map.py`.

I left `clean_segments()` where it was for now since that thread still read to me as an open API-shape preference discussion rather than a settled requested change.

Validation:
- `pytest -q tests/devices/traits/b01/q7 tests/devices/rpc` -> 43 passed

Commit:
- `8303a8c` — `🦎 q7: align B01 map helpers with maintainer feedback`

arduano:
Just woke up, got it to address it. The commit looked reasonable

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,87 @@
...
+
+@dataclass
+class Q7MapListEntry(RoborockBase):
+    """Single map list entry returned by `service.get_map_list`."""
+
+    id: int | None = None
+    cur: bool | None = None
+
+
+@dataclass
```
Lash-L: Let's move this into the data folder for b01 q7

roborock/devices/traits/b01/q7/map.py
```diff
@@ -0,0 +1,87 @@
...
+from roborock.devices.rpc.b01_q7_channel import send_decoded_command, send_map_command
+from roborock.devices.traits import Trait
+from roborock.devices.transport.mqtt_channel import MqttChannel
+from roborock.exceptions import RoborockException
+from roborock.protocols.b01_q7_protocol import B01_Q7_DPS, Q7RequestMessage
+from roborock.roborock_typing import RoborockB01Q7Methods
+
+
+@dataclass
+class Q7MapListEntry(RoborockBase):
```
Lash-L: Same here, data package for b01 q7

Lash-L (approved):

roborock/devices/rpc/b01_q7_channel.py
```diff
@@ -5,31 +5,67 @@
...
+        return await asyncio.wait_for(future, timeout=_TIMEOUT)
+    finally:
+        unsub()
 
 
 async def send_decoded_command(
     mqtt_channel: MqttChannel,
     request_message: Q7RequestMessage,
-) -> dict[str, Any] | None:
+) -> Any:
```
allenporter, line 64: Can we get more specific here? I believe we know find_response returns a dict so this can too.

arduano: Got AI to address it https://github.com/Python-roborock/python-roborock/pull/786

PR #779, opened 2026-03-07
fix: don't reconnect on no active subscribers

Lash-L:
Relates to: https://github.com/home-assistant/core/issues/164865

So I think what is happening is:

We send a message on the mqtt topic when we start up (i.e. get_network_info), we then don't send any more messages, we unsubscribe. Then, we are connected to the mqtt server with no active topics. Roborock ends up booting us off the server as we are not subscribers. Then, we self-heal and reconnect. The same thing happens until we reconnect and disconnect enough that we get rate limited and the user gets told to reauth.

I'm not sure this is a long term solution, but I don't think it is a bad solution. I think we might always want to be subscribed (https://github.com/Python-roborock/python-roborock/pull/731 would relate). But, I think it is fine safe keeping to not reconnect if for any reason we don't have subscribers. I think in the future it is likely we wont hit this check, but i think it is still a fine check to have?

### DIFF
--- a/roborock/mqtt/roborock_session.py
+++ b/roborock/mqtt/roborock_session.py
@@ -56,9 +56,12 @@ class RoborockMqttSession(MqttSession):
     The client is run as a background task that will run until shutdown. Once
     connected, the client will wait for messages to be received in a loop. If
     the connection is lost, the client will be re-created and reconnected. There
-    is backoff to avoid spamming the broker with connection attempts. The client
-    will automatically re-establish any subscriptions when the connection is
-    re-established.
+    is backoff to avoid spamming the broker with connection attempts.
+
+    Reconnect attempts are deferred while there are no active subscriptions,
+    which avoids unnecessary reconnect churn for idle sessions. Reconnects
+    resume as soon as a subscription is added again. The client automatically
+    re-establishes any existing subscriptions when the connection returns.
     """
 
     def __init__(
--- a/tests/mqtt/test_roborock_session.py
+++ b/tests/mqtt/test_roborock_session.py
@@ -539,27 +539,23 @@ async def test_session_unauthorized_after_start(
         await session.close()
 
 
-async def test_session_defers_reconnect_when_idle() -> None:
+async def test_session_defers_reconnect_when_idle(
+    mock_aenter_client: AsyncMock,
+    message_iterator: FakeAsyncIterator,
+    mqtt_client_lite: AsyncMock,
+) -> None:
     """Test that reconnects are deferred when there are no active subscriptions."""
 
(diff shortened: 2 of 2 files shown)


tests/mqtt/test_roborock_session.py
```diff
@@ -529,8 +529,37 @@ def succeed_then_fail_unauthorized() -> Any:
...
+    connect_attempts = 0
+
+    async def fake_run_connection(start: asyncio.Future[None] | None) -> None:
+        nonlocal connect_attempts
+        connect_attempts += 1
+        if start and not start.done():
+            start.set_result(None)
+
+    with patch.object(session, "_run_connection", side_effect=fake_run_connection):
+        reconnect_task = asyncio.create_task(session._run_reconnect_loop(start_future))
```
allenporter: Can we patch the aiomqtt client instead of our internal objects? The other tests patch `roborock.mqtt.roborock_session.aiomqtt.Client.__aenter__`

tests/mqtt/test_roborock_session.py
```diff
@@ -529,8 +529,37 @@ def succeed_then_fail_unauthorized() -> Any:
...
+        try:
+            await start_future
+            await asyncio.sleep(0.1)
+            assert connect_attempts == 1
+            assert session._diagnostics.as_dict().get("reconnect_deferred", 0) >= 1
```
allenporter: Can you get this from `params` instead of peeking into internal members?

roborock/mqtt/roborock_session.py
allenporter, line 56: Can you update this section to describe the expected behavior?

roborock/mqtt/roborock_session.py
```diff
@@ -175,6 +175,16 @@ async def _run_reconnect_loop(self, start_future: asyncio.Future[None] | None) -
             if self._stop:
                 _LOGGER.debug("MQTT session closed, stopping retry loop")
                 return
+            if not self._client_subscribed_topics and not self._listeners.keys():
```
allenporter, line 181: I think this implementation looks good.  One thought is this behavior could live in `LazyMqttSession`. I can see that on one hand (a) it could be easier to make a more efficient "wake up" method in the lazy session if the logic is simpler (e.g. using an `asyncio.Event` to do wake up when a new subscriber is added) but (b) might make it more complex to have an "API" between the two classes exposed for this. Just sharing a thought but not need to act on it.

Lash-L: Probably a good idea, I don't think i'll act on it right now, but we can see how things end up looking with future work with the mqtt connection and how we handle the subscriptions

PR #780, opened 2026-03-07
feat: use get_rooms to limit issues with missing room names

Lash-L:
Another one I don't know for sure if this was the root of the problem but is hopefully enough to fix the problem.

The app seems to rely a lot more on the get_rooms endpoint than home data rooms, it's possible some vacs are using it very limitedly. So what we do is this:

If we get a room back with an iot id we have not seen before/ is unknown, we call get_rooms(), try to connect it, and update the home_info as well. 

If the room is unknown but we have already seen it, then we do not call get_rooms to not overwhelm the web api.

Relates to: https://github.com/home-assistant/core/issues/165006

### DIFF
--- a/roborock/devices/traits/v1/home.py
+++ b/roborock/devices/traits/v1/home.py
@@ -120,22 +120,23 @@ class HomeTrait(RoborockBase, common.V1TraitMixin):
 
         rooms: dict[int, NamedRoomMapping] = {}
         if map_info.rooms:
-            # Not all vacuums resopnd with rooms inside map_info.
+            # Not all vacuums respond with rooms inside map_info.
+            # If we can determine if all vacuums will return everything with get_rooms, we could remove this step.
             for room in map_info.rooms:
                 if room.id is not None and room.iot_name_id is not None:
                     rooms[room.id] = NamedRoomMapping(
                         segment_id=room.id,
                         iot_id=room.iot_name_id,
-                        name=room.iot_name or "Unknown",
+                        name=room.iot_name or f"Room {room.id}",
                     )
 
         # Add rooms from rooms_trait.
-        # Keep existing names from map_info unless they are unknown.
+        # Keep existing names from map_info unless they are fallback names.
         if self._rooms_trait.rooms:
             for room in self._rooms_trait.rooms:
                 if room.segment_id is not None and room.name:
                     existing_room = rooms.get(room.segment_id)
-                    if existing_room is None or existing_room.name == "Unknown":
+                    if existing_room is None or existing_room.name == f"Room {room.segment_id}":
                         rooms[room.segment_id] = room
 
         return CombinedMapInfo(
--- a/tests/devices/traits/v1/test_home.py
+++ b/tests/devices/traits/v1/test_home.py
@@ -633,6 +633,7 @@ async def test_refresh_map_info_prefers_map_info_names_and_adds_missing_rooms(
     This test verifies:
     1. Existing map_info room names are preserved.
     2. Missing rooms from RoomsTrait are added.
+    3. map_info fallback names are replaced by RoomsTrait names.
     """
     map_info = MultiMapsListMapInfo(
         map_flag=0,
@@ -648,6 +649,11 @@ async def test_refresh_map_info_prefers_map_info_names_and_adds_missing_rooms(
                 iot_name_id="2362042",
                 iot_name="Bedroom from map_info",
             ),
+            MultiMapsListRoom(
+                id=20,
+                iot_name_id="9999001",
+                iot_name=None,
+            ),
         ],
     )
 
@@ -656,6 +662,7 @@ async def test_refresh_map_info_prefers_map_info_names_and_adds_missing_rooms(
     # - segment_id 19 with valid name: exists in map_info, should NOT override
     # - segment_id 17 with fallback name: not in map_info, should be added
     # - segment_id 18 with valid name: not in map_info, should be added
+    # - segment_id 20 with valid name: overrides map_info fallback "Room 20"
     rooms_trait.rooms = [
         NamedRoomMapping(segment_id=16, iot_id="2362048", name="Room 16"),  # Exists in map_info, should not override
         NamedRoomMapping(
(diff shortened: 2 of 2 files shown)


roborock/devices/traits/v1/rooms.py
```diff
@@ -32,10 +33,12 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
 
     command = RoborockCommand.GET_ROOM_MAPPING
 
-    def __init__(self, home_data: HomeData) -> None:
+    def __init__(self, home_data: HomeData, web_api: UserWebApiClient | None = None) -> None:
```
allenporter: i believe web api is always set.  it this a test hack?

roborock/devices/traits/v1/home.py
```diff
@@ -138,6 +138,8 @@ async def _refresh_map_info(self, map_info) -> CombinedMapInfo:
                         # or if the room name isn't unknown.
                         rooms[room.segment_id] = room
 
+        await self._rooms_trait.resolve_unknown_room_names(rooms)
```
allenporter: Can this be handled by `_rooms_trait.refresh()`? It can notice that it sees rooms that are not found and decide what to do about it directly, rather than later trying to reverse engineer what rooms were unseen that need to be re-resolved.

I would just fix it up front and change how `_iot_id_room_name_map` works so that we check for unknown rooms, then populate the missing rooms in the home data, then create the named room map once based on what rooms we have.

roborock/devices/traits/v1/rooms.py
Lash-L: Yes - removed

roborock/devices/traits/v1/home.py
Lash-L: Yeah moved things around. I kept some in home, as we need to do the room renaming of unknowns there. (i.e. Room 3 instead of unknown)

roborock/devices/traits/v1/home.py
```diff
@@ -138,6 +138,14 @@ async def _refresh_map_info(self, map_info) -> CombinedMapInfo:
                         # or if the room name isn't unknown.
                         rooms[room.segment_id] = room
 
+        for segment_id, room in rooms.items():
+            if room.name == "Unknown":
```
allenporter: Can RoomsTrait also handle this? It seems like it has enough information to be able to create the rooms with the correct names from the start.

roborock/devices/traits/v1/rooms.py
allenporter: Something like this?

```suggestion
                NamedRoomMapping(segment_id=segment_id, iot_id=iot_id, name=name_map.get(iot_id, f"Room {segment_id}"))
```

roborock/devices/traits/v1/rooms.py
```diff
@@ -32,10 +33,25 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
...
+        self._web_api = web_api
+        self._seen_unknown_room_iot_ids: set[str] = set()
+
+    async def refresh(self) -> None:
+        """Refresh room mappings and backfill unknown room names from the web API."""
+        response = await self.rpc_channel.send_command(self.command)
+        if not isinstance(response, list):
+            raise ValueError(f"Unexpected RoomsTrait response format: {response!r}")
+
+        segment_pairs = _extract_segment_pairs(response)
```
allenporter: OK now that its all in one file, there is a lot of back and forth between these methods, and mutating of state, multiple checks for the default name, and a few of these methods and properties called a few different times.

I think he flow is:
- get the list of present iot ids in the response (and segments)
- determine if any are missing in the room map
- fetch the new room map
- then we merge the room maps (can we just overwrite given we just got the latest rooms rather than merge?)
- build the namedRoomMapping (again extracting the present iot ids and segments in the response

I think what would simplify things is to make `_extract_segment_pairs` return a dict instead of a tuple, then this function can do something like this:

```
segment_map = _extract_segment_map(response)    # Get list of present iot ids  
if (set(segment_map.values()) - set(_iot_id_room_name_map.keys()):  # Check missing
    self._home_data.rooms = await self._refresh_rooms()  # Overwrite, updates _iot_id_room_name_map
new_data = self._parse_response(response, segment_map)   # Build new map
self._update_trait_values(new_data)
```

Would a flow like that work? I think it would have all the side effects visible here. (This sill has a hidden interaction with _iot_id_room_name_map inside parse_response but maybe ok)

allenporter (changes requested):

roborock/devices/traits/v1/rooms.py
Lash-L: I think that makes sense from my understanding of the apis now.

I keep a 'Unknown' check even though i don't think it will hit as I'm not 100% sure how the cache will handle things.

Check it out let me know your thoughts, appreciate you being thorough on this, hopefully we can get it right and stop having this bug.

allenporter (approved):
i'm still not sure why home trait needs to change at all since its just hacking on room trait values? I don't think this really addressed the feedback about code structure at all in terms of not separating home traits vs room trait responsibilities and avoiding side effects within functions of room traits.  Feel free to look closer later at the feedback i guess?

Lash-L:
> i'm still not sure why home trait needs to change at all since its just hacking on room trait values? I don't think this really addressed the feedback about code structure at all in terms of not separating home traits vs room trait responsibilities and avoiding side effects within functions of room traits. Feel free to look closer later at the feedback i guess?

Marking as a draft for now and will take another look in the morning. I may have been misunderstanding things so I'll give everything a good read. But HomeTrait's change is pretty simple, it's mainly being changed because we have two different 'unknown'/fallback map names now. Unknown(I think this should only be from cache now) or Room {#} and we need to check both.

But for the larger code structure comments, I'll take a bigger look and dive into it some more tomorrow. I don't think it's worth merging in right now if you think there are still some structural issues, so I'm going to hold off until I get a better grasp of your comments.

tests/devices/traits/v1/test_home.py
```diff
@@ -690,3 +697,143 @@ async def test_refresh_map_info_room_override_and_addition_logic(
...
+        ],
+    )
+
+    # Web API returns fresh room names
+    mock_web_api.get_rooms.return_value = [
+        HomeDataRoom(id=9999801, name="Living Room"),
+        HomeDataRoom(id=9999802, name="Kitchen"),
+    ]
+
+    result = await home_trait._refresh_map_info(map_info)
```
allenporter: Either move to room trait tests or use public home apis?

allenporter (approved):

PR #781, opened 2026-03-08
feat: Improve room naming and data integration 

allenporter:
Simplify how we track unset names by adding a `raw_name` field on `NamedRoomMapping`.

This also adds other helper methods to the containers to simplify the code in the traits. The traits are changed to have fewer functions with side effects that mutate data on the trait, and removing duplicate checks for empty lists or value types.

### DIFF
--- a/roborock/devices/traits/v1/home.py
+++ b/roborock/devices/traits/v1/home.py
@@ -222,10 +222,10 @@ class HomeTrait(RoborockBase, common.V1TraitMixin):
         return self._home_map_info.get(current_map_flag)
 
     @property
-    def current_rooms(self) -> list[NamedRoomMapping] | None:
+    def current_rooms(self) -> list[NamedRoomMapping]:
         """Returns the room names for the current map."""
         if self.current_map_data is None:
-            return None
+            return []
         return self.current_map_data.rooms
 
     @property


roborock/devices/traits/v1/home.py
```diff
@@ -114,35 +115,29 @@ async def discover_home(self) -> None:
...
-        if self._rooms_trait.rooms:
-            for room in self._rooms_trait.rooms:
-                if room.segment_id is not None and room.name:
-                    existing_room = rooms.get(room.segment_id)
-                    if existing_room is None or existing_room.name == f"Room {room.segment_id}":
-                        rooms[room.segment_id] = room
-
+        # We have room names from two sources. The map_info.rooms which we just
+        # received from the MultiMapsList or the self._rooms_trait.rooms which
+        # comes from the GET_ROOM_MAPPING command. We merge them, giving priority
```
Lash-L: I think worth noting in this comment that only some devices return it in get_room_mapping

roborock/devices/traits/v1/home.py
```diff
@@ -114,35 +115,29 @@ async def discover_home(self) -> None:
...
+        # to map_info rooms, and excluding unset rooms from the trait.
+        rooms: list[NamedRoomMapping] = list(
```
Lash-L: This is a pretty big transformation, I think it may make sense to split out.

I think just moving the dict to a different variable first would be enough to make it a bit more readable as it would lose the indent.

roborock/devices/traits/v1/home.py
```diff
@@ -114,35 +115,29 @@ async def discover_home(self) -> None:
...
+            {
+                **map_info.rooms_map,
+                **{
+                    room.segment_id: room
+                    for room in self._rooms_trait.rooms or ()
```
Lash-L: Could it be worthwhile to set .rooms to a property? I'm figuring we do this other places as well, may be easier to just know we can always assume it is a list. Not needed for this PR though, just a thought

roborock/devices/traits/v1/home.py
```diff
@@ -114,35 +115,29 @@ async def discover_home(self) -> None:
...
+                    if (existing := map_info.rooms_map.get(room.segment_id)) is None
+                    or (room.raw_name and existing.raw_name is None)
```
Lash-L: Do you think we should just always update existing even if raw_name is not None?

Theoretically, the rooms trait will always be less stale (I think) as it is responsible for calling get_room_mapping or get_rooms(). Do you get that same impression?

roborock/devices/traits/v1/home.py
allenporter: Would it be current_rooms?

roborock/devices/traits/v1/home.py
allenporter: Done.

roborock/devices/traits/v1/home.py
allenporter: Done.

roborock/devices/traits/v1/home.py
allenporter: Done.

roborock/devices/traits/v1/home.py
Lash-L: I think that makes sense!

Lash-L (dismissed):

roborock/devices/traits/v1/home.py
allenporter: This depends on one of the map infos returned from an RPC before it is saved so I don't think this works. I still added the attribute, but it won't make this code simpler.

roborock/devices/traits/v1/home.py
```diff
@@ -232,6 +221,13 @@ def current_map_data(self) -> CombinedMapInfo | None:
             return None
         return self._home_map_info.get(current_map_flag)
 
+    @property
+    def current_rooms(self) -> list[NamedRoomMapping] | None:
+        """Returns the room names for the current map."""
+        if self.current_map_data is None:
+            return None
```
Lash-L: Couldn't we return () or [] here and simplify typing?

Lash-L (approved):

PR #783, opened 2026-03-09
Separate trait response handling logic from refresh logic and merge

allenporter:
This separates the trait data response converting from the refresh and value updating. This follows a pattern in Q10, and seems like it may make it easier to support listeners.


allenporter (commented):
Fixed lint.

Lash-L (approved):

PR #785, opened 2026-03-15
q7: add map_content trait for B01 current map

arduano:
## Summary

Add Q7/B01 `map_content` support on top of the already-merged raw map transport work.

This PR:
- adds a Q7 `map_content` trait with the same basic refresh/cached-fields pattern used by v1
- wires B01/Q7 trait creation through `device_manager` with access to the device/product metadata needed for map decoding
- adds a minimal internal B01 map parser that:
  - decodes the raw `MAP_RESPONSE` payload
  - parses the inflated SCMap payload with the protobuf runtime
  - renders a PNG image
  - stores observed room names in `map_data.additional_parameters["room_names"]`
- adds parser + trait tests and fixture data

## Design intent

The goal here is to keep the new work aligned with the existing codebase shape rather than inventing a separate parser/integration framework:
- transport stays in the existing Q7 `map` trait
- parsed map access lives in a dedicated `map_content` trait
- `refresh()` performs I/O and populates cached fields
- parsing internals stay private/minimal
- the runtime decode path is narrowed to the live-observed payload format

## Validation

- full `pytest -q` → `398 passed, 1 warning`
- full `ruff check .` → clean
- focused parser + trait tests rerun after the final parser/comment/lint fixes:
  - `tests/map/test_b01_map_parser.py`
  - `tests/devices/traits/b01/q7/test_map_content.py`
  - result: `7 passed`
- live read-only sanity check on a real Q7T+ account/device:
  - device model: `roborock.vacuum.sc05`
  - map fetch/decode/render succeeded
  - rendered PNG bytes: `17457`
  - raw payload bytes: `35584`
  - room names discovered: `10..19` / `room1..room10`

## Notes

- The inner SCMap blob uses protobuf-style message encoding; this PR now uses the protobuf runtime with a small local runtime descriptor for the subset of fields consumed here.
- Repo-wide mypy is not currently clean on `main`; this branch does not add new parser-specific mypy failures, but I did not include unrelated mypy cleanups in this PR.

### DIFF
--- a/roborock/map/b01_map_parser.py
+++ b/roborock/map/b01_map_parser.py
@@ -18,7 +18,6 @@ import zlib
 from dataclasses import dataclass
 
 from Crypto.Cipher import AES
-from Crypto.Util.Padding import pad, unpad
 from google.protobuf.message import DecodeError, Message
 from PIL import Image
 from vacuum_map_parser_base.config.image_config import ImageConfig
@@ -26,6 +25,7 @@ from vacuum_map_parser_base.map_data import ImageData, MapData
 
 from roborock.exceptions import RoborockException
 from roborock.map.proto.b01_scmap_pb2 import RobotMap  # type: ignore[attr-defined]
+from roborock.protocol import Utils
 
 from .map_parser import ParsedMapData
 
@@ -83,7 +83,7 @@ def _derive_map_key(serial: str, model: str) -> bytes:
     model_suffix = model.split(".")[-1]
     model_key = (model_suffix + "0" * 16)[:16].encode()
     material = f"{serial}+{model_suffix}+{serial}".encode()
-    encrypted = AES.new(model_key, AES.MODE_ECB).encrypt(pad(material, AES.block_size))
+    encrypted = Utils.encrypt_ecb(material, model_key)
     md5 = hashlib.md5(base64.b64encode(encrypted), usedforsecurity=False).hexdigest()
     return md5[8:24].encode()
 
@@ -99,15 +99,16 @@ def _decode_base64_payload(raw_payload: bytes) -> bytes:
 
 def _decode_b01_map_payload(raw_payload: bytes, *, serial: str, model: str) -> bytes:
     """Decode raw B01 `MAP_RESPONSE` payload into inflated SCMap bytes."""
+    # TODO: Move this lower-level B01 transport decode under `roborock.protocols`
+    # so this module only handles SCMap parsing/rendering.
     encrypted_payload = _decode_base64_payload(raw_payload)
     if len(encrypted_payload) % AES.block_size != 0:
         raise RoborockException("Unexpected encrypted B01 map payload length")
 
     map_key = _derive_map_key(serial, model)
-    decrypted_hex = AES.new(map_key, AES.MODE_ECB).decrypt(encrypted_payload)
 
     try:
-        compressed_hex = unpad(decrypted_hex, AES.block_size).decode("ascii")
+        compressed_hex = Utils.decrypt_ecb(encrypted_payload, map_key).decode("ascii")
         compressed_payload = bytes.fromhex(compressed_hex)
         return zlib.decompress(compressed_payload)
     except (ValueError, UnicodeDecodeError, zlib.error) as err:


tests/map/debug_b01_scmap.py
```diff
@@ -0,0 +1,304 @@
+"""Developer helper for inspecting B01/Q7 SCMap payloads.
```
arduano: Happy to remove this file if you think it's unnecessary, but it may help future investigations

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,277 @@
+"""Module for parsing B01/Q7 map content.
+
+Observed Q7 MAP_RESPONSE payloads are:
+- base64-encoded ASCII
+- AES-ECB encrypted with the derived map key
+- PKCS7 padded
+- ASCII hex for a zlib-compressed SCMap payload
+
+The inner SCMap blob appears to use protobuf wire encoding, but we do not have
+an upstream `.proto` schema for it. We intentionally keep a tiny schema-free
```
arduano: I did explicitly drill the AI on whether we can simplify it further with protobufs, but it thinks this approach is cleaner. If you wish, I can still get it to try reverse-engineering the protobuf schema.

roborock/devices/traits/b01/q7/map_content.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@dataclass
+class MapContent(RoborockBase):
+    """Dataclass representing map content."""
+
+    image_content: bytes | None = None
+    """The rendered image of the map in PNG format."""
+
+    map_data: MapData | None = None
```
Lash-L, line 32: This can hold the map image on it right? Or is it not easy to do that with this approach? 

Still reading through code so I may answer my own question

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -1,10 +1,14 @@
 """Traits for Q7 B01 devices.
-Potentially other devices may fall into this category in the future."""
+
+Potentially other devices may fall into this category in the future.
```
allenporter: Not really, this directory is only for q7

arduano:
ah I wasn't sure from our last conversation if you explicitly preferred a pure code approach, must've misunderstood 

I'll adjust today

allenporter:
> ah I wasn't sure from our last conversation if you explicitly preferred a pure code approach, must've misunderstood
> 
> I'll adjust today

yeah i think i said something vague over chat -- but seeing that there is a very large amount of code manually constructing the messages, it seems preferred to do something with less code.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,10 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-exclude: "CHANGELOG.md"
+exclude: >
```
arduano, line 3: AI claims regex is the only way, that the python ecosystem has no way of doing lists here

.pre-commit-config.yaml
```diff
@@ -34,15 +38,17 @@ repos:
         rev: v0.13.2
         hooks:
             -   id: ruff-format
+                exclude: ^roborock/map/proto/.*_pb2\.py$
```
arduano: regex here too. Pretty ugly yes, not sure if the ecosystem has a better approach, I'm not familiar enough

.pre-commit-config.yaml
Lash-L: That seems to be correct per what I see as well

.pre-commit-config.yaml
Lash-L: Here you should be able to do the exclude within the pyproject.till under ruff settings- I think that is ideal if possible

Lash-L: Looks like you already do it for ruff linting

roborock/devices/traits/b01/q7/map_content.py
```diff
@@ -0,0 +1,105 @@
...
+        self.raw_api_response = parsed.raw_api_response
+
+    def parse_map_content(self, response: bytes) -> MapContent:
+        """Parse map content from raw bytes.
+
+        This mirrors the v1 trait behavior so cached map payload bytes can be
+        reparsed without going back to the device.
+        """
+        if not self._serial or not self._model:
+            raise RoborockException(
```
Lash-L: So it seems like we always need self._serial and self.model.

Does it make sense to make them not optional in init?

roborock/map/proto/b01_scmap.proto
```diff
@@ -0,0 +1,72 @@
+// Source of truth for the B01/Q7 SCMap schema.
+//
+// Regenerate the checked-in Python module after edits with:
+//   python -m grpc_tools.protoc -I. --python_out=. roborock/map/proto/b01_scmap.proto
```
Lash-L: Comment seems not needed as it seems to reference a file we don't have commited.

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,326 @@
...
+
+    try:
+        compressed_hex = unpad(decrypted_hex, AES.block_size).decode("ascii")
+        compressed_payload = bytes.fromhex(compressed_hex)
+        return zlib.decompress(compressed_payload)
+    except (ValueError, UnicodeDecodeError, zlib.error) as err:
+        raise RoborockException("Failed to decode B01 map payload") from err
+
+
+def _parse_proto(blob: bytes, message: Any, *, context: str) -> None:
```
Lash-L: I see a lot of Any typing - can we avoid that if possible?

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,326 @@
...
+        room_name_post=_parse_sc_point(parsed.roomNamePost) if parsed.HasField("roomNamePost") else None,
+        color_id=parsed.colorId if parsed.HasField("colorId") else None,
+        floor_direction=parsed.floor_direction if parsed.HasField("floor_direction") else None,
+        global_seq=parsed.global_seq if parsed.HasField("global_seq") else None,
+    )
+
+
+def _parse_scmap_payload(payload: bytes) -> _ScMapPayload:
+    """Parse inflated SCMap bytes into typed map metadata."""
+    parsed: Any = getattr(b01_scmap_pb2, "RobotMap")()
```
Lash-L: I guess this is kind of the starting place. Is there a way we can type this? Again not familiar with protobuf.

ideally, we can avoid all of the HasFields and such that we have to do if possible

roborock/devices/traits/b01/q7/map_content.py
```diff
@@ -0,0 +1,100 @@
...
+
+This intentionally mirrors the v1 `MapContentTrait` contract:
+- `refresh()` performs I/O and populates cached fields
+- `parse_map_content()` reparses cached raw bytes without I/O
+- fields `image_content`, `map_data`, and `raw_api_response` are then readable
+
+For B01/Q7 devices, the underlying raw map payload is retrieved via `MapTrait`.
+"""
+
+from __future__ import annotations
```
allenporter: Remove since this will mess with types and is no longer needed for modern code.

roborock/map/proto/b01_scmap.proto
```diff
@@ -0,0 +1,70 @@
+// Checked-in B01/Q7 SCMap schema for the generated runtime protobuf module.
+// Regenerate the checked-in Python module after edits with:
+//   python -m grpc_tools.protoc -I./roborock/map/proto --python_out=./roborock/map/proto roborock/map/proto/b01_scmap.proto
```
allenporter, line 3: We're getting closer! My impression though is that you can output the _pb.py files as described in https://protobuf.dev/reference/python/python-generated/ so you get much more than just a descriptor.

Since you have `DevicePointInfo` defined in .proto you should be able to get fully generated python code for parsing that message and not need `_ScPointMessage`.

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,330 @@
...
+)
+
+from .map_parser import ParsedMapData
+
+_B64_CHARS = set(b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=")
+_MAP_FILE_FORMAT = "PNG"
+
+
+@dataclass(frozen=True)
+class _ScPoint:
```
allenporter: are these still needing to move to generated messages?

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,188 @@
...
+from PIL import Image
+from vacuum_map_parser_base.config.image_config import ImageConfig
+from vacuum_map_parser_base.map_data import ImageData, MapData
+
+from roborock.exceptions import RoborockException
+from roborock.map.proto.b01_scmap_pb2 import RobotMap  # type: ignore[attr-defined]
+
+from .map_parser import ParsedMapData
+
+_B64_CHARS = set(b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=")
```
allenporter: can `import string` then do
`_B64_CHARS = set((string.ascii_letters + string.digits + "+/=").encode())`

but also maybe just consider if the check for these is really needed as b64 decoding may just fail anyway with a similar error message without the check?

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def _decode_b01_map_payload(raw_payload: bytes, *, serial: str, model: str) -> bytes:
+    """Decode raw B01 `MAP_RESPONSE` payload into inflated SCMap bytes."""
+    encrypted_payload = _decode_base64_payload(raw_payload)
+    if len(encrypted_payload) % AES.block_size != 0:
+        raise RoborockException("Unexpected encrypted B01 map payload length")
+
+    map_key = _derive_map_key(serial, model)
+    decrypted_hex = AES.new(map_key, AES.MODE_ECB).decrypt(encrypted_payload)
```
allenporter: Does the existing `roborock.protocol.Utils` work for this? or if not, can that be modified/extended? I think its nice to have all the AES logic for all platforms in the same place.

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,188 @@
...
+
+class B01MapParser:
+    """Decoder/parser for B01/Q7 SCMap payloads."""
+
+    def __init__(self, config: B01MapParserConfig | None = None) -> None:
+        self._config = config or B01MapParserConfig()
+
+    def parse(self, raw_payload: bytes, *, serial: str, model: str) -> ParsedMapData:
+        """Parse a raw MAP_RESPONSE payload and return a PNG + MapData."""
+        inflated = _decode_b01_map_payload(raw_payload, serial=serial, model=model)
```
allenporter, line 51: Similar to v1 lets separate the lower level b01 protocol decoding from the map level decoding and stick this part in `roborock.protocols`

Or please add a TODO and I can move it?


You can see v1 an example:
https://github.com/Python-roborock/python-roborock/blob/7137ccd0acc84ce0e29100abbac7bf5b4fd9b4e3/roborock/protocols/v1_protocol.py#L220

Though i think its fine to level the protocol buffer parts here.

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,188 @@
...
+        return zlib.decompress(value)
+    except zlib.error:
+        return value
+
+
+def _parse_scmap_payload(payload: bytes) -> RobotMap:
+    """Parse inflated SCMap bytes into a generated protobuf message."""
+    parsed = RobotMap()
+    _parse_proto(payload, parsed, context="B01 SCMap")
+    return parsed
```
allenporter, line 129: I would say just inline/flatten this

```suggestion
def _decode_map_data_bytes(value: bytes) -> bytes:
    try:
        return zlib.decompress(value)
    except zlib.error:
        return value


def _parse_scmap_payload(payload: bytes) -> RobotMap:
    """Parse inflated SCMap bytes into a generated protobuf message."""
    parsed = RobotMap()
    try:
        parsed.ParseFromString(payload)
    except DecodeError as err:
        raise RoborockException(f"Failed to parse B01 SCMap") from err
    return parsed
```

tests/map/test_b01_map_parser.py
```diff
@@ -0,0 +1,175 @@
...
+def _derive_map_key(serial: str, model: str) -> bytes:
+    model_suffix = model.split(".")[-1]
+    model_key = (model_suffix + "0" * 16)[:16].encode()
+    material = f"{serial}+{model_suffix}+{serial}".encode()
+    encrypted = AES.new(model_key, AES.MODE_ECB).encrypt(pad(material, AES.block_size))
+    md5 = hashlib.md5(base64.b64encode(encrypted), usedforsecurity=False).hexdigest()
+    return md5[8:24].encode()
+
+
+def _encode_varint(value: int) -> bytes:
```
allenporter: Let's not encode fields manually, but use the same protobuf fields.

If you want to have some raw binary payloads somewhere to test the protobuf parsing that seems OK, but maybe in another test to verify the message parsing works.

roborock/map/b01_map_parser.py
```diff
@@ -0,0 +1,188 @@
...
+        message.ParseFromString(blob)
+    except DecodeError as err:
+        raise RoborockException(f"Failed to parse {context}") from err
+
+
+def _decode_map_data_bytes(value: bytes) -> bytes:
+    try:
+        return zlib.decompress(value)
+    except zlib.error:
+        return value
```
allenporter: Is this expected to work? If so maybe worth adding a comment about the scenarios when the response is compressed vs uncompressed would be helpful.

I think this will be hard to debug otherwise, so maybe worth a log debug.  So i'd say:
(1) If this should not happen, then raise an exception
(2) if it should happen sometimes, then it probably should log
(3) if you are unsure, ok to say that.

Issue #739, opened 2026-01-08
Add Missing Q7 Feature Support


Lash-L:
This issue is a omnibus of all of the missing features from Q7 devices.

## Higher Priority

- [ ] Map
- [ ] Cleaning history

## Lower Priority
- [ ] Consumable Resets
- [ ] Carpet Cleaning Setting
- [ ] Floor Cleaning Setting
- [ ] Robot voice volume
- [ ] DnD
- [ ] Off peak
- [ ] Child lock
- [ ] button lights
- [ ] Remote Control

Lash-L:
> how could i help here? I have a `Q7 L5+`.

Hi @ximex are you looking to help development/coding wise?

Lash-L:
> yes. or data capturing e.g. what you need. i need only some guidance how i could help. I'm not the python/c++ developer (i'm a webdev) but i help fix bugs or decode some bitmasks etc

Send me a ping on discord. You can find me on the HA server if you just search for Lash-L

RaddedMC:
I was able to obtain binary map data from my Q7 L5 unit but I'm not sure how to interpret it. It doesn't appear to be gzip or lz4 compressed so I fear it could be encrypted. 
`vacuum-map-parser-roborock` doesn't understand it either.

How could I share my data and what could I do to reverse-engineer it?

MortTalmon:
Hello. Just wanted to ask if these missing features are still worked on? And thank you fo an awesome job and the answer.

ximex:
I hope i could finish it in 2 weeks

sylr26-ui:
Hi, try to resolve map exposure from a roborock Q7 L5+ (2025).
In debug logs, roborock exposes rooms ids but cannot see the command sent from mobile app to API cloud.

All I found is "service.set_room_clean" if can help for new features.

PR #790, opened 2026-03-17
feat: Q7 Get battery level

RaddedMC:
This will complete the battery requirement for #739 .

I have tested this to be working on my Q7 L5+. I don't understand why the API denotes battery level as "quantity", but it works.

### DIFF
--- a/roborock/data/b01_q7/b01_q7_containers.py
+++ b/roborock/data/b01_q7/b01_q7_containers.py
@@ -114,7 +114,7 @@ class B01Props(RoborockBase):
     wind: SCWindMapping | None = None
     water: WaterLevelMapping | None = None
     mode: CleanTypeMapping | None = None
-    quantity: int | None = None # The Q7 L5 reports its battery level as 'quantity'
+    quantity: int | None = None  # The Q7 L5 reports its battery level as 'quantity'
     alarm: int | None = None
     volume: int | None = None
     hypa: int | None = None
@@ -168,7 +168,7 @@ class B01Props(RoborockBase):
     serial_number: str | None = None
     recommend: Recommend | None = None
     add_sweep_status: int | None = None
-    
+
     @property
     def battery(self) -> int | None:
         """


roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -84,6 +84,19 @@ async def set_clean_path_preference(self, preference: CleanPathPreferenceMapping
...
+    async def get_battery_level(self) -> int | None:
+        """Get the battery level (%)"""
+        level_query = await self.send(
+            command=RoborockB01Q7Methods.GET_PROP,
+            params={
+                "property": ["quantity"]
+            }
+        )
+        try:
+            return level_query['quantity']
```
Lash-L: what else is in the response dict?

roborock/devices/traits/b01/q7/__init__.py
```diff
@@ -84,6 +84,19 @@ async def set_clean_path_preference(self, preference: CleanPathPreferenceMapping
     async def set_repeat_state(self, repeat: CleanRepeatMapping) -> None:
         """Set the cleaning repeat state (cycles)."""
         await self.set_prop(RoborockB01Props.REPEAT_STATE, repeat.code)
+        
+    async def get_battery_level(self) -> int | None:
```
Lash-L: Is this the same quantity we get back from the main get props?

RaddedMC: Yes it is. I don't quite understand why but this corresponds to the current battery level at least on my vacuum.

Lash-L: Then we should actually be able to avoid adding this function at all and just grab the quantity property on the home assistant side I think? Does that seem to align with your understanding? Assuming you are adding this functionality for HA

RaddedMC:
Yeah I can do that! I was just exploring the Home Assistant code
I'll update my PR to add that property, and then update my PR here and create one on the Home Assistant side.

Thank you for the help!

roborock/data/b01_q7/b01_q7_containers.py
```diff
@@ -115,6 +115,7 @@ class B01Props(RoborockBase):
     water: WaterLevelMapping | None = None
     mode: CleanTypeMapping | None = None
     quantity: int | None = None
+    battery: int | None = None
```
Lash-L: So to clarify, I instead mean add a @property called battery that returns quantity. As this stands now, battery will always be None here

roborock/roborock_message.py
```diff
@@ -165,6 +165,7 @@ class RoborockB01Props(StrEnum):
     WATER = "water"
     MODE = "mode"
     QUANTITY = "quantity"
+    BATTERY = "quantity"
```
Lash-L: See the above message, you can throw a comment next to quantity though and state that it is really battery, but I don't think we should have this extra enum member.

RaddedMC:
That's been done!

Lash-L (approved):

RaddedMC:
Thank you! Now I just have to wrap up my changes on the Home Assistant side

PR #798, opened 2026-03-22
refactor: Migrate to typing.Self and remove __future__ annotations.

allenporter:
This removes the use of __future__ annotations which are converting all the types to string, making it harder to do type reflection.

This is needed to simplify https://github.com/Python-roborock/python-roborock/pull/746

One instance of annotations is left that would otherwise cause a circular import and needs to be fixed.

### DIFF
--- a/roborock/roborock_message.py
+++ b/roborock/roborock_message.py
@@ -37,7 +37,7 @@ class RoborockDataProtocol(RoborockEnum):
 
     @classmethod
     def _missing_(cls: type[Self], key) -> Self:
-        raise ValueError("%s not a valid key for Data Protocol", key)
+        raise ValueError(f"{key} not a valid key for Data Protocol")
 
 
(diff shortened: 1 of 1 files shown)


roborock/diagnostics.py
```diff
@@ -49,7 +47,7 @@ def as_dict(self) -> Mapping[str, Any]:
             data[k] = v
         return data
 
-    def subkey(self, key: str) -> Diagnostics:
+    def subkey(self, key: str) -> "Diagnostics":
```
Lash-L: Can this do Self?

I really dislike forward references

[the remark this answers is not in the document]
allenporter: I couldn't figure it out how to do it, but now see how to do it.

allenporter: I couldn't figure it out how to do it, but now see how to do it.

Lash-L (approved):

PR #799, opened 2026-03-23
feat: Implement v1 streaming data updates

allenporter:
This implements a listener that will receive streaming updates to data protocol messages. This sets the mapping for each field using the `dps` metadata and adds corresponding update listeners to the associated traints.

This uses the same dps converter pattern used by q10, but does not share code explicitly.


Lash-L (approved):
Thanks Allen

PR #813, opened 2026-04-10
feat: add Q10 remote trait

Kraina68512:
I really sorry for opening a new pull request, but i must
So, I updated the tests and checked it. should be fine now


Lash-L:
Looks pretty good. Mind addressing the nits from copilot?

roborock/data/b01_q10/b01_q10_code_mappings.py
```diff
@@ -217,3 +219,11 @@ class YXDeviceDustCollectionFrequency(RoborockModeEnum):
     INTERVAL_30 = "interval_30", 30
     INTERVAL_45 = "interval_45", 45
     INTERVAL_60 = "interval_60", 60
+
+
+class RemoteCommand(Enum):
```
Lash-L: Please use RoborockEnum or IntEnum

roborock/data/b01_q10/b01_q10_code_mappings.py
```diff
@@ -217,3 +219,11 @@ class YXDeviceDustCollectionFrequency(RoborockModeEnum):
...
+    FORWARD = 0
+    LEFT = 2
+    RIGHT = 3
+    STOP = 4
+    EXIT = 5
```
Lash-L, line 229: What is 1? Does that exist?

roborock/devices/traits/b01/q10/remote.py
```diff
@@ -0,0 +1,43 @@
...
+
+    async def right(self) -> None:
+        """Turn right."""
+        await self._send_remote(RemoteCommand.RIGHT)
+
+    async def stop(self) -> None:
+        """Stop last moving command or start remote control."""
+        await self._send_remote(RemoteCommand.STOP)
+
+    async def exit(self) -> None:
```
Lash-L: Let's name this exit_remote

roborock/devices/traits/b01/q10/remote.py
```diff
@@ -0,0 +1,43 @@
...
+
+    async def left(self) -> None:
+        """Turn left."""
+        await self._send_remote(RemoteCommand.LEFT)
+
+    async def right(self) -> None:
+        """Turn right."""
+        await self._send_remote(RemoteCommand.RIGHT)
+
+    async def stop(self) -> None:
```
Lash-L, line 37: Stop both starts and stops the control?

If it uses the same REmoteCommand thats okay, but can we split it up into two different functions? one start() and one stop()?

Lash-L:
Thanks @Kraina68512 

I left a few comments, should be fast, mind fixing them? I will be faster on the merge afterwards.

roborock/data/b01_q10/b01_q10_code_mappings.py
Kraina68512: I don't really know, but looks like it meaned to be "backward", but on the device level it was deleted, so there's a space between 0 and 2
I tried to use it, nothing

roborock/devices/traits/b01/q10/remote.py
Kraina68512: Every command that uses remote control on device starts the remote mode, except the exit command
Stop command only stops the moving command, cause the moving is forever until the device get command "stop" (idk if really moving is forever, but looks like)

Kraina68512:
Sorry for the time, I really fixed that in 2 mins
Was busy..

allenporter (approved):

PR #815, opened 2026-04-22
fix: allow protobuf 7.x

bdraco:
Widens the protobuf upper bound so downstream projects can use protobuf 7 alongside python-roborock; the generated pb2 modules remain compatible with the 7.x runtime per the protobuf cross version policy.

Home Assistant is bumping to protobuf 7.34.1 and currently cannot resolve dependencies against python-roborock 5.x; see https://github.com/home-assistant/core/pull/168811.


Lash-L (approved):
Thanks @bdraco!

PR #817, opened 2026-04-25
feat: add better cleaning mode support

Lash-L:
Adds a clean mode selector


roborock/data/v1/v1_clean_modes.py
```diff
@@ -68,6 +70,14 @@ class WashTowelModes(RoborockModeEnum):
     SUPER_DEEP = ("super_deep", 8)
 
 
+class CleaningModes(StrEnum):
```
allenporter: `CleaningMode`?

roborock/data/v1/v1_clean_modes.py
```diff
@@ -174,7 +184,105 @@ def get_water_mode_mapping(features: DeviceFeatures) -> dict[int, str]:
...
+    if features.is_pure_clean_mop_supported:
+        options.append(CleaningModes.MOP)
+    if features.is_customized_clean_supported:
+        options.append(CleaningModes.CUSTOM)
+    if features.is_smart_clean_mode_set_supported:
+        options.append(CleaningModes.SMART_MODE)
+    return options
+
+
+def get_mop_only_vacuum_mode(features: DeviceFeatures) -> VacuumModes:
```
allenporter, line 220: This is getting the vaccuum mode when the vacuum is in mop only mode?

roborock/devices/traits/v1/status.py
```diff
@@ -91,3 +100,29 @@ def mop_route_name(self) -> str | None:
...
+    def current_cleaning_mode(self) -> CleaningModes | None:
+        return get_current_cleaning_mode(
+            clean_mode=self.fan_power,
+            water_mode=self.water_box_mode,
+            mop_mode=self.mop_mode,
+            features=self._device_features_trait,
+        )
+
+    @property
+    def cleaning_mode_name(self) -> str | None:
```
allenporter: Should be `current_cleaning_mode_name`?

roborock/devices/traits/v1/status.py
```diff
@@ -91,3 +100,29 @@ def mop_route_name(self) -> str | None:
...
+        if (cleaning_mode := self.current_cleaning_mode) is None:
+            return None
+        return cleaning_mode.value
+
+    def get_cleaning_mode_parameters(self, cleaning_mode: str | CleaningModes) -> list[dict[str, int]]:
+        """Get the RPC payload for the selected high-level cleaning mode."""
```
allenporter: I was going to say this can be private, but really I think it can just be inlined. in `set_cleaning_mode`

roborock/data/v1/v1_clean_modes.py
allenporter: Some pydoc here describing how this differs from the other lower level modes would be useful. Basically never use those and always use this?

roborock/data/v1/v1_clean_modes.py
```diff
@@ -174,7 +184,105 @@ def get_water_mode_mapping(features: DeviceFeatures) -> dict[int, str]:
...
+def _get_default_mopping_water_code(features: DeviceFeatures) -> int:
+    """Pick a sensible default water code when mopping for the device."""
+    # Water-slide devices use a disjoint set of water codes; pick a mid-flow
+    # slide code instead of the standard 202, which they don't accept.
+    if features.is_water_slide_mode_supported:
+        return WaterModes.PURE_WATER_FLOW_MIDDLE.code
+    return WaterModes.STANDARD.code
+
+
+def _get_clean_motor_mode_params(mode: CleaningModes, features: DeviceFeatures) -> tuple[int, int, int]:
```
allenporter: I think this should be using the enum types instead of the code everywhere in this file, then only convert to the code when added to the params dict

roborock/data/v1/v1_clean_modes.py
```diff
@@ -174,7 +184,105 @@ def get_water_mode_mapping(features: DeviceFeatures) -> dict[int, str]:
...
+        raise RoborockUnsupportedFeature(f"Cleaning mode {mode.value!r} is not supported")
+
+    fan_power, water_box_mode, mop_mode = _get_clean_motor_mode_params(mode, features)
+    params: dict[str, int] = {"fan_power": fan_power, "water_box_mode": water_box_mode}
+    if features.is_clean_route_setting_supported:
+        params["mop_mode"] = mop_mode
+    return [params]
+
+
+def get_current_cleaning_mode(
```
allenporter, line 308: One thought is that this could be on `CleaningModes` as a classmethod like `CleaningModes.from_modes`
or similarly`get_cleaning_mode_parameters` can be `CleaningModes.get_params(...)`

I think the only advantage could be the names would be shorter, but otherwise its not to different than what is here and can just be a style choice (optional)

roborock/data/v1/v1_clean_modes.py
```diff
@@ -174,7 +184,105 @@ def get_water_mode_mapping(features: DeviceFeatures) -> dict[int, str]:
...
+    clean_mode: int | None,
```
allenporter: Can this also take the enum values instead of the codes?

roborock/data/v1/v1_clean_modes.py
```diff
@@ -174,7 +184,105 @@ def get_water_mode_mapping(features: DeviceFeatures) -> dict[int, str]:
...
+        return (VacuumModes.CUSTOMIZED.code, WaterModes.CUSTOMIZED.code, CleanRoutes.CUSTOMIZED.code)
+    if mode == CleaningModes.SMART_MODE:
+        return (VacuumModes.SMART_MODE.code, WaterModes.SMART_MODE.code, CleanRoutes.SMART_MODE.code)
+    raise RoborockUnsupportedFeature(f"Cleaning mode {mode.value!r} is not supported")
+
+
+def get_cleaning_mode_parameters(cleaning_mode: str | CleaningModes, features: DeviceFeatures) -> list[dict[str, int]]:
+    """Get the RPC payload for switching the high-level cleaning mode."""
+    try:
+        mode = CleaningModes(cleaning_mode)
```
allenporter: i'd say do the string to enum conversation as high up as possible and just take the enum here

allenporter (changes requested):

roborock/data/v1/v1_clean_modes.py
Lash-L: I have added a comment that I believe should help this make more sense. lmk

roborock/data/v1/v1_clean_modes.py
Lash-L: I don't have too strong of an opinion on this (so i can easily be convinced), but I think I'll keep it like this for now as it matches the rest of the file. Perhaps it makes sense to make all of these use clash methods instead though.

allenporter (approved):

PR #819, opened 2026-04-26
feat: Fix AppInitStatus to handle missing fields and add RoborockParsingException

allenporter:
Fix AppInitStatus to handle missing fields and improves debug messages for handling missing fields overall.

https://github.com/home-assistant/core/issues/167674

### DIFF
--- a/roborock/data/v1/v1_containers.py
+++ b/roborock/data/v1/v1_containers.py
@@ -792,7 +792,7 @@ class AppInitStatusLocalInfo(RoborockBase):
 class AppInitStatus(RoborockBase):
     local_info: AppInitStatusLocalInfo
     feature_info: list[int]
-    new_feature_info: int
+    new_feature_info: int = 0
     new_feature_info_str: str = ""
     new_feature_info_2: int | None = None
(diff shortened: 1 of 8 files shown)


Lash-L (approved):
Looks great - good exception for us to have and I think the default value will work logically wise

But I don't think this is a true fix for https://github.com/home-assistant/core/issues/167674

I think the device manager gate is being too liberal. We are just checking if protocol is v1, then we call our normal commands. I think we instead need to check to see if protocol is v1, then also check if the category mower. In my head we were already gating it, but I guess not.

allenporter:
OK so perhaps this just makes it not crash or at least makes the error message better. But you're saying it shouldn't call this method?

Lash-L:
> OK so perhaps this just makes it not crash or at least makes the error message better. But you're saying it shouldn't call this method?

It might work - but my thought is an error will pop up down the line. I.e it will try to do other vacuum specific commands on the mower. But maybe the feature checks will stop it?

I don't know enough about the mowers logic yet. I haven't looked at rr api for it yet

PR #825, opened 2026-05-03
fix: handle Web API unauthorized errors

allenporter:
# Description

This PR resolves an issue where background Web API calls (like fetching map room names) can silently fail when credentials expire, causing Home Assistant to fall back to default values instead of triggering a reauthentication flow.

The `UserWebApiClient` has been updated to automatically catch `RoborockInvalidCredentials` exceptions. Instead of allowing these to bubble up uncontrollably or get swallowed by trait-level exception handlers, `UserWebApiClient` now accepts an `unauthorized_hook` which it invokes before propagating the exception. The `DeviceManager` propagates this hook from the `create_device_manager` factory down into the `UserWebApiClient` instance.

Fixes home-assistant/core#169639

## Changes
- **`roborock/web_api.py`**: Updated `UserWebApiClient` to accept `unauthorized_hook` and wrapped `get_home_data`, `get_routines`, `get_rooms`, and `execute_routine` to invoke it when `RoborockInvalidCredentials` is raised.
- **`roborock/devices/device_manager.py`**: Updated `create_web_api_wrapper` to supply the hook to the `UserWebApiClient`.
- **`tests/`**: Added `test_user_web_api_client_unauthorized_hook` to `tests/test_web_api.py` to cover the new hook behavior and ensured existing `DeviceManager` reauth tests continue to pass.

## Testing
- [x] Verified full unit test suite passes cleanly via `uv run pytest`.
- [x] Tested the `DeviceManager` reauth propagation and `UserWebApiClient` logic directly.

### DIFF
--- a/roborock/devices/device_manager.py
+++ b/roborock/devices/device_manager.py
@@ -173,6 +173,7 @@ def create_web_api_wrapper(
     *,
     cache: Cache | None = None,
     session: aiohttp.ClientSession | None = None,
+    unauthorized_hook: SessionUnauthorizedHook | None = None,
 ) -> UserWebApiClient:
     """Create a home data API wrapper from an existing API client."""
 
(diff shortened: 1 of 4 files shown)


Lash-L (approved):
Looks good. Only side effect is that in a asyncio gathered situation the unauthorized hook may fire multiple times, but that isn't a problem with this implementation or a problem in general - we should just be mindful

PR #826, opened 2026-05-04
feat: add shared rooms endpoint

Lash-L:
relates to https://github.com/home-assistant/core/issues/169639#issuecomment-4367173154

Add a new endpoint for getting shared room ids.

### DIFF
--- a/roborock/devices/traits/v1/rooms.py
+++ b/roborock/devices/traits/v1/rooms.py
@@ -2,7 +2,6 @@
 
 import logging
 from dataclasses import dataclass
-from functools import cached_property
 
 from roborock.data import HomeData, HomeDataRoom, NamedRoomMapping, RoborockBase
 from roborock.devices.traits.v1 import common
@@ -90,17 +89,16 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
         super().__init__()
         self._home_data = home_data
         self._device_uid = device_uid
+        self._shared_device_uid = next(
+            (device.duid for device in home_data.received_devices if device.duid == device_uid), None
+        )
         self._web_api = web_api
         self._discovered_iot_ids: set[str] = set()
-        self._shared_room_names: dict[str, str] = {}
-
-    @cached_property
-    def _is_shared(self) -> bool:
-        return any(d.duid == self._device_uid for d in self._home_data.received_devices)
+        self._room_names: dict[str, str] = dict(home_data.rooms_name_map)
 
     @property
     def _room_name_map(self) -> dict[str, str]:
-        return {**self._home_data.rooms_name_map, **self._shared_room_names}
+        return self._room_names
 
     async def refresh(self) -> None:
         """Refresh room mappings and backfill unknown room names from the web API."""
@@ -120,9 +118,8 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
             _LOGGER.debug("Refreshing room list to discover new room names")
             if updated_rooms := await self._refresh_rooms():
                 _LOGGER.debug("Updating rooms: %s", list(updated_rooms))
-                if self._is_shared:
-                    self._shared_room_names = {room.iot_id: room.name for room in updated_rooms}
-                else:
+                self._room_names = {room.iot_id: room.name for room in updated_rooms}
+                if self._shared_device_uid is None:
                     self._home_data.rooms = updated_rooms
             self._discovered_iot_ids.update(new_iot_ids)
         try:
@@ -141,8 +138,11 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
     async def _refresh_rooms(self) -> list[HomeDataRoom]:
         """Fetch the latest rooms from the web API."""
         try:
-            if self._is_shared:
-                return await self._web_api.get_shared_device_rooms(self._device_uid)
+            if self._shared_device_uid is not None:
+                rooms_by_id = {room.iot_id: room for room in self._home_data.rooms}
+                shared_rooms = await self._web_api.get_shared_device_rooms(self._shared_device_uid)
+                rooms_by_id.update({room.iot_id: room for room in shared_rooms})
+                return list(rooms_by_id.values())
             return await self._web_api.get_rooms()
         except Exception:
             _LOGGER.debug("Failed to fetch rooms from web API", exc_info=True)
--- a/roborock/devices/traits/v1/rooms.py
+++ b/roborock/devices/traits/v1/rooms.py
@@ -2,6 +2,7 @@
 
 import logging
 from dataclasses import dataclass
+from functools import cached_property
 
 from roborock.data import HomeData, HomeDataRoom, NamedRoomMapping, RoborockBase
 from roborock.devices.traits.v1 import common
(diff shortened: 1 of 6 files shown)


roborock/devices/traits/v1/rooms.py
```diff
@@ -84,12 +85,22 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
...
         self._home_data = home_data
+        self._device_uid = device_uid
         self._web_api = web_api
         self._discovered_iot_ids: set[str] = set()
+        self._shared_room_names: dict[str, str] = {}
+
+    @cached_property
+    def _is_shared(self) -> bool:
+        return any(d.duid == self._device_uid for d in self._home_data.received_devices)
+
```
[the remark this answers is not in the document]
Lash-L: I think it's a pretty big edge case. It's very unlikely that shared devices changes mid run and we don't even really support that right now

 But defer to your thoughts @allenporter

roborock/devices/traits/v1/rooms.py
```diff
@@ -84,12 +85,22 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
...
         super().__init__()
         self._home_data = home_data
+        self._device_uid = device_uid
         self._web_api = web_api
         self._discovered_iot_ids: set[str] = set()
+        self._shared_room_names: dict[str, str] = {}
+
+    @cached_property
+    def _is_shared(self) -> bool:
+        return any(d.duid == self._device_uid for d in self._home_data.received_devices)
```
allenporter: We can check this up front in the constructor and if true, then set `self._shared_device_uid` then directly use that below when checking whether or not to check for shared device rooms

roborock/devices/traits/v1/rooms.py
```diff
@@ -84,12 +85,22 @@ class RoomsTrait(Rooms, common.V1TraitMixin):
...
+
+    @property
+    def _room_name_map(self) -> dict[str, str]:
+        return {**self._home_data.rooms_name_map, **self._shared_room_names}
 
     async def refresh(self) -> None:
```
allenporter, line 103: The `refresh_rooms` method used to return all rooms such that we could overwrite the room list, but now it has a conditional to call a different API then we need to check that conditional again to correctly evaluate the room response type to know how to use it.  I would say this could be improved to push the logic inside of `refresh_rooms` so that it always returns the final list of rooms.

tests/devices/traits/v1/test_rooms.py
```diff
@@ -207,3 +207,34 @@ async def test_refresh_unknown_room_names_failure_falls_back_to_room_segment_id(
...
+        await rooms_trait.refresh()
+
+        assert rooms_trait.rooms
+        assert rooms_trait.rooms[0] == NamedRoomMapping(segment_id=16, iot_id="2362048", raw_name="Example room 1")
+        assert rooms_trait.rooms[1] == NamedRoomMapping(segment_id=17, iot_id="9999999", raw_name="Office")
+        assert rooms_trait._home_data.rooms == original_rooms
+        web_api_client.get_shared_device_rooms.assert_called_once_with(rooms_trait._device_uid)
+        web_api_client.get_rooms.assert_not_called()
+    finally:
+        rooms_trait._home_data.rooms = original_rooms
```
allenporter: do we need this try/finally for this test?

tests/devices/traits/v1/test_rooms.py
```diff
@@ -207,3 +207,34 @@ async def test_refresh_unknown_room_names_failure_falls_back_to_room_segment_id(
...
+    rooms_trait: RoomsTrait,
+    web_api_client: AsyncMock,
+    mock_rpc_channel: AsyncMock,
+) -> None:
+    """Test shared devices resolve room names via the shared-device room list."""
+    original_rooms = list(rooms_trait._home_data.rooms or ())
+    try:
+        # Mark the device as shared by adding it to received_devices
+        device = next(d for d in rooms_trait._home_data.devices if d.duid == rooms_trait._device_uid)
+        rooms_trait._home_data.received_devices = [device]
```
allenporter: Instead of mutating internal state of the trait, add a new home data fixture input to `device_fixture`  so that the test can build this up as input.

(Keep in mind there is other wiring like the product information etc that needs to be matched up)

tests/devices/traits/v1/test_rooms.py
```diff
@@ -207,3 +207,34 @@ async def test_refresh_unknown_room_names_failure_falls_back_to_room_segment_id(
...
+        ]
+        room_mapping_data = [[16, "2362048"], [17, "9999999"]]
+        mock_rpc_channel.send_command.side_effect = [room_mapping_data]
+
+        await rooms_trait.refresh()
+
+        assert rooms_trait.rooms
+        assert rooms_trait.rooms[0] == NamedRoomMapping(segment_id=16, iot_id="2362048", raw_name="Example room 1")
+        assert rooms_trait.rooms[1] == NamedRoomMapping(segment_id=17, iot_id="9999999", raw_name="Office")
+        assert rooms_trait._home_data.rooms == original_rooms
```
allenporter: don't think we need to assert on this

tests/devices/traits/v1/test_rooms.py
```diff
@@ -207,3 +207,34 @@ async def test_refresh_unknown_room_names_failure_falls_back_to_room_segment_id(
...
+            HomeDataRoom(id=9999999, name="Office"),
+        ]
+        room_mapping_data = [[16, "2362048"], [17, "9999999"]]
+        mock_rpc_channel.send_command.side_effect = [room_mapping_data]
+
+        await rooms_trait.refresh()
+
+        assert rooms_trait.rooms
+        assert rooms_trait.rooms[0] == NamedRoomMapping(segment_id=16, iot_id="2362048", raw_name="Example room 1")
+        assert rooms_trait.rooms[1] == NamedRoomMapping(segment_id=17, iot_id="9999999", raw_name="Office")
```
allenporter: Worth asserting that there are two?

```suggestion
        assert rooms_trait.rooms == [
            NamedRoomMapping(segment_id=16, iot_id="2362048", raw_name="Example room 1"),
            NamedRoomMapping(segment_id=17, iot_id="9999999", raw_name="Office"),
        ]
```

tests/devices/traits/v1/test_rooms.py
```diff
@@ -207,3 +207,34 @@ async def test_refresh_unknown_room_names_failure_falls_back_to_room_segment_id(
...
+        rooms_trait._home_data.received_devices = [device]
+        rooms_trait._home_data.devices = []
+
+        web_api_client.get_shared_device_rooms.return_value = [
+            HomeDataRoom(id=9999999, name="Office"),
+        ]
+        room_mapping_data = [[16, "2362048"], [17, "9999999"]]
+        mock_rpc_channel.send_command.side_effect = [room_mapping_data]
+
+        await rooms_trait.refresh()
```
allenporter: worth asserting on the state before this is run? To verify these rooms weren't already set

allenporter (changes requested):
Very nice work!

Lash-L:
will fix commit lint on merge

allenporter:
Marking as draft just for making it easier to track inbox, happy to take a look when ready.

Lash-L:
> Marking as draft just for making it easier to track inbox, happy to take a look when ready.

I think it is ready. I did not resolve your comments on GH, but I believe I did fix them. I can check later today to be sure though.

allenporter (dismissed):
marking as requesting changes

allenporter:
Looks though theres still a whitespace lint failure.

Lash-L:
> Looks though theres still a whitespace lint failure.

Thank you, fixed!

allenporter (approved):

PR #837, opened 2026-06-03
Initial room configuration support for Q10 devices

lboue:
## Summary

Adds initial room configuration support for Q10 (B01 protocol) devices.

Introduces parsing for the `CUSTOMER_CLEAN` DPS payload and a dedicated `RoomsTrait` that is wired into `Q10PropertiesApi`. Room data is refreshed on demand and updated automatically via the subscribe loop.

## What Changed

- **`b01_q10_containers.py`** — new dataclasses (`Q10RoomVertex`, `Q10RoomConfig`, `Q10RoomsConfig`) and `parse_customer_clean_payload()` with upfront layout validation; sentinel values (0xFF / 0xFFFF) map to `None`; `rr_*` firmware names normalized via `normalize_q10_room_name()`
- **`rooms.py`** — new `RoomsTrait` exposing `refresh()`, `room_map`, `room_names`, `get_room()`, `get_room_name()`, and update listener support
- **`__init__.py`** — wires `RoomsTrait` into `Q10PropertiesApi` subscribe loop as `api.rooms`
- **Tests** — parsing edge cases (sentinels, name length variants, multiple rooms), streaming updates, listener notifications, refresh request payload

## Behaviour

- Calling `await api.rooms.refresh()` sends the `CUSTOMER_CLEAN_REQUEST` DPS; the robot replies with a `CUSTOMER_CLEAN` update that is parsed automatically
- `api.rooms.room_names` returns a `dict[int, str]` of room id → human-readable name
- Firmware names like `rr_living_room` are normalized to `Living Room`

## Validation

- 13 focused Q10 rooms tests pass locally
- Smoke-tested against a real Roborock Q10 account via local script

### DIFF
--- a/roborock/data/b01_q10/b01_q10_containers.py
+++ b/roborock/data/b01_q10/b01_q10_containers.py
@@ -6,6 +6,7 @@ dataclass fields to device Data Points (DPS). This metadata is utilized by the
 automatically update objects from raw device responses.
 """
 
+import base64
 from dataclasses import dataclass, field
 
 from ..containers import RoborockBase
@@ -22,6 +23,14 @@ from .b01_q10_code_mappings import (
     YXFanLevel,
     YXWaterLevel,
 )
+
+_Q10_CUSTOMER_CLEAN_COUNT_SIZE = 1
+_Q10_CUSTOMER_CLEAN_ROOM_BLOCK_SIZE = 26
+_Q10_CUSTOMER_CLEAN_NAME_BLOCK_SIZE = 20
+_Q10_CUSTOMER_CLEAN_VERTEX_SIZE = 4
+_Q10_CUSTOMER_CLEAN_NAME_MAX_LENGTH = _Q10_CUSTOMER_CLEAN_NAME_BLOCK_SIZE - 1
+_Q10_SENTINEL_U8 = 0xFF
+_Q10_SENTINEL_U16 = 0xFFFF
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


tests/data/b01_q10/test_b01_q10_containers.py
```diff
@@ -0,0 +1,98 @@
...
+        water_level=0xFF,
+        material=0xFF,
+        clean_line=0xFF,
+    )
+
+    parsed = parse_customer_clean_payload(payload_b64)
+
+    assert parsed.declared_count == 1
+    assert parsed.parsed_count == 1
+    room = parsed.rooms[0]
```
Lash-L, line 58: Can you add a test for an instance with multiple rooms?

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -81,6 +90,156 @@ class dpTimeZone(RoborockBase):
...
+    if not raw:
+        return Q10RoomsConfig()
+
+    count = raw[0]
+    offset = _Q10_CUSTOMER_CLEAN_COUNT_SIZE
+    rooms: list[Q10RoomConfig] = []
+
+    for index in range(count):
+        if offset + _Q10_CUSTOMER_CLEAN_ROOM_BLOCK_SIZE > len(raw):
+            break
```
Lash-L: So I'm  not sure about all of theses breaks. if one step fails, we are returning partial data, which I think can be problematic.

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -35,12 +37,26 @@ class Q10PropertiesApi(Trait):
     remote: RemoteTrait
     """Trait for sending remote control related commands to Q10 devices."""
 
+    rooms: RoomsTrait
+    """Trait for reading room configuration from Q10 devices."""
+
+    @property
```
allenporter: Seems like these should just be on RoomsTrait and not here

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -64,6 +80,18 @@ async def refresh(self) -> None:
         # to the device. Updates will be received by the subscribe loop below.
         await self.command.send(B01_Q10_DP.REQUEST_DPS, params={})
 
+    async def refresh_room_config(self) -> None:
```
allenporter: Similar here, these should all be on the rooms trait

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -81,6 +90,156 @@ class dpTimeZone(RoborockBase):
...
+    @property
+    def parsed_count(self) -> int:
+        return len(self.rooms)
+
+    @property
+    def room_map(self) -> dict[int, Q10RoomConfig]:
+        return {room.room_id: room for room in self.rooms}
+
+
+def parse_customer_clean_payload(payload_b64: str) -> Q10RoomsConfig:
```
allenporter, line 144: This seems to me like it fits better next to RoomsTrait.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -81,6 +90,156 @@ class dpTimeZone(RoborockBase):
...
+                vertices_num=vertices_num,
+                vertices=vertices,
+            )
+        )
+
+    return Q10RoomsConfig(raw_length=len(raw), declared_count=count, rooms=rooms)
+
+
+def normalize_q10_room_name(room_name: str) -> str:
+    """Normalize room names reported by the Q10 firmware."""
```
allenporter: Maybe would be good to give some examples in the pydoc of what this is doing.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -81,6 +90,156 @@ class dpTimeZone(RoborockBase):
...
+                room_name=room_name,
+                vertices_num=vertices_num,
+                vertices=vertices,
+            )
+        )
+
+    return Q10RoomsConfig(raw_length=len(raw), declared_count=count, rooms=rooms)
+
+
+def normalize_q10_room_name(room_name: str) -> str:
```
allenporter, line 219: Could also consider making this a property on `Q10RoomConfig` which exposes the name (e.g.raw_name and `name` as a property.)

roborock/data/b01_q10/b01_q10_containers.py
allenporter: Also we could just check them all at once.  Seems like the total length be the count * (26+20+1] then read vertex count then check its vertex_count * _Q10_CUSTOMER_CLEAN_VERTEX_SIZE, then we don't ned to check again.

allenporter: Wonder if using `construct` would be helpful like how some of the other packages are parsed in `protocol.py`...

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -81,6 +90,156 @@ class dpTimeZone(RoborockBase):
...
+    y_raw: int
+    x: float
+    y: float
+
+
+@dataclass
+class Q10RoomConfig(RoborockBase):
+    index: int
+    room_id: int
+    room_type: int
```
allenporter, line 115: Are these all some kind of enum value? trying to reason about what "-1" will mean here and wondering if `int | None` would be better or `RoomTypeEnum | None` if we can figure those out.

allenporter (changes requested):

Lash-L:
Marking as draft, mark it as ready once you have addressed the reviews and we can take another look!

tests/data/b01_q10/test_b01_q10_containers.py
lboue: Done — added `test_parse_customer_clean_payload_with_multiple_rooms` in `tests/data/b01_q10/test_b01_q10_containers.py`. It builds two rooms with different ids, names, and vertex counts and asserts both are parsed correctly.

roborock/data/b01_q10/b01_q10_containers.py
lboue: Replaced all the `break` statements with an upfront `_validate_customer_clean_layout` pass that walks the entire payload before any room objects are built. If validation fails we return an empty `Q10RoomsConfig` — no partial data is ever returned.

On the `construct` question: I kept the manual parsing for now. The `construct` library is used in `protocol.py` for the network-framing layer, but the `CUSTOMER_CLEAN` payload is a flat binary blob with a fixed header + per-room blocks — the current named constants and helper functions feel proportionate. Happy to revisit if the format grows more complex.

roborock/devices/traits/b01/q10/__init__.py
lboue: Done — `room_map`, `room_names`, `get_room`, and `get_room_name` are now only on `RoomsTrait`. `Q10PropertiesApi` exposes them through `api.rooms.*`.

roborock/devices/traits/b01/q10/__init__.py
lboue: Same fix — the duplicate wrappers on `Q10PropertiesApi` are gone.

roborock/data/b01_q10/b01_q10_containers.py
lboue: Done — `room_map` is now a property on `RoomsTrait` only. `Q10RoomsConfig` is a plain data class holding `raw_length`, `declared_count`, `rooms`, and `parsed_count`.

roborock/data/b01_q10/b01_q10_containers.py
lboue: Done — added four inline examples to the `normalize_q10_room_name` docstring showing the `rr_` prefix stripping, whitespace trimming, edge case for a bare `rr_` prefix, and pass-through for names that don't start with the prefix.

roborock/data/b01_q10/b01_q10_containers.py
lboue: `Q10RoomConfig` now stores `raw_room_name: str` and exposes a `room_name` property that calls `normalize_q10_room_name`.

roborock/data/b01_q10/b01_q10_containers.py
lboue: Changed to `int | None` sentinel values 0xFF and 0xFFFF now map to `None` via `_u8_to_optional` / `_u16_to_optional` helpers. The `room_type` field is left as plain `int` for now since the enum values are not yet documented; we can introduce a `RoomTypeEnum` in a follow-up once we have more data points.

PR #845, opened 2026-06-09
fix: make V1Channel re-subscribable after a failed subscribe

pike00:
## Summary

A device whose **first** connect attempt hits a transient error is bricked for the rest of the session: it never gets entities and the log shows the misleading `ValueError: Only one subscription allowed at a time`. This was observed with two V1 vacuums on one account — the second vacuum (sharing the MQTT session) reliably failed while the first connected fine.

## Root cause

The connect path is `RoborockDevice.start_connect()` → `connect_loop()` → `connect()` → `V1Channel.subscribe()`.

1. `V1Channel.subscribe()` set `self._callback = callback` **last**, and the returned `unsub()` closure tore down the reconnect task and MQTT/local subscriptions but **never cleared `self._callback`**.
2. In `connect()`, if `v1_properties.start()` raised a transient `RoborockException` (e.g. the initial status fetch timing out for a second device contending for the MQTT session), the `except RoborockException` block called `unsub()` and re-raised — leaving `self._callback` still set.
3. `connect_loop()` caught the `RoborockException` and **retried** `connect()`. The retry's `subscribe()` saw `self._callback is not None` and raised `ValueError("Only one subscription allowed at a time")`. That `ValueError` is **not** a `RoborockException`, so it escaped to `connect_loop()`'s outer `except Exception`, which logs `Uncaught error during connect` and **returns — permanently giving up**. No ready callbacks fire, so no entities are created.

A secondary leak: if `subscribe()` raised partway through (after starting `_reconnect_task` / setting `_mqtt_unsub`), it returned no `unsub`, so the reconnect task and subscriptions leaked on every retry.

## Fix

- `V1Channel.subscribe()` now **claims `self._callback` up front** and wraps the local-connect / reconnect-task / MQTT-subscribe setup in `try/except BaseException: self._teardown(); raise`.
- The teardown logic is factored into an idempotent `_teardown()` (also returned as the unsub function) that cancels the reconnect task, drops the MQTT/local subscriptions, closes the local channel, and **clears `self._callback`** — leaving the channel cleanly re-subscribable.
- `RoborockDevice.connect()`'s cleanup is broadened from `except RoborockException` to `except BaseException` so any `start()` failure (not just Roborock errors) releases the channel before propagating.

Net effect: a retry after a transient first-attempt failure now re-subscribes cleanly instead of tripping the stale-callback guard. A genuinely persistent failure becomes a normal backoff-retry loop surfacing the real error, rather than a hard brick behind a misleading `ValueError`.

## Tests

Added regression tests (all fail against the pre-fix code):

- `tests/devices/rpc/test_v1_channel.py`
  - `subscribe → unsub → subscribe` no longer raises.
  - a failure during MQTT subscribe leaves `_reconnect_task is None` and the channel re-callable (no leaked task/subscription).
- `tests/devices/test_v1_device.py`
  - `connect()` unsubscribes when `start()` fails (parametrized: `RoborockException` and a non-Roborock error).
  - end-to-end: `start()` raises a transient `RoborockException` once then succeeds → `connect_loop()` reconnects and the device becomes ready, with no `ValueError`.

`uv run pytest` — full suite passes; `ruff check` and `mypy` clean.


Lash-L:
Can you take a look at Copilots first round of reviews @pike00. I can then take a look afterwards, just mark the PR as 'ready'. Thank you!

tests/devices/rpc/test_v1_channel.py
```diff
@@ -642,3 +642,63 @@ async def test_v1_channel_dps_listener_raises_exception(
...
+) -> None:
+    """A subscribe -> unsub -> subscribe cycle must not raise.
+
+    Regression: unsub() previously failed to clear ``_callback``, so the second
+    subscribe() tripped the "Only one subscription allowed at a time" guard.
+    This is the exact failure that bricked a second vacuum sharing an account.
+    """
+    mock_mqtt_channel.response_queue.append(TEST_NETWORK_INFO_RESPONSE)
+    unsub = await v1_channel.subscribe(Mock())
+    assert v1_channel._callback is not None
```
allenporter: We should be able to test without asserting on internal components. Can the tests exercise what they need without doing this?

e.g. subscribe, then verify the subscribe callback is invoked when messages arrive (or not etc)

tests/devices/test_v1_device.py
```diff
@@ -181,3 +185,92 @@ async def test_device_trait_command_parsing(
...
+    unsub = Mock()
+    channel.subscribe = AsyncMock(return_value=unsub)
+    device.v1_properties.start = AsyncMock(side_effect=start_error)
+
+    with pytest.raises(type(start_error)):
+        await device.connect()
+
+    channel.subscribe.assert_awaited_once()
+    unsub.assert_called_once()  # channel released before propagating
+    assert device._unsub is None  # not marked connected; safe for connect_loop to retry
```
allenporter: similar here

roborock/devices/rpc/v1_channel.py
```diff
@@ -295,45 +295,66 @@ async def subscribe(self, callback: Callable[[RoborockMessage], None]) -> Callab
...
-            self._reconnect_task = loop.create_task(self._background_reconnect())
-
-        # We maintain an active MQTT subscription even when connected locally to receive
-        # unsolicited status updates (DPS push messages) directly from the cloud.
+        # Claim the subscription up front. Any failure in the setup below routes
+        # through _teardown(), which clears this again so the channel is left in
+        # a clean, re-subscribable state. Without this, a partially-completed
+        # subscribe (e.g. a transient failure later in connect()) would leave a
+        # stale callback and the next subscribe() would raise the guard above.
+        self._callback = callback
```
allenporter, line 303: what happens in the case where mqtt also fails, won't it stay stuck?

allenporter:
Thank you for the fixes here.

> V1Channel.subscribe() set self._callback = callback last, and the returned unsub() closure tore down the reconnect task and MQTT/local subscriptions but never cleared self._callback.
> In connect(), if v1_properties.start() raised a transient RoborockException (e.g. the initial status fetch timing out for a second device contending for the MQTT session), the except RoborockException block called unsub() and re-raised — leaving self._callback still set.
> connect_loop() caught the RoborockException and retried connect(). The retry's subscribe() saw self._callback is not None and raised ValueError("Only one subscription allowed at a time"). That ValueError is not a RoborockException, so it escaped to connect_loop()'s outer except Exception, which logs Uncaught error during connect and returns — permanently giving up. No ready callbacks fire, so no entities are created.

Makes sense to me. Unsetting `self._callback` on unsub() makes sense here. I'm not sure we need to change where callback is set though. Thanks for the fix!

> A secondary leak: if subscribe() raised partway through (after starting _reconnect_task / setting _mqtt_unsub), it returned no unsub, so the reconnect task and subscriptions leaked on every retry.

How would that happen? I would think we wouldn't retry this. Notice we only deal with known RoborockException cases and not other uncaught exceptions which are considered failure. co-pilot made a comment about this catching BaseException and I agree we shouldn't catch such a broad exception here. if we're missing a new unexpeceted exception type we want to fix the bug instead.

roborock/devices/device.py
```diff
@@ -202,7 +202,10 @@ async def connect(self) -> None:
                 await self.v1_properties.start()
             elif self.b01_q10_properties is not None:
                 await self.b01_q10_properties.start()
-        except RoborockException:
+        except BaseException:
+            # Any failure in start() must unsubscribe before propagating, so a
+            # retry by connect_loop() gets a clean channel. Broader than
+            # RoborockException so non-Roborock errors also release the channel.
             unsub()
             raise
```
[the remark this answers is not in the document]
pike00: Narrowed to `except Exception` in c429fb2.

roborock/devices/rpc/v1_channel.py
```diff
@@ -295,45 +295,66 @@ async def subscribe(self, callback: Callable[[RoborockMessage], None]) -> Callab
...
+                if not self.is_local_connected:
+                    # Propagate error if both local and MQTT failed
+                    self._logger.debug("MQTT connection also failed: %s", err)
+                    raise
+                self._logger.debug("MQTT subscription failed, continuing with local-only connection: %s", err)
+        except BaseException:
+            # Any failure during setup must leave the channel re-subscribable:
+            # cancel the reconnect task, drop subscriptions, and clear _callback.
+            self._teardown()
+            raise
```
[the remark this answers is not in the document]
pike00: Narrowed to `except Exception` in c429fb2.

tests/devices/test_v1_device.py
```diff
@@ -181,3 +185,92 @@ async def test_device_trait_command_parsing(
...
+    local_session = Mock(spec=LocalSession, return_value=local_channel)
+
+    # Cache the network info so local connect doesn't need an MQTT round-trip.
+    cache = InMemoryCache()
+    device_cache = DeviceCache(duid, cache)
+    await device_cache.set(DeviceCacheData(network_info=NetworkInfo.from_dict(mock_data.NETWORK_INFO)))
+
+    v1_channel = V1Channel(
+        device_uid=duid,
+        security_data=SecurityData(endpoint="test_endpoint", nonce=b"test_nonce_16byte"),
```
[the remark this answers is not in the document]
pike00: Fixed in c429fb2 — both test fixtures now use `b"test_nonce_16byt"` (16 bytes).

roborock/devices/rpc/v1_channel.py
```diff
@@ -295,45 +295,66 @@ async def subscribe(self, callback: Callable[[RoborockMessage], None]) -> Callab
...
+            self._mqtt_unsub()
+            self._mqtt_unsub = None
+        if self._local_unsub:
+            self._local_unsub()
+            self._local_unsub = None
+        if self._local_channel:
+            self._local_channel.close()
+            self._local_channel = None
+        self._callback = None
+        self._logger.debug("Unsubscribed from device")
```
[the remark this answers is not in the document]
pike00: Leaving this as-is. The three handles (`_mqtt_unsub`, `_local_unsub`, `_local_channel.close`) are internal teardown callbacks that do not raise in normal operation; wrapping each in its own try/except is speculative robustness against an exception type we have no evidence exists, and it would swallow a real bug if one ever did. Consistent with the preference to fix the underlying bug rather than add broad cleanup catches. If a specific handle is shown to raise, I would rather guard that one explicitly.

pike00:
On the partway leak — it's reachable through the existing retry path, not a new one. In `subscribe()`, after `_reconnect_task` is created, the MQTT subscribe re-raises `RoborockException` when `not is_local_connected` (the "both failed" branch). Pre-fix that re-raise handed back no `unsub`, so the running `_reconnect_task` was orphaned. The `RoborockException` propagates out of `connect()` to `connect_loop()`, which catches `RoborockException` and retries `connect()` → `subscribe()` again → a second `_reconnect_task` is created, leaking the first. So each transient "local + MQTT both failed" round leaks a task on the normal backoff loop.

On where `_callback` is set — you're right that it isn't required for the fix. The fix is teardown clearing `_callback` plus wrapping setup so teardown runs on failure. Claiming it up front only does two minor things: keeps teardown symmetric (clears exactly what subscribe claims) and closes a small window where a message arriving mid-setup would be dropped while `_callback` was still `None`. Neither is necessary — it stays correct with `_callback` set last, since teardown clears it on the failure path regardless. Happy to move it back to the end for a minimal diff if you'd prefer.

Agreed on dropping `BaseException` — narrowed both sites to `Exception` in c429fb2.

allenporter (commented):
I had some pending review comments i forgot ot send.

roborock/devices/rpc/v1_channel.py
allenporter: I think this was resolved.

tests/devices/rpc/test_v1_channel.py
pike00: Reworked in 9090ef8 to exercise behavior rather than internals:

- `test_v1_channel_resubscribe_after_unsub`: subscribes a callback, pushes a message via the channel and asserts the callback fires; unsubs and asserts the old callback stops receiving; re-subscribes a fresh callback and asserts it receives. No more `_callback` / `_reconnect_task` assertions.
- `test_v1_channel_subscribe_failure_is_atomic`: drops the `_callback` / `_reconnect_task` / `_mqtt_unsub` assertions. It now asserts the failed subscribe raises, leaves no subscriber registered on the channel, and that a subsequent subscribe succeeds with its callback receiving messages.

One trade-off worth flagging: in behavioral form the atomicity test no longer catches the original *leaked reconnect task* regression. That leak isn't observable through the public API (pre-fix code left the channel re-subscribable anyway, since `_callback` was set last), so the test now documents the contract (both transports down raises cleanly and stays re-subscribable) instead of guarding the internal cleanup. The other three regression tests still fail when run against the pre-fix commit.

tests/devices/test_v1_device.py
pike00: Same treatment in 9090ef8:

- `test_connect_unsubscribes_when_start_fails`: keeps `unsub.assert_called_once()` (the channel is released before the error propagates) but replaces `assert device._unsub is None` with a behavioral retry: once `start()` recovers, a second `connect()` succeeds (`subscribe.await_count == 2`) instead of raising "Already connected to the device".
- `test_connect_retries_after_transient_start_failure`: drops the `_unsub` assertions and relies on the retry `connect()` succeeding plus `device.is_connected`.

Both still fail against the pre-fix code (the non-Roborock-exception case and the transient-retry case respectively).

roborock/devices/rpc/v1_channel.py
pike00: Confirmed. If MQTT also fails while local is down, `subscribe()` re-raises inside the `except Exception` block, which runs `_teardown()` (cancels the reconnect task, drops the subscriptions, clears the callback) before propagating. So the channel isn't left stuck: it's reset to a re-subscribable state and `connect_loop()` retries with backoff against the real error.

`test_v1_channel_subscribe_failure_is_atomic` covers this exact case: both transports down raises, then a later subscribe succeeds and its callback receives messages.

roborock/devices/rpc/v1_channel.py
```diff
@@ -295,45 +295,66 @@ async def subscribe(self, callback: Callable[[RoborockMessage], None]) -> Callab
...
+            # unsolicited status updates (DPS push messages) directly from the cloud.
+            try:
+                self._mqtt_unsub = await self._mqtt_channel.subscribe(self._on_mqtt_message)
+            except RoborockException as err:
+                if not self.is_local_connected:
+                    # Propagate error if both local and MQTT failed
+                    self._logger.debug("MQTT connection also failed: %s", err)
+                    raise
+                self._logger.debug("MQTT subscription failed, continuing with local-only connection: %s", err)
+        except Exception:
```
allenporter, line 333: This is a style nit but: I'd like to make this clear if its expected or unexpected exceptions here by explicitly catching and forwarding both `RoborockException` and `Exception`. w/ teardown code. Essentially i'd say it like: We do expect RoborockException but this is so critical to get right to avoid leaks that we also catch Exception. I'd also like to log that it was un expected uncaught exception here.  Same in `Device`.

That is, i'd not like it to be normal to catch Exception in this code base. We can violate the rules here given this case is important to not leak, but it's a "shouldn't happen".

allenporter (commented):
Looks great, thank you for this! Ready to merge but one last comment

roborock/devices/rpc/v1_channel.py
pike00: Done in 5f25def. Both `V1Channel.subscribe()` and `RoborockDevice.connect()` now catch `RoborockException` (expected) and `Exception` (unexpected) as separate branches; each runs the teardown/unsub and re-raises, so behavior is unchanged.

The `except Exception` branch logs at `exception` level ("Unexpected error during subscribe/connect; ... to avoid a leak") and carries a comment making clear it is a deliberate, signposted exception to the usual "don't catch Exception" rule, kept only because a leaked subscription or reconnect task would brick the device. The expected `RoborockException` path stays quiet (the transient cause is already debug-logged at the point it happens).

allenporter (approved):

PR #846, opened 2026-06-14
feat: expand Q10 (B01/ss07) status support and add device info

tubededentifrice:
## Summary

Adds Q10 (`roborock.vacuum.ss07`, B01 protocol) device metadata, expands the decoded Q10 status model, and adds Q10 settings writers for the values now exposed by status.

This was checked against two physical Q10 devices.

## Changes

- Add the `roborock.vacuum.ss07` device schema to `device_info.yaml`.
- Decode the additional Q10 status fields reported by a full `dpRequestDps` response, including volume, child lock, Do Not Disturb, mop state, dust settings, map switches, cleaning line, time zone, country code, and network info.
- Fix `dpNetInfo` and `dpTimeZone` container field names so decamelized device payloads construct correctly.
- Add `SettingsTrait` writers for volume, child lock, Do Not Disturb, button light, dust collection, and dust collection frequency. Q10 setting writes are wrapped through `dpCommon` (101), which is what the device accepts.
- Allow the CLI `status` command to refresh and display decoded status for B01 Q10 devices.
- Add a scrubbed vacuum-only status fixture and tests for expanded status decoding, nested containers, and the new settings writers.

## Testing

- `uv run pytest` — 507 passed on this branch
- Live checked Q10 status and settings writes, then restored changed values


tubededentifrice:
Added Q10 settings writers for the values this PR now reads: volume, child lock, Do Not Disturb, button light, and dust collection.

Q10 setting writes need the `dpCommon` (101) wrapper. For example, setting volume sends `{"dps":{"101":{"26":65}}}`; direct writes to `26` were ignored by the tested device.

I checked volume and child lock by writing a new value, reading status back, and restoring the original value. The session CLI now has `q10-set-volume`, `q10-set-child-lock`, `q10-set-dnd`, `q10-set-led`, and `q10-set-dust-collection`.

tubededentifrice:
Added `SettingsTrait.set_dust_collection_frequency()` for `dpDustSetting` (50), using the same `dpCommon` (101) wrapper as the other Q10 settings writers.

The method accepts either `YXDeviceDustCollectionFrequency` or the raw interval code (`0`, `15`, `30`, `45`, or `60`). The session CLI exposes this as `q10-set-dust-frequency`.

Live check: changed a Q10 from daily (`0`) to every 15 cleans, read the updated value back, then restored daily.

tubededentifrice:
@Lash-L @allenporter This is ready for review when you get a chance — part of the Q10 series (#846 → #847 → #848, stacked in that order). Thanks!

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -51,18 +52,21 @@ class dpSelfIdentifyingCarpet(RoborockBase):
 
 @dataclass
 class dpNetInfo(RoborockBase):
-    wifiName: str
-    ipAdress: str
-    mac: str
-    signal: int
+    # Field names are snake_case so they match the device's camelCase keys once
```
allenporter: I think these comments may not be needed after this change is made? (e.g. this is in line with how all data classes work for roborockbase i think?)

Same for dpTimeZone below.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -108,3 +113,32 @@ class Q10Status(RoborockBase):
     back_type: YXBackType | None = field(default=None, metadata={"dps": B01_Q10_DP.BACK_TYPE})
     cleaning_progress: int | None = field(default=None, metadata={"dps": B01_Q10_DP.CLEAN_PROGRESS})
     fault: int | None = field(default=None, metadata={"dps": B01_Q10_DP.FAULT})
+
+    # Additional settings and state reported in the device's full status dump.
+    volume: int | None = field(default=None, metadata={"dps": B01_Q10_DP.VOLUME})
```
allenporter: It seems like we should be sticking some of these on different traits for the device, especially since you added setters on different traits.

Should we mirror the v1 traits and have different logical groupings for many of these things?

roborock/devices/traits/b01/q10/settings.py
```diff
@@ -0,0 +1,62 @@
...
+
+    async def set_button_light(self, enabled: bool) -> None:
+        """Enable or disable the indicator / button light (LED)."""
+        await self._write(B01_Q10_DP.BUTTON_LIGHT_SWITCH, int(enabled))
+
+    async def set_dust_collection(self, enabled: bool) -> None:
+        """Enable or disable automatic dust collection at the dock."""
+        await self._write(B01_Q10_DP.DUST_SWITCH, int(enabled))
+
+    async def set_dust_collection_frequency(self, frequency: YXDeviceDustCollectionFrequency | int) -> None:
```
allenporter: Seems like we should just make this accept YXDeviceDustCollectionFrequency and force the caller to do the enum parsing.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -108,3 +113,32 @@ class Q10Status(RoborockBase):
...
+    dust_setting: int | None = field(default=None, metadata={"dps": B01_Q10_DP.DUST_SETTING})
+    map_save_switch: bool | None = field(default=None, metadata={"dps": B01_Q10_DP.MAP_SAVE_SWITCH})
+    multi_map_switch: int | None = field(default=None, metadata={"dps": B01_Q10_DP.MULTI_MAP_SWITCH})
+    recent_clean_record: bool | None = field(default=None, metadata={"dps": B01_Q10_DP.RECENT_CLEAN_RECORD})
+    breakpoint_clean: int | None = field(default=None, metadata={"dps": B01_Q10_DP.BREAKPOINT_CLEAN})
+    valley_point_charging: bool | None = field(default=None, metadata={"dps": B01_Q10_DP.VALLEY_POINT_CHARGING})
+    carpet_clean_type: int | None = field(default=None, metadata={"dps": B01_Q10_DP.CARPET_CLEAN_TYPE})
+    clean_line: YXCleanLine | None = field(default=None, metadata={"dps": B01_Q10_DP.CLEAN_LINE})
+    ground_clean: int | None = field(default=None, metadata={"dps": B01_Q10_DP.GROUND_CLEAN})
+    line_laser_obstacle_avoidance: int | None = field(
```
allenporter: I think we may need an enum for this and all the ints below?

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -108,3 +113,32 @@ class Q10Status(RoborockBase):
...
+        default=None, metadata={"dps": B01_Q10_DP.LINE_LASER_OBSTACLE_AVOIDANCE}
+    )
+    add_clean_state: int | None = field(default=None, metadata={"dps": B01_Q10_DP.ADD_CLEAN_STATE})
```
allenporter: Does this also need an enum? I'm not sure what "add_clean_state" means as far as a status goes...

allenporter (changes requested):
Very nice, thank you for this contribution!

As always, the hard part here is naming and grouping, so curious to get your thoughts on some of my questions.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Good call, removed. I kept a single short note that the ip_adress spelling deliberately mirrors the device's ipAdress typo.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Done. I split Q10Status and the catch-all SettingsTrait into per-concern read/write traits matching the v1 layout: SoundVolumeTrait, ChildLockTrait, DoNotDisturbTrait, DustCollectionTrait, ButtonLightTrait, NetworkInfoTrait, ConsumableTrait. Each with is_on/enable()/disable() like the v1 switch traits. Q10Status now holds just core vacuum state.

roborock/devices/traits/b01/q10/settings.py
tubededentifrice: Done, it's now set_frequency(frequency: YXDeviceDustCollectionFrequency) with no int union and no manual validation; the caller does the parsing (the CLI does from_code once).

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Partly done. I converted the clear on/off flags to bool and mapped dust_setting to YXDeviceDustCollectionFrequency. For the rest I don't yet have decoded value > meaning mappings, so I left them as int with a TODO for now. Will try to reverse engineer in next commit.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: add_clean_state is the "add clean area" feature's state. I agree it's unclear; Needs some reverse engineering, will try in the next commit.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Done, reverse-engineered the rest against my robots (toggling each setting in the app and diffing the status dump). Pushed in 2ece942:
- carpet_clean_type > YXCarpetCleanType (rise=0, avoid=1, ignore=2, cross=3)
- area_unit > YXAreaUnit (square_meter=0, square_feet=1)
- ground_clean > bool (the app's "clean along floor direction" toggle)
- line_laser_obstacle_avoidance > bool
- mop_state > bool (mop module attached)
- also renamed YXDeviceDustCollectionFrequency.DAILY > REGULAR to match the app wording (codes unchanged)

The only ones still int are breakpoint_clean, timer_type, user_plan and robot_type for which I couldn't find an app control for them (or they're internal/constant?), so I left them as documented ints for now.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Decoded it on my robots: dpAddCleanState (96) pulses 0 > 1 while the app's "re cleaning" feature (draw a rectangle / add an area to the current job) request is in progress, then drops back to 0 once the robot has the area. So it's effectively a transient flag — modeled it as a bool with a comment explaining it.
Side note from the same capture: the actual drawn rectangle is sent as a zone clean (it shows up appended to dpZonedUp/59 in the same coordinate encoding as the no-go zone overlays), so a proper "clean this rectangle" write command is feasible as a follow-up, but that's a separate feature from this status-support PR.

tubededentifrice:
I've managed to get and decompile the plugin fetched from the app, I need small adjustments

tubededentifrice:
Alright @allenporter this one should be gtg 😓

tubededentifrice:
Just put together also #851 for room cleaning -- had to do packet capture in the emulator for this one! It's stacked on top of this one, so I kept it as draft for the moment, I'll properly rebase it once this one merges

tubededentifrice:
@allenporter is there anything missing before merging? thanks

lboue:
@tubededentifrice Your work is great. You've added a lot of missing features.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -51,18 +55,19 @@ class dpSelfIdentifyingCarpet(RoborockBase):
 
 @dataclass
 class dpNetInfo(RoborockBase):
-    wifiName: str
-    ipAdress: str
-    mac: str
-    signal: int
+    wifi_name: str | None = None
+    # "ip_adress" intentionally mirrors the device's "ipAdress" key (sic).
```
allenporter, line 59: we could also add an `ip_address` property in a future PR which is an alias for this.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -99,12 +106,93 @@ class Q10Status(RoborockBase):
     total_clean_area: int | None = field(default=None, metadata={"dps": B01_Q10_DP.TOTAL_CLEAN_AREA})
     total_clean_count: int | None = field(default=None, metadata={"dps": B01_Q10_DP.TOTAL_CLEAN_COUNT})
     total_clean_time: int | None = field(default=None, metadata={"dps": B01_Q10_DP.TOTAL_CLEAN_TIME})
-    main_brush_life: int | None = field(default=None, metadata={"dps": B01_Q10_DP.MAIN_BRUSH_LIFE})
```
allenporter, line 102: I think your change is the right one, but just a flag that this will be a breaking change (may be harder to push upstream changes). We'll either need to change the PR title to indicate that it is breaking, or you can leave for now and/or alias in both places and we can remove in a second step.

allenporter (changes requested):
Overall looks great. One note below about a breaking change here that we need to handle (or defer) then this is ready to merge.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Done in 39661db, added an ip_address property on dpNetInfo that aliases ip_adress, so callers can use the correctly-spelled name now.

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: Went with the alias approach in 39661db: main_brush_life/side_brush_life/filter_life/sensor_life are back on Q10Status as deprecated aliases (still populated from the same DPS), and also live on the new Q10Consumable trait. So this PR is no longer breaking — we can drop them from Q10Status in a follow-up. Left a # DEPRECATED … remove in a follow-up comment marking them.

tubededentifrice:
Thanks @allenporter kept the fields for backward compatibility, should be gtg as well

allenporter (dismissed):

allenporter:
Looks good, though there is a merge conflict now to reconcile.

I think some of the remaining attributes added to the main status object feel like its arbitrary where they end up, so i get the proposal to just add them to the status object. (However, it may be they could also belong in some other grouping, I just don't have a specific proposal and it seems arbitrary)

tubededentifrice:
@allenporter indeed, just fixed the merge conflict, unfortunately that dismissed your approval

allenporter (approved):

PR #847, opened 2026-06-14
feat: Q10 (B01/ss07) map support — rooms + rendered map image

tubededentifrice:
## Summary

Adds Q10 (`roborock.vacuum.ss07`, B01 protocol) map support: requesting current state makes the robot push a protocol-301 map packet, which this PR can parse into a rendered map image, room list, and current-session path data.

This branch is independent of #846 and is based on `main`.

## Protocol Notes

- Q10 does not appear to have a dedicated get-map command. Sending `dpRequestDps` causes the robot to publish `MAP_RESPONSE` packets shortly afterward. Firmware also throttles these pushes.
- Full map packets use the `01 01` marker. The parser reads the map id, grid width, compressed layout length, and an LZ4 block containing the occupancy grid plus room records.
- Trace packets use the `02 01` marker. On the tested firmware these contain a 10-byte header followed by big-endian int16 `(x, y)` point pairs for the current cleaning session path.
- The payloads are unencrypted and do not use the Q7 `SCMap` protobuf format.

## Changes

- Add `roborock/map/b01_q10_map_parser.py` for Q10 map, room, LZ4, and trace parsing.
- Add `request_map()` and `request_trace()` helpers on the Q10 channel.
- Add `MapContentTrait` for `refresh()`, `parse_map_content()`, `refresh_trace()`, `image_content`, `map_data`, `rooms`, `path`, `robot_position`, and `raw_api_response`.
- Dispatch CLI `map-image`, `rooms`, and Q10 position flows through the Q10 map trait.
- Add synthetic/scrubbed fixtures and tests, including maps with no room records yet.

## Testing

- `uv run pytest` — 510 passed on this branch
- Live checked map image, rooms, and trace parsing on Q10 hardware

## Credit

The initial `01 01` / `02 01` packet layout reference came from [`roborock-qseries-map-bridge`](https://github.com/v1b3c0d3x3r/roborock-qseries-map-bridge) and the related [Home Assistant community thread](https://community.home-assistant.io/t/roborock-q-series-q10-live-map-path-proof-of-concept/1009469). The implementation here is separate and is covered by tests against synthetic or scrubbed fixtures.

## Follow-ups

- Calibrate trace coordinates to image pixels for floorplan overlays.
- Add dock/charger placement once calibration is available.


tubededentifrice:
Added Q10 trace parsing for `02 01` `MAP_RESPONSE` packets and exposed it through `MapContentTrait.refresh_trace()` plus the `q10-position` session command.

The tested firmware uses a 10-byte header followed by int16 `(x, y)` point pairs. The first capture I had was a single-point packet, so the parser handles that case.

Later captures from the same cleaning session showed multi-point packets containing the accumulated current-session path; the follow-up comment below covers that final framing and the added fixture.

tubededentifrice:
Updated the trace handling with a real 15-point current-session path fixture.

A live corridor run produced `02 01` packets with 1, then 3, then 15 points. Each larger packet contained the earlier points, so the robot is serving the accumulated path for the active cleaning session rather than only the latest position.

The parser now treats the packet body as int16 `(x, y)` pairs after the 10-byte header. Byte 3 is kept as the session counter, and bytes 8-9 match point-count-minus-one on the captured packet. The path is exposed as structured data through `MapContentTrait.path` and `robot_position`; drawing it onto the map is intentionally left to the calibration/layers PR.

allenporter (changes requested):
Thank your pushing this forward!

My only high level issue is for Q10 we want to avoid faking a synchronous API where there is not one and keep things entirely asynchronous. WDYT?

tubededentifrice:
@allenporter yeah I think that's the right call, and will avoid problems with that logic in the first place, let me refactor that a bit

tubededentifrice:
@allenporter Done, pushed the push-driven refactor. Lines up with how Q10 status already works.
I also rebased the stacked map-layers PR (#848) on top of this so it uses the same push model. Entirely asynchronous now.

andrewlyeats:
Thanks for this — great to see first-class Q10 map support landing. I am independently reverse-engineering the
B01/Q10 protocol in a separate decoder (https://github.com/andrewlyeats/roborock-q10-cli), and I can corroborate most of your approach with what I'm seeing. I came to very slightly
different interpretations in a few minor areas and may have
caught a bug in your implementation that only happens to show up because my map has a dimension >256.

**1. Independent confirmations:**
- `01 01` map packet: map id `u32be@2`, compressed-layout length `u16be@27`, LZ4 block from `29` — all match.
- `02 01` trace: big-endian int16 `(x, y)` pairs, accumulating, emitted only during an active session.
- Byte 3 of the trace as a session counter (clean count, not a per-packet sequence), and reading all pairs
  rather than trusting the count field — same conclusions here.

**2. Grid width — I read it at a different offset, and `_WIDTH_OFFSET = 8` (`u16le`) breaks on my map.** In my
captures the `01 01` header carries **two consecutive `u16be` values — width at `[7:9]`, height at `[9:11]`**. A
`u16le` read at offset 8 sits one byte to the right, so its high byte is byte 9 — the **height's** high byte; it
only equals the true width when width and height land in the same 256-band.

My grid is **222 × 261**, which straddles a 256 boundary, so it's a live failure case. Header bytes 7–10 are
`00 DE 01 05`:
- `[7:9]` BE → width **222**, `[9:11]` BE → height **261**, and `_infer_layout` then finds 7 rooms cleanly.
- `[8:10]` LE → width **478** (`0xDE + 256·0x01`); no `room_count` makes `area % 478 == 0`, so `_infer_layout`
  raises **`Could not infer Q10 layout dimensions`** on this map.

The `[7:9]`/`[9:11]` BE read is consistent across all 124 of my map frames, and an independent row-stride
heuristic (the stride that maximizes vertical row-similarity — no header assumption) agrees on every one of
them. Reading height straight from `[9:11]` BE would also drop the brute-force inference. Happy to share the
capture or open a small PR with the BE read + a 222×261 regression test.

**3. Some `02 01` packets prepend a stray leading point.** It varies by clean — present in several of my
captures, absent in others (including my main one). When present, the first `(x, y)` sits far outside the map
(`≈ (0, -1907)`, while the real path starts near `(-3760, -1920)`). `robot_position` (the last point) is fine,
but it skews the rendered start/bounding box and a path-based calibration fit (relevant to #848). I drop it only
when `points[0]`'s step to `points[1]` is a gross outlier versus the median step, so a genuine first point is
never lost. I haven't pinned down the trigger (longer / multi-segment cleans are my current lead) — flagging in
case it's behind the 10-vs-18-byte trace-header discrepancy you noted against roborock-qseries-map-bridge (a
sometimes-present 4-byte leading point reads as "extra header" to a fixed-offset parser).

Thanks again, and glad to open small PRs with tests for either if that's easier than comments.

tubededentifrice:
Thanks @andrewlyeats — this is excellent, careful work, and both issues you flagged were real. Fixed in 629887a, with you credited in the commit message and code comments.

**1. Grid width (the >256 bug).** You were exactly right. We were reading width as `u16le` at offset 8, whose high byte is actually the *height's* high byte — so it only matched the true width when width and height shared a 256-band. Your `00 DE 01 05` walkthrough decodes cleanly as two consecutive `u16be` fields: width `[7:9]` = 222, height `[9:11]` = 261.

I verified this against your `FRAME_ANATOMY.md` (map id BE `2–5`, width BE `7–8`, height BE `9–10`, compressed size BE `27–28`) and independently against the ioBroker roborock adapter, which reads the same fields big-endian. It also matches our own data — our captures just happened to have both dimensions <256, where `le@8` and `be@7` coincide, which is why this slipped through.

The fix reads **both** width and height as `u16be` straight from the header (your point that this drops the brute-force inference entirely is well taken), falling back to the old `_infer_layout` only when the height field is absent. Added a **222×261 regression test** reproducing your exact straddle case (asserts the old `le@8` read would have given 478).

**2. Stray leading point.** Also reproduced and fixed, using your rule: drop `points[0]` only when its step to `points[1]` exceeds **20× the median step** of the rest of the path — so a genuine first point is never lost, and `robot_position` (the last point) is untouched. Our real 15-point corridor fixture has a 4.8× first step (an initial reposition), well under the cut, so it stays intact; added tests for both the drop and the keep cases.

On the trace-header discrepancy you mentioned: we're keeping the 10-byte header for now because it's what verifies on our ss07 (the count field reads points−1, and 1→3→15-point packets are exact supersets with a 10-byte header). Your hypothesis that a sometimes-present 4-byte leading point reads as "extra header" to a fixed-offset parser is a very plausible explanation for the 10-vs-16/18-byte spread across decoders — and the outlier-drop above should make the parser robust to it either way.

Happy to take the capture if you'd like it added as a fixture, but the synthetic 222×261 regression covers the failure deterministically. Thanks again for the precise diagnosis. 🙏

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -65,14 +73,32 @@ async def refresh(self) -> None:
         await self.command.send(B01_Q10_DP.REQUEST_DPS, params={})
 
     async def _subscribe_loop(self) -> None:
-        """Persistent loop to listen for status updates."""
-        async for decoded_dps in stream_decoded_responses(self._channel):
```
allenporter, line 69: Could we preserve this code structure here where the protocol is `b01_q10_channel` and then it returns back a message (with a type or protocol?) that is dispatched to the traits?

The idea is that `RoborockMessage` decoding lives in the rpc/protocol layer, then traits deal with parsed bytes. I realize it probably looks a little arbitrary, but it isn't too far from the structure you have already.  For example if you compare 
https://github.com/Python-roborock/python-roborock/blob/4508640aa6dbcc6bcfa8dd1ba720b7a7247ed924/roborock/protocols/v1_protocol.py#L254 is similar to the objects defined in the Map Parser.

So to be specific i'd say the protocol RPC layer should parse:
- Q10TracePacket
- Q10MapPacket

Then those can be further parsed by the map parser etc.  I noticed that the map parsing code decodes the map packet twice, and it looks there may have been an intention to parse the map packet back into the map parser, so it's not far off from working this way.


So maybe it needs to be `stream_decoded_messages` that can return either a dps or map, or trace packet?

roborock/devices/traits/b01/q10/map.py
```diff
@@ -0,0 +1,148 @@
...
+        """Reparse the cached raw map payload without performing any I/O."""
+        if self.raw_api_response is None:
+            raise RoborockException("No map payload available; no map has been pushed yet")
+
+        try:
+            parsed = self._map_parser.parse(self.raw_api_response)
+            packet = parse_map_packet(self.raw_api_response)
+        except RoborockException:
+            raise
+        except Exception as ex:
```
allenporter: Would rather not catch arbitrary exceptions (bugs) here. Is there a specific case this is trying to handle? I'd say push this handling into parse_map_packet such that it only exposes RoborockException

roborock/devices/traits/b01/q10/map.py
```diff
@@ -0,0 +1,148 @@
...
+            except RoborockException as ex:
+                _LOGGER.debug("Failed to parse Q10 trace packet: %s", ex)
+                return True
+            self.path = trace.points
+            self.robot_position = trace.robot_position
+            self._notify_update()
+            return True
+        return False
+
+    def parse_map_content(self) -> None:
```
allenporter: Can this (a) be private and (b) accept raw_api_response as an argument to avoid extra unnecessary state on the object?  A test like test_parse_without_data_raises should not be needed.

I think its helpful not to expose the raw api response.

roborock/map/b01_q10_map_parser.py
```diff
@@ -0,0 +1,392 @@
...
+# (e.g. ~(0, -1907) when the real path starts near (-3760, -1920)); it skews the
+# rendered start/bounding box and any path-based calibration. We drop points[0]
+# only when its step to points[1] is a gross outlier (this multiple of the
+# median step of the rest of the path), so a genuine first point is never lost.
+# The current position (last point) is unaffected. Trigger and threshold
+# reported and verified by @andrewlyeats across independent B01/Q10 captures.
+_STRAY_POINT_STEP_RATIO = 20
+
+
+def is_map_packet(payload: bytes) -> bool:
```
allenporter, line 142: Perhaps the code in the other modules should also use this rather than duplicating the constants

allenporter (changes requested):
Thank you very much, this is great. I've got some suggested improvements, but overall looking forward to merging this.

roborock/devices/traits/b01/q10/__init__.py
tubededentifrice: Done, restored the stream-of-messages structure. decode_message in b01_q10_protocol.py now returns a Q10Message union (Q10DpsUpdate | Q10MapPacket | Q10TracePacket), stream_decoded_messages yields it, and the subscribe loop dispatches by type. This also removed the double-parse: the map parser now exposes parse_packet(packet) so the trait renders the packet the protocol layer already parsed.

roborock/devices/traits/b01/q10/map.py
tubededentifrice: Removed — that catch was in parse_map_content, which is gone. Packet parsing only raises RoborockException now and it's handled in stream_decoded_messages.

roborock/devices/traits/b01/q10/map.py
tubededentifrice: Done — parse_map_content and raw_api_response are removed. The trait takes already-parsed packets via update_from_map_packet/update_from_trace_packet, so the raw response is no longer exposed and test_parse_without_data_raises is deleted.

roborock/map/b01_q10_map_parser.py
tubededentifrice: Good call, removed the duplicated markers from the trait. They live only in the parser now (is_map_packet/is_trace_packet), which the protocol layer uses.

tubededentifrice:
Thanks @allenporter should be gtg now (hopefully 😂 )

allenporter (approved):

allenporter:
thank you 💯

PR #850, opened 2026-06-15
feat: B01 grid-layer decomposition + Q10 vector overlay decoding

tubededentifrice:
## Summary

Splits out the standalone, low-level map-parsing primitives from #848 so they can be reviewed in parallel, without waiting on #846 and #847. Both modules are pure transforms with **no dependencies on the rest of the package**, and their tests build all inputs inline (no binary fixtures).

This is branched directly off `main` — independent of #846/#847/#848. Once it merges, #848 rebases on top and shrinks to the parts that genuinely depend on the Q10 map parser (calibration wiring, parser integration, CLI, Q7 reuse).

## Changes

- `roborock/map/b01_grid_layers.py` — device-agnostic occupancy-grid decomposition: split a grid into background / wall / per-room floor layers (via a caller-supplied `classify` callable), render any class or room as a transparent RGBA PNG, and fit a `GridCalibration` (resolution / origin / y-sign) from a known path with `solve_calibration()`.
- `roborock/map/b01_q10_overlays.py` — decode Q10 restricted-zone blobs into no-go, no-mop, and virtual-wall overlays.

## Testing

- `uv run pytest tests/map/test_b01_grid_layers.py tests/map/test_b01_q10_overlays.py` — 12 passed
- `uv run pre-commit run --files ...` — ruff format, ruff, mypy all pass

### DIFF
--- a/roborock/map/b01_grid_layers.py
+++ b/roborock/map/b01_grid_layers.py
@@ -1,0 +1,253 @@
+"""Device-agnostic decomposition of a B01 occupancy grid into map layers.
+
+Both the Q10 (custom binary) and Q7 (SCMap protobuf) deliver their map as a
+single-byte-per-cell occupancy grid where the cell *value* encodes a semantic
+class (background / wall / per-room floor / ...). This module turns such a grid
+into separable **layers** a frontend can stack, without knowing the device's
+specific value encoding -- the caller supplies a ``classifier`` mapping a cell
+value to a class name, plus the room metadata.
+
+Coordinates here are **grid-pixel** space (origin top-left of the raw grid, before
+any rendering flip/scale). Vector overlays in world/robot coordinates (path,
+zones, ...) are placed into this same space by the device's calibration.
+"""
+
+import io
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass, field
+from math import ceil
+
+from PIL import Image
+
+# Canonical layer class names. Devices map their raw cell values onto these.
+LAYER_BACKGROUND = "background"
+LAYER_WALL = "wall"
+LAYER_FLOOR = "floor"
+LAYER_UNKNOWN = "unknown"
+
+_PNG = "PNG"
+
+
+@dataclass
+class RoomLayer:
+    """A single room (segment) and where its pixels sit in the grid."""
+
+    id: int
+    name: str
+    pixel_value: int
+    pixel_count: int
+    bbox: tuple[int, int, int, int]
+    """Inclusive ``(min_x, min_y, max_x, max_y)`` bounding box in grid pixels."""
+
+    @property
+    def center(self) -> tuple[float, float]:
+        """Center of the bounding box in grid-pixel space (for label placement)."""
+        min_x, min_y, max_x, max_y = self.bbox
+        return ((min_x + max_x) / 2, (min_y + max_y) / 2)
+
+
+@dataclass
+class GridLayers:
+    """Separable layers decomposed from a single occupancy grid.
+
+    Holds a reference to the raw ``grid`` plus the classifier, and renders any
+    layer to a transparent RGBA PNG on demand (so we don't eagerly build a mask
+    per room). ``class_counts`` reports how many cells fall in each class.
+    """
+
+    width: int
+    height: int
+    grid: bytes
+    rooms: list[RoomLayer]
+    classifier: Callable[[int], str]
+    class_counts: dict[str, int] = field(default_factory=dict)
+    flip: bool = True
+    """Whether display rendering flips the grid top-to-bottom. Devices whose grid
+    is stored bottom-up (V1/Q7 convention) flip; the Q10 grid is stored top-down,
+    so it does not. Used as the default for the ``render_*`` methods so every
+    layer matches the device's composited map orientation."""
+
+    def cell_class(self, value: int) -> str:
+        """Classify a single raw cell value into a canonical layer name."""
+        return self.classifier(value)
+
+    def render_mask(
+        self,
+        predicate: Callable[[int], bool],
+        color: tuple[int, int, int, int],
+        *,
+        scale: int = 1,
+        flip: bool | None = None,
+    ) -> bytes:
+        """Render cells matching ``predicate`` as ``color`` over transparency.
+
+        ``flip`` applies the same top-to-bottom flip the composited map uses so
+        layers line up pixel-for-pixel (defaults to the device's :attr:`flip`);
+        ``scale`` nearest-neighbour upsamples.
+        """
+        if flip is None:
+            flip = self.flip
+        transparent = (0, 0, 0, 0)
+        px = bytearray()
+        for value in self.grid:
+            px.extend(color if predicate(value) else transparent)
+        img = Image.frombytes("RGBA", (self.width, self.height), bytes(px))
+        if flip:
+            img = img.transpose(Image.Transpose.FLIP_TOP_BOTTOM)
+        if scale > 1:
+            img = img.resize((self.width * scale, self.height * scale), Image.Resampling.NEAREST)
+        buf = io.BytesIO()
+        img.save(buf, format=_PNG)
+        return buf.getvalue()
+
+    def render_class(
+        self, layer: str, color: tuple[int, int, int, int], *, scale: int = 1, flip: bool | None = None
+    ) -> bytes:
+        """Render a whole class layer (e.g. ``"wall"``) to an RGBA PNG."""
+        return self.render_mask(lambda v: self.classifier(v) == layer, color, scale=scale, flip=flip)
+
+    def render_room(
+        self, room_id: int, color: tuple[int, int, int, int], *, scale: int = 1, flip: bool | None = None
+    ) -> bytes:
+        """Render a single room's pixels to an RGBA PNG."""
+        room = next((r for r in self.rooms if r.id == room_id), None)
+        if room is None:
+            raise KeyError(f"No room with id {room_id}")
+        target = room.pixel_value
+        return self.render_mask(lambda v: v == target, color, scale=scale, flip=flip)
+
+
+@dataclass
+class GridCalibration:
+    """Affine transform between device world coordinates and grid pixels.
+
+    ``resolution`` is world-units per pixel. The Y axis is flipped between world
+    and grid space via ``y_sign`` (devices typically have world-Y increasing
+    upward while the grid's Y increases downward).
+    """
+
+    resolution: float
+    origin_x: float
+    origin_y: float
+    y_sign: int = 1
+
+    def world_to_pixel(self, x: float, y: float) -> tuple[float, float]:
+        """Map a world ``(x, y)`` to grid-pixel ``(px, py)``."""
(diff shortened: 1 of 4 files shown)


andrewlyeats:
Thanks for decomposing the B01 grid and overlays.

One data point in case it helps, on the **no-mop** type. On our Q10 (`roborock.vacuum.ss07`, firmware `03.11.24`) we consistently read back, from the restricted-zone DP (`RESTRICTED_ZONE_UP`):

- `0x00` → **no-go** (matches yours)
- `0x02` → **no-mop**
- `0x03` → **door-threshold** (the thin rectangle you draw across a doorway in the app)

i.e. drawing a "no-mop" area in the app and reading the DP back gives type byte `0x02`, while the door-threshold rectangle gives `0x03`.

(For us, virtual walls read back through a separate DP, `VIRTUAL_WALL_UP` (57); in this restricted-zone DP we've only observed these three types. ioBroker splits walls out the same way.)

This lines up with ioBroker's Q10 parser, which decodes the restricted-zone payload as `areaType` `2 → mop`, `3 → threshold`, otherwise `forbid` (no-go), and handles virtual walls separately as type `1`: [`parseQ10RestrictedZoneDpPayload`](https://github.com/copystring/ioBroker.roborock/blob/435e80cf148fb929fdbed709a2113abd42d8112f/src/lib/map/q10/Q10YxMapParser.ts#L1281-L1289).

Since this PR currently treats `3` as no-mop (`ZONE_TYPE_NO_MOP = 3`), I wanted to flag the possible overlap with the threshold type — could the `3` you observed have come from a door-threshold zone, or might this be a firmware/region variant? Happy to share our raw read-backs (zone drawn → DP echoed) or to test a specific case if that would help pin it down.

tubededentifrice:
Thanks @andrewlyeats — you're right, and I've corrected it in aa321fd (credited in the commit and code comments).

I verified your read-back against the ioBroker roborock adapter's `parseQ10RestrictedZoneDpPayload`, which decodes the restricted-zone DP exactly as you describe:

```
areaType === 2 ? "mop" : areaType === 3 ? "threshold" : "forbid"
```

with virtual walls handled separately as `areaType: 1`. That matches your ss07 (fw 03.11.24) read-backs (`0x02` → no-mop, `0x03` → door-threshold), so the `3` I originally observed was indeed a door-threshold rectangle, not a no-mop area.

Fix:
- `ZONE_TYPE_NO_MOP = 2` (was `3`)
- added `ZONE_TYPE_THRESHOLD = 3` and `ZONE_TYPE_VIRTUAL_WALL = 1`
- anything else (incl. `0`) stays no-go; the raw value is still preserved on `Q10Zone.type`

Added a test asserting the constant values and one that a no-mop (2) and a door-threshold (3) zone keep distinct types. The overlay-routing consumer in the stacked map-layers PR imports `ZONE_TYPE_NO_MOP` from here, so it picks up the corrected value automatically (and threshold zones will no longer be miscolored as no-mop).

I'll still gladly take your raw `zone drawn → DP echoed` samples to add as fixtures — particularly a real threshold and a real no-mop blob — if you're up for sharing them. Thanks for catching this. 🙏

allenporter (approved):

allenporter:
Thank you @tubededentifrice 👍🏼 👍🏼 👍🏼

andrewlyeats:
Thanks @tubededentifrice — glad it helped, and thanks for the credit.

**How these were produced:** each blob below is the device's own `RESTRICTED_ZONE_UP` (DP 55) report, read from our MQTT client right after drawing the zone(s) in the official Roborock app — so an *app draw* is what provoked each one, and the base64 is exactly what the device sent, unprocessed. With nothing drawn, DP 55 reads `AQAA` (`01 00 00`, count 0). It also goes the other way: a zone written to `RESTRICTED_ZONE` (DP 54) comes back on the DP-55 report, and our codec re-encodes every blob here byte-for-byte (`encode(parse(b)) == b`).

**Where these sit in the protocol:** they're B01 data-points on the cloud-MQTT channel. The blobs above were issued by the device as its `dpRestrictedZoneUp` (55) report (`_UP` = the device→app direction); virtual walls and zoned-clean areas report the same way on `dpVirtualWallUp` (57) and `dpZonedUp` (59). The matching writes are the even code one lower — `dpRestrictedZone` (54) / `dpVirtualWall` (56) / `dpZoned` (58) — sent wrapped in `dpCommon` (101), e.g. `{"dps": {"101": {"54": "<base64>"}}}` (the same string-key COMMON envelope your #846 settings writers use). So it's an **even = write / odd = `_UP` report** pairing.

The base64 is verbatim; the hex is an annotated breakdown of those same bytes (separators, decoded `(x, y)`, and the 20-byte zero pad shown as `<20B pad>`). Layout: `[0x01 version][count : u8]`, then per zone a fixed **38-byte slot** `[type : u8][nverts : u8 = 0x04]` + `4 × (x, y) BE int16` + `20-byte zero pad`. Coords are ~5 mm units in the same world frame as the cleaning path — so they place onto the grid through the same `GridCalibration` you're building.

**1 — drew one threshold (the app's "Set Threshold" tool, type `0x03`)**
```
base64: AQEDBACD/9sAtACrAMoApgCZ/9YAAAAAAAAAAAAAAAAAAAAAAAAAAA==
hex   : 01 01 | 03 04 | 0083 ffdb 00b4 00ab 00ca 00a6 0099 ffd6 | <20B pad>
        version=01 count=01 | type=03 nverts=04 | (131,-37)(180,171)(202,166)(153,-42)
```

**2 — drew a No-mop Zone + a Set Threshold (types `0x02` then `0x03`, count = 2)** — the real **no-mop** you asked for, and it shows the `count` byte + the fixed 38-byte stride between slots:
```
base64: AQICBP1//Sr+SP0q/kj8Yv1//GIAAAAAAAAAAAAAAAAAAAAAAAAAAAMEAIP/3AC0AKsAyQClAJj/1gAAAAAAAAAAAAAAAAAAAAAAAAAA
hex   : 01 02
        | 02 04 fd7f fd2a fe48 fd2a fe48 fc62 fd7f fc62 <20B pad>   ← type=02 no-mop    (-641,-726)(-440,-726)(-440,-926)(-641,-926)
        | 03 04 0083 ffdc 00b4 00ab 00c9 00a5 0098 ffd6 <20B pad>   ← type=03 threshold (131,-36)(180,171)(201,165)(152,-42)
```

**3 — drew a No-Go Zone + a No-mop Zone (types `0x00` then `0x02`, count = 2)** — adds the no-go (`0x00`) ground truth:
```
base64: AQIABADH/6cBM/+nATP+WwDH/lsAAAAAAAAAAAAAAAAAAAAAAAAAAAIEASr+GwHy/hsB8v1TASr9UwAAAAAAAAAAAAAAAAAAAAAAAAAA
hex   : 01 02
        | 00 04 00c7 ffa7 0133 ffa7 0133 fe5b 00c7 fe5b <20B pad>   ← type=00 no-go   (199,-89)(307,-89)(307,-421)(199,-421)
        | 02 04 012a fe1b 01f2 fe1b 01f2 fd53 012a fd53 <20B pad>   ← type=02 no-mop  (298,-485)(498,-485)(498,-685)(298,-685)
```

These decode cleanly through `parse_zone_blob`, so they should drop straight in as fixtures for the zone overlays in #848. For cross-reference against the app's own map-edit tools (we drew each and confirmed the echo): **No-Go Zone → `0x00`**, **No-mop Zone → `0x02`**, **Set Threshold → `0x03`** — all three are the *same* `RESTRICTED_ZONE` DP (54/55), even though the app presents "Set Threshold" as its own tool. The app's **Invisible Wall** is the separate `VIRTUAL_WALL` DP (56/57) below, and **Carpet** is a different DP again (`CARPET_UP`, JSON, not this binary scheme). (Matches ioBroker's `areaType` 2 → mop / 3 → threshold.)

**One heads-up that matters for #848:** `parse_zone_blob`'s docstring lists `VIRTUAL_WALL_UP` (DP 57) among the blobs it handles — but on our ss07, **virtual walls (the app's "Invisible Wall") use a different shape and coordinate order**, so a DP-57 blob won't parse through it correctly. DP 57 is `[count : u8]` (no version byte, no per-record `type`/pad) then, per wall, **two `(y, x)` int16 points** — `(y, x)`, not `(x, y)`. Provenance: drew one wall in the app, read back from DP 57:
```
base64: Aflu+cX87PoO          (= 01 | f96e f9c5 fcec fa0e)
                              count=01 | wall (y=-1682, x=-1595) -> (y=-788, x=-1522)
```
Fed to `parse_zone_blob`, the leading `01` is taken as a version and the next byte (`0xf9`, a coordinate) as a count of 249 records, so the blob mis-frames. Walls just need their own small decode: `[count]`, then 8-byte `(y1, x1, y2, x2)` records. An empty wall list is `AA==` (`00`).

Two more in the same family, FWIW: zoned-clean rectangles ride `ZONED_UP` (DP 59) — same `[version][count] + [type=0x01][nverts][4×(x,y)] + pad` scheme as the restricted zones above; and all of these are writable, in case writers ever come up — we set zones/walls back through the string-key `COMMON(101)` envelope (e.g. `{"101":{"54": <blob>}}`) and the device echoes them on the `_UP` report, round-trip-verified.

If our working decode is handy as a reference (sometimes clearer than prose), it all ships in `decode_map.py` (v0.1.3): the virtual-wall `(y, x)` parse/encode is [L418–449](https://github.com/andrewlyeats/roborock-q10-cli/blob/v0.1.3/decode_map.py#L418-L449), the slot-aware restricted-zone parse/encode (the 38-byte stride walk) is [L499–538](https://github.com/andrewlyeats/roborock-q10-cli/blob/v0.1.3/decode_map.py#L499-L538), and `parse_carpets` (the carpet JSON) is just below.

Thanks again — happy to capture more (a specific type, more vertices, an empty → set → clear sequence, …) if it'd help.

andrewlyeats:
Two coordinate follow-ups — including a correction to myself. As far as I can tell none of this affected the merge (`parse_zone_blob` looks unwired so far), so I'm flagging it for completeness before the overlays get used — and it's all scoped to what we see on our one ss07, so treat it as a data point rather than a general rule.

**Correction:** I said walls (DP 57) use a *different coordinate order* than the zones (DP 55) — that was wrong. On our captures they're the **same** order; the only DP-57-vs-DP-55 difference we see is the record **framing** (`[count]` + bare 8-byte records vs `[version][count]` + typed/padded slots), which is what stops a wall blob parsing through `parse_zone_blob`.

**The bit that might matter for `GridCalibration`:** on our ss07 we're seeing the zone/wall vertices come back with x and y **swapped relative to the `0201` path points** — a vertex's first int16 lines up with the path's **Y** and its second with **X** (`(y, x)` in path terms). `parse_zone_blob` returns them as `(x, y)` (same as the trace) and the docstring places them "in the same space as the cleaning path," so on our map, feeding parsed vertices straight into the path-fit calibration lands the zones **transposed**. Swapping each vertex (`(y, x)`) before calibrating fixes it for us — framing/types/values unchanged. I can't rule out that some orientation/setting on our map makes this true for us and not in general, so it's worth a sanity check on another device before relying on it.

We hit this exact transpose in our own renderer; caught it by drawing a zone in the app and predicting its grid cell — happy to share a worked before/after when you wire the overlays in. Apologies for the mix-up!

PR #851, opened 2026-06-15
feat: Q10 (B01/ss07) room/segment cleaning (clean_segments)

tubededentifrice:
## Summary

Adds Q10 (`roborock.vacuum.ss07`, B01 protocol) **room / segment cleaning** — sending the robot to clean only a chosen set of rooms.

Room **retrieval** already exists (the map parser in #847 exposes `map.rooms`, each with an `id` and `name`). This PR adds the missing **command** side: `VacuumTrait.clean_segments(segment_ids)`.

This is stacked on **#846** (which introduces the Q10 `VacuumTrait`). Until #846 merges, GitHub will show those commits here too; the incremental work is the single `feat: Q10 (B01/ss07) room/segment cleaning` commit.

## Protocol notes

Unlike whole-home (`{"dps": {"201": 1}}`) and spot (`{"dps": {"201": 5}}`), which take a bare task code, segment cleaning sends `dpStartClean` (201) as an **object**:

```json
{"dps": {"201": {"cmd": 2, "clean_paramters": [<room_id>, ...]}}}
```

- `cmd` `2` is the segment-clean task code (`1` = whole-home, `2` = segment, `3` = zone, `4` = build map, `5` = spot).
- `clean_paramters` (sic) is the list of room ids — this intentionally mirrors the **device's misspelling** of "parameters"; the firmware only accepts that exact key.
- The room ids are the **same ids the map reports** (`map.rooms[i].id`), so callers can go straight from `rooms` to `clean_segments`.

The payload was **captured from the official app** (the generic vacuum plugin only implements whole-home; room cleaning is built in the app's native map flow, so the shape isn't visible in the plugin source) and then **verified live against ss07 hardware**: it drives the device into `clean_task_type -> 2` (electoral) and cleans exactly the selected room.

## Changes

- Add `VacuumTrait.clean_segments(segment_ids: list[int])`.
- Update the `start_clean` docstring now that the segment payload is known.
- Add a `q10-clean-segments` CLI command (`--segments 9,2`).
- Add trait payload tests for `clean_segments`.

## Testing

- `uv run pytest tests/devices/traits/b01/q10/ tests/protocols/test_b01_q10_protocol.py` — 50 passed
- `uv run pre-commit run --files ...` — ruff format, ruff, mypy, codespell all pass
- Live: confirmed on ss07 hardware that `clean_segments([9])` starts a room clean (status → cleaning, `clean_task_type` → electoral), cross-checked against the room ids from `rooms`.

## Usage

```python
[...]

PR #852, opened 2026-06-18
fix(web_api): sign request body in Hawk auth for B01 /jobs writes

### LINKED ISSUE #849: B01 /jobs writes (schedules + one-time room cleans) 401 — Hawk auth signs path only; POST/PUT/DELETE must sign the request body
## B01 `/jobs` writes 401 because Hawk auth omits the request-body hash

**Device:** Roborock Q10 S5+ (`roborock.vacuum.ss07`, B01 protocol)
**Source:** empirical — captured the official app's HTTPS traffic and reproduced its Hawk MAC byte-for-byte, June 2026

### Summary

For B01 devices, the schedule/clean REST API lives at `/user/devices/{duid}/jobs`. **GET works, but every write (POST/PUT/DELETE) returns `401 auth.err.invalid.token`.** This is *not* a token-scope problem — the same credentials read fine. The cause is the **Hawk signature**: the library signs the request **path only** and leaves the body-hash slot empty, which is correct for GET but wrong for writes. The app signs `md5(compact_json(body))` in that slot for writes. Once the body hash is included, the identical credentials write successfully.

This was a long-standing red herring (it looks exactly like a read-only token) — worth documenting so others don't chase token scope.

### Where it is

`web_api.py`'s Hawk helper (the `_get_hawk_authentication`/request-signing path used by e.g. `execute_scene`) builds the normalized request string with the **`hash`/`ext` (body) field empty for all methods**. Hawk's normalized string includes a payload-hash position; for a write, that position must contain the base64 (or, here, the hex `md5`) digest of the exact request body. GET legitimately uses an empty payload hash, which is why reads have always worked and writes have always 401'd.

### How to reproduce / how we confirmed it

- GET `/user/devices/{duid}/jobs` with a normal logged-in credential → **200**.
- PUT/POST/DELETE the same path → **401 `auth.err.invalid.token`**, regardless of token freshness or re-login.
- Captured the app performing the same write and recomputed its Hawk `mac`. It reproduces **exactly** when, and only when, the signed payload slot = `md5(compact_json(body))` (compact = `json.dumps(body, separators=(",", ":"))`). Path-only and `k=v` form-encoded payloads both fail to reproduce the MAC.

### The fix (two halves, both required)

1. **Sign the body.** Put `md5` of the **compact** JSON (`separators=(",", ":")`, no spaces) into the Hawk payload-hash position for POST/PUT/DELETE. Keep it empty for GET/HEAD.
2. **Send the exact same bytes you signed.** Don't hand the dict to the HTTP client's `json=` serializer — it re-serializes with spaces, so the server's hash of the *received* body won't match your signature. Send `data=<the same compact bytes>` with `Content-Type: application/json`.

Because the signed string and the wire body are byte-identical, the server's body-hash check passes regardless of key ordering.

### What this unblocks for B01

- **Scheduling writes:** create / enable / disable / delete cron jobs at `/jobs`.
- **One-time room-targeted cleans:** a one-time job (`repeated: false`, dated cron a minute or two out) with `param.rooms: [...]` is effectively "clean these rooms now" — the practical room-clean path for B01, which has no working MQTT segment-clean command exposed to third-party clients.

Job body schema (captured):
```
{ cron, repeated, enabled,
  param: { mapId, rooms:[...], roomCount, cleanMode, cleanRoute, fanLevel, waterLevel, cleanCount } }
```
(One gotcha: the REST `fanLevel` scale differs from the MQTT fan enum at the top tier — the max tier is contiguous on REST, not the MQTT enum's value.)

Happy to open a PR implementing the body-signing path (GET unchanged; POST/PUT/DELETE sign + send compact body) and a small test that reproduces the app's MAC.

andrewlyeats:
Fixes #849.

B01 `/jobs` writes fail with `401 auth.err.invalid.token`. This is the REST path for **schedules** and any
deferred/cron job — the one place `/jobs` is the only option (immediate room cleans now have an MQTT route via
#851; *scheduling* a future clean still has to go through `/jobs`). `_get_hawk_authentication` always derives
the Hawk payload slot from `formdata`, which is empty for a JSON body, so the MAC never covers the body the
server receives. GET and bodyless `DELETE /jobs/{id}` are unaffected.

The fix is two coupled parts:
- When a JSON `body` is signed, put `md5(compact-JSON body)` in the payload slot.
- Send those same compact bytes (`data=`), since `json=` re-serializes with spaces and breaks the MAC.

With `body=None`, GET and form-post signing stay byte-identical.

Verified on a Q10 S5+ (`roborock.vacuum.ss07`): with body signing, `POST`/`PUT /jobs` return 200; the current
empty-payload-slot signature returns 401. The official Roborock **iOS app** sends the body **compact** (its
`Content-Length` of 179 equals the no-space serialization, not the longer spaced form), so the exact bytes
must be signed and sent (`data=`, not `json=`). The complete `PUT /jobs` from the iOS app, observed via an
HTTPS proxy — every line in its original order, only sensitive **values** redacted in place (`Authorization`,
`X-UID`, `Cookie`) and device identifiers placeheld:

```
PUT /user/devices/{deviceId}/jobs/{jobId}
Host: api-us.roborock.com
X-IOTSDK-VERSION: 3.4.0_5_09
Accept: */*
Authorization: Hawk id=<redacted>,s=<redacted>,ts=1781315205,nonce=<redacted>,mac=<redacted>
X-APP-NAME: com.roborock.smart
X-UID: <redacted>
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US;q=1, es-US;q=0.9, zh-Hans-US;q=0.8
X-APP-VERSION-CODE: 4.64.02
Content-Type: application/json
Content-Length: 179
User-Agent: com.roborock.smart/4.64.02_2 (iPhone14,8; iOS 26.5; Scale/3.00)
Connection: keep-alive
Cookie: <redacted>
{"cron":"05 10 * * ?","repeated":true,"enabled":true,"param":{"mapId":<mapId>,"cleanRoute":2,"fanLevel":3,"cleanMode":1,"rooms":[],"roomCount":0,"waterLevel":2,"cleanCount":2}}
```

The auth change is generic for any body-bearing write; I've wired one `create_job` consumer in the existing
`get_*` style. Happy to adjust naming or scope in review.

_This is my first PR; AI-assisted, reviewed and refined with care by me: Andrew, a human._

### DIFF
--- a/roborock/web_api.py
+++ b/roborock/web_api.py
@@ -1,6 +1,7 @@
 import base64
 import hashlib
 import hmac
+import json
 import logging
 import math
 import secrets
@@ -646,6 +647,31 @@ class RoborockApiClient:
         else:
             raise RoborockException(f"schedule_response result was an unexpected type: {schedules}")
 
+    async def create_job(self, user_data: UserData, device_id: str, job: dict) -> dict:
+        """Create a /jobs entry (schedule or one-time room clean) on a B01 device.
+
+        Body-bearing writes must sign the request body in the Hawk payload slot and send those same
+        compact bytes via ``data=``; ``json=`` would re-serialize with spaces and break the MAC.
+        """
+        rriot = user_data.rriot
+        if rriot is None:
+            raise RoborockException("rriot is none")
+        if rriot.r.a is None:
+            raise RoborockException("Missing field 'a' in rriot reference")
+        path = f"/user/devices/{device_id}/jobs"
+        job_request = PreparedRequest(
+            rriot.r.a,
+            self.session,
+            {
+                "Authorization": _get_hawk_authentication(rriot, path, body=job),
+                "Content-Type": "application/json",
+            },
+        )
+        response = await job_request.request("post", path, data=_compact_json(job).encode())
+        if not response.get("success"):
+            raise RoborockException(response)
+        return response
+
     async def get_products(self, user_data: UserData) -> ProductResponse:
         """Gets all products and their schemas, good for determining status codes and model numbers."""
         base_url = await self.base_url
@@ -738,11 +764,25 @@ def _process_extra_hawk_values(values: dict | None) -> str:
         return hashlib.md5("&".join(result).encode()).hexdigest()
 
 
-def _get_hawk_authentication(rriot: RRiot, url: str, formdata: dict | None = None, params: dict | None = None) -> str:
+def _compact_json(body: dict) -> str:
+    """Serialize a JSON body to the exact compact bytes that are both signed and sent."""
+    return json.dumps(body, separators=(",", ":"))
+
+
+def _get_hawk_authentication(
+    rriot: RRiot,
+    url: str,
+    formdata: dict | None = None,
+    params: dict | None = None,
+    body: dict | None = None,
+) -> str:
     timestamp = math.floor(time.time())
     nonce = secrets.token_urlsafe(6)
-    formdata_str = _process_extra_hawk_values(formdata)
     params_str = _process_extra_hawk_values(params)
+    if body is not None:
+        payload_str = hashlib.md5(_compact_json(body).encode()).hexdigest()
+    else:
+        payload_str = _process_extra_hawk_values(formdata)
 
     prestr = ":".join(
         [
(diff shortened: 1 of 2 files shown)


Lash-L (approved):
All looks right by the app code!

PR #853, opened 2026-06-19
fix: move CLI dependencies from `dev` to `cli`

shields:
The roborock CLI imports click, click-shell, pyyaml and pyshark, but pyyaml and pyshark were only in the dev dependency group, so a plain `pip install python-roborock` or `uvx --from python-roborock roborock` would fail. Also, click/click-shell were in the main deps but are used only by the CLI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,11 @@ Install this via pip (or your favourite package manager):
 
 `pip install python-roborock`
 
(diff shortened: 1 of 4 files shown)


Lash-L (approved):

Lash-L:
Something I have been meaning to do for a while - thank you!

PR #856, opened 2026-06-26
feat: add set_start_time and set_end_time to V1 timer traits

allenporter:
Expose new methods on V1 timer traits to update quiet/off-peak start and end times directly with datetime.time objects.

### DIFF
--- a/roborock/devices/traits/v1/do_not_disturb.py
+++ b/roborock/devices/traits/v1/do_not_disturb.py
@@ -1,3 +1,5 @@
+import datetime
+
 from roborock.data import DnDTimer
 from roborock.devices.traits.v1 import common
 from roborock.roborock_typing import RoborockCommand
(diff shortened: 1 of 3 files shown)


roborock/devices/traits/v1/do_not_disturb.py
```diff
@@ -21,6 +23,30 @@ async def set_dnd_timer(self, dnd_timer: DnDTimer) -> None:
...
+            end_hour=self.end_hour,
+            end_minute=self.end_minute,
+            enabled=self.enabled,
+        )
+        await self.rpc_channel.send_command(RoborockCommand.SET_DND_TIMER, params=timer.as_list())
+        await self.refresh()
+
+    async def set_end_time(self, end_time: datetime.time) -> None:
+        """Set the end time of the Do Not Disturb (DND) timer."""
+        timer = DnDTimer(
```
Lash-L, line 40: should we be checking if by chance self.start_hour or minute is None here? That's the root of the bug Paulus was having right? same thing above?

allenporter: OK, sure. Once we check them for None what are you expecting to happen? Send in zeros or nulls? (i believe this is current behavior) or send nothing? lie to the caller and set end time on the trait without sending the RPC?

Lash-L: So I think it shouldn't be possible to set a DND end time without passing in a start time. I'd have to test it but it seems like it would be wrong. i.e. if DND is somehow None, then the start time should be set before the end time.

Although, start time would be broken then, as the same issue could happen there.

Lash-L: I can play around with some tests later.

Lash-L: Okay confirmed @allenporter 

Vacuum will respond with '-10005: At least 1 parameter is not number' if any of the keys are None. 

So we have to make sure we don't send None, maybe we need one more function that allows setting start and end time at the same time? I'm not sure if start/end will ever be None? My gut assumption would be that they default to some value, but I don't have proof for that.

PR #857, opened 2026-06-27
feat: Q10 (B01/ss07) clean-record history trait

andrewlyeats:
Implements #767 for B01/ss07 (Q10).

Adds a `CleanHistoryTrait` that fetches and decodes the Q10 clean-record history. Unlike the Q7's synchronous `get_record_list` RPC, the Q10 is push-driven: `refresh()` sends `{"op":"list"}` for `dpCleanRecord` (DP 52) over `dpCommon`, and the device publishes its record list back on the subscribe stream.

### Shape
- `CleanRecordConverter` parses a `dpCleanRecord` envelope — the full `op:"list"` reply or a single `op:"notify"` push — into a `CleanRecordPush`; the trait's `_apply` does the merge-or-replace, one sort, one notify.
- Each record decodes into a `Q10CleanRecord` (12 underscore fields), with `clean_mode`/`work_mode`/`cleaning_result`/`start_method` typed as enums via `from_code_optional` (an unmapped code → `None`; the original stays in `raw`).

### Validation (real ss07 Q10, fw 03.11.24)
- `op:"list"` decoded 25 records live on the device's own reply topic; the end-of-clean `op:"notify"` push was captured on a genuine whole-apartment completion and upserted correctly.
- 12-field layout cross-confirmed against ioBroker's `Q10CleanRecordService` (independent implementation).
- Unit tests + real-payload fixtures under `tests/protocols/testdata/b01_q10_protocol/`.

### Caveats (one device observed)
- `clean_mode=2` and `start_method=0` were never seen on our unit — left unmapped rather than guessed (they surface as `None`; `raw` retains the value).
- `work_mode` uses `YXCleanType` (matching `Q10Status`); it caps at 4 while the live `CLEAN_MODE` DP can reach 6, though records only ever show 1/2/3.
- Per-record map download (the `map_len` gate) is intentionally out of scope — a natural follow-up.


roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -32,6 +35,146 @@ class dpCleanRecord(RoborockBase):
...
+    REMOTE = 0
+    APP = 1
+    TIMER = 2
+    BUTTON = 3
+
+
+_E = TypeVar("_E", bound=IntEnum)
+
+
+def _safe_clean_enum(enum_cls: type[_E], value: int | None) -> _E | None:
```
allenporter: Can you take a look at the enum definitions in code_mappings and use them above? I believe it will give you equivalent functionality to what is needed here, but i didn't look really closely.

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -32,6 +35,146 @@ class dpCleanRecord(RoborockBase):
...
+        return None
+
+
+@dataclass
+class Q10CleanRecord(RoborockBase):
+    """A single Q10 (ss07) clean record decoded from a ``dpCleanRecord`` (DP 52) entry.
+
+    The device returns each record as an underscore-delimited string in the ``data``
+    list of a ``{"op": "list"}`` query. The field names and meanings here follow the
+    device's own app, which splits the string into these 12 values. The ``*_len``
```
allenporter: don't need to mention device app here, just say its how the format works

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -32,6 +35,146 @@ class dpCleanRecord(RoborockBase):
...
+    """Cleaning time, minutes."""
+    clean_area: int | None = None
+    """Cleaned area in square meters."""
+    map_len: int | None = None
+    """Length of the saved map blob for this record (0 = none stored)."""
+    path_len: int | None = None
+    """Length of the saved path blob for this record (0 = none stored)."""
+    virtual_len: int | None = None
+    """Length of the saved virtual-restriction blob for this record (0 = none stored)."""
+    clean_mode: int | None = None
```
allenporter: Once you update the fields above it should be possible to just directly add the enum types here and RoborockBase should parse them correctly, so you don't need the additional enum property method converters.

roborock/devices/traits/b01/q10/clean_history.py
```diff
@@ -0,0 +1,104 @@
...
+        :meth:`update_from_dps` processes the ``dpCleanRecord`` push.
+        """
+        if self._command is None:
+            raise ValueError("Trait is read-only; no command channel was provided")
+        await self._command.send(
+            B01_Q10_DP.COMMON,
+            params={str(B01_Q10_DP.CLEAN_RECORD.code): {"op": "list"}},
+        )
+
+    def update_from_dps(self, decoded_dps: dict[B01_Q10_DP, Any]) -> None:
```
allenporter, line 144: Can you review the converter pattern used in other objects and extract separate objects for parsing the data vs updating state on the object? I realize this is more complex since it supports a merge vs an update, but i do think we still want to separate concerns a little bit more here. The trait can manage merge vs update still.

We can simplify this so we don't need two copies of the code that rebuilds the list, sorts it, and notifies, since that can happen in one place.   So i think that would be:
- define a format for the notification
- have a class for parsing the dps to that object
- the trait then will get the object, decide if its merging or updating (single if statement branch), then sort (one impementation) and notify (one implementation).

My gut says the string conversion logic belongs in the converter /trait and not in the container.

roborock/data/b01_q10/b01_q10_containers.py
allenporter: Can you add some full payload examples in /tests/protocols/testdata/b01_q10_protocol? it serves as nice documentation of strings we actually see in practice from the device, since it can be a little harder to understand the full payload in unit tests

allenporter (changes requested):
Thank you @andrewlyeats this is fantastic. I have some code structure related comments, though overall look forward to merging this feature.

roborock/devices/traits/b01/q10/clean_history.py
```diff
@@ -0,0 +1,104 @@
...
+        UpdatableTrait.__init__(self, command, _LOGGER)
+        self.records: list[Q10CleanRecord] = []
+        """Decoded clean records, most recent first."""
+
+    @property
+    def last_record(self) -> Q10CleanRecord | None:
+        """The most recent clean record, or ``None`` if there are none."""
+        return self.records[0] if self.records else None
+
+    async def refresh(self) -> None:
```
allenporter, line 130: I assume that the top level request dps refresh does not refresh this list, is that right?

andrewlyeats:
Thanks — addressed all five, pushed.

**Enums:** `work_mode` now uses `YXCleanType` (matching `Q10Status`'s live clean-mode). Added `YXCleanScope`/`YXCleaningResult`/`YXStartMethod` for the three with no existing match. Two notes:
- `clean_mode` deliberately isn't `YXDeviceCleanTask` — the persisted record uses a different int encoding than the live task DP: a full clean records `0` (`idle` there), a single-room clean records `1` (`smart` there). Checked across ~90 records vs the live DP-138 distribution. Own enum; the docstring notes the relationship.
- `work_mode = YXCleanType` for consistency with `Status`; one caveat — it caps at 4 and the live `CLEAN_MODE` DP can reach 6, though records only ever show 1/2/3. Say the word if you'd prefer `YXDeviceWorkMode` (covers 5/6).

**Fields/converters:** enums directly, property layer gone. Conversion runs in the converter via `from_code_optional` — records come from the string, not the DPS stream, so `RoborockBase` isn't in that path; unmapped → `None`, `raw` keeps the original.

**Structure:** `CleanRecordConverter` (envelope → `CleanRecordPush`) + `_apply` (one merge/replace, one sort, one notify).

**Docs/tests:** dropped the app mention; added `dpCleanRecord_{list,notify}.json` fixtures (snapshot-covered).

---

These "contributions" (I hope they are!) have been a collaboration between a physicist trying to learn about modern coding tools and one of those tools, Claude Code. We both hope our efforts to be gracious and follow open source software etiquette are apparent in our contributions here. If we are in fact injecting noise, please tell us and we will stop! We thank you for your work on this repo!

roborock/devices/traits/b01/q10/clean_history.py
```diff
@@ -0,0 +1,164 @@
...
+        Returns ``None`` for an envelope that carries nothing usable (so the trait can
+        ignore it without changing state). Malformed individual records are skipped.
+        """
+        if envelope.get("op") == "notify":
+            record = self.parse_record(envelope.get("id"))
+            return CleanRecordPush([record], replace=False) if record is not None else None
+        data = envelope.get("data")
+        if not isinstance(data, list):
+            return None
+        records = [record for item in data if (record := self.parse_record(item)) is not None]
```
allenporter: parse record is static so doesn't need a "self" reference

roborock/devices/traits/b01/q10/clean_history.py
```diff
@@ -0,0 +1,164 @@
...
+        return CleanRecordPush(records, replace=True)
+
+    @staticmethod
+    def parse_record(raw: object) -> Q10CleanRecord | None:
```
allenporter:
```suggestion
    def parse_record(raw: Any | None) -> Q10CleanRecord | None:
```

allenporter (dismissed):
Can you explicitly address / reply / resolve each comments so we know they're completed? (or ask the AI to do it)

roborock/data/b01_q10/b01_q10_containers.py
andrewlyeats: Done. `work_mode` now uses the existing `YXCleanType` (matching `Q10Status.clean_mode`); the three fields with no existing match got new enums in `code_mappings` — `YXCleanScope`, `YXCleaningResult`, `YXStartMethod`. `clean_mode` deliberately stays its own enum (`YXCleanScope`) rather than reusing `YXDeviceCleanTask`: the persisted record encodes those ints differently from the live DP-138 (a full clean records `0`, a single-room clean `1`, checked across ~90 records), so reusing it would mislabel them; the docstring notes the relationship.

roborock/data/b01_q10/b01_q10_containers.py
andrewlyeats: Done — removed the app reference; the docstring just describes the format now.

roborock/data/b01_q10/b01_q10_containers.py
andrewlyeats: Done — the fields are the enum types directly now and the property-converter layer is gone. One nuance worth flagging: these records arrive as an underscore-joined **string** (`op:list`/`op:notify`), not via the DPS stream, so `RoborockBase`'s converter isn't in that path — the string→enum conversion runs in `CleanRecordConverter.parse_record` via `from_code_optional` (an unmapped code → `None`, with `raw` keeping the original). So the property methods are gone as requested; the conversion just lives in the converter.

roborock/devices/traits/b01/q10/clean_history.py
andrewlyeats: Done — split into `CleanRecordConverter` (parses an `op:list`/`op:notify` envelope → a `CleanRecordPush`) and `_apply` on the trait. `_apply` is the single merge-vs-replace branch → one sort → one notify, so there's no longer a duplicated list-rebuild/sort/notify. The string parsing moved out of the container into the converter, per your note.

roborock/data/b01_q10/b01_q10_containers.py
andrewlyeats: Done — added `dpCleanRecord_list.json` and `dpCleanRecord_notify.json` under `tests/protocols/testdata/b01_q10_protocol/`, using real device payloads (snapshot-covered).

roborock/devices/traits/b01/q10/clean_history.py
andrewlyeats: Correct — and I just verified it live against the device with this trait: after the top-level `refresh()` (which sends `REQUEST_DPS`), the general status fields populate but `clean_history.records` stays at `0`; a `clean_history.refresh()` (the `op:list`) then brings the full 25. So the record list isn't in the `REQUEST_DPS` dump — it's pull-only via `{"op":"list"}` on `dpCleanRecord` (52) over `dpCommon` (plus the `op:notify` single-record push on clean-completion). That's why the trait has its own `refresh()`.

roborock/devices/traits/b01/q10/clean_history.py
andrewlyeats: I class-qualified it in ec279ef (`CleanRecordConverter.parse_record(...)`) since it's static. If I misunderstood, I'm happy to pull `parse_record` out to a module-level function instead.

andrewlyeats:
Replied to all comments in their respective threads and closed them out. Made both changes you suggested. Two of my commit messages didn't pass commit-message lint so had to force-push to correct. That force-push should not have changed any code, just the messages. I think this should now be resolved and ready for you. Thanks for your guidance!

allenporter (approved):

allenporter:
Thank you @andrewlyeats for the contribution!

PR #859, opened 2026-07-02
refactor: decouple B01 (Q7/Q10) protocol layer from transport layer

allenporter:
# Decouple B01 (Q7 / Q10) Protocol and Transport Layers

This PR decouples protocol-specific codecs, encryption/decryption, map-key derivation, and payload formatting from the device traits layer for B01 version (Q7 and Q10) devices.

## Rationale & Rationale for Home Assistant Test Fakes
Previously, Q7 and Q10 traits communicated directly with raw `MqttChannel` objects. This forced the traits (and their corresponding unit tests) to handle low-level packet encoding, encryption signing, AES ECB/CBC padding, and local keys.

By separating the traits layer from the transport layer, we introduce clean boundaries where traits communicate over simple plaintext RPC channels (`Q7RpcChannel`, `Q7MapRpcChannel`, and `Q10RpcChannel`).

A key motivation for this change is to **improve the test fakes exposed to Home Assistant**. Rather than forcing downstream integrations to mock deep internals or mock raw MQTT messages, Home Assistant will be able to instantiate the actual `python-roborock` device objects (using real traits classes) and simply feed them with plaintext channel fakes. This will happen in separate PR and will make testing the Home Assistant integration much more straightforward and reliable.

---

## Architectural Changes

- **Unified Channel Layer**: Introduced `B01Q7Channel` and `B01Q10Channel` wrapping MQTT transport. They implement the low-level `Channel` interface and directly conform to the plaintext RPC protocols (`Q7RpcChannel`, `Q7MapRpcChannel`, and `Q10RpcChannel`). All AES block padding, encryption, signature verification, and map key derivation details are fully encapsulated here.
- **Simplified Traits**: Refactored the Q7 and Q10 traits to use the plaintext interfaces, removing all dependencies on raw messages, local keys, and transport codecs.
- **Simplified Unit Tests**: Trait unit tests now use simple fake channels (`FakeQ7Channel` and `FakeB01Q10Channel`) that verify plaintext parameters and enqueue plaintext mock responses, eliminating the need for `B01MessageBuilder` in trait tests.
- **New Channel Tests**: Created dedicated unit test suites (`test_b01_q7_channel.py` and `test_b01_q10_channel.py`) to verify encryption/decryption, padding, and packet routing in isolation.


roborock/devices/rpc/b01_q10_channel.py
```diff
@@ -1,9 +1,13 @@
 """Thin wrapper around the MQTT channel for Roborock B01 Q10 devices."""
 
+from __future__ import annotations
```
Lash-L: Where are we using this?

allenporter: Agent added this, but its not needed. Removed. (For context, when this is used its not to import directly, but to make typing changes that are compatible with older versions of python, but we aren't using those)

Lash-L: Yes sorry! Know what future annotations does, just didn't see anything that needed it so I figured it was a agent add

Lash-L (changes requested):
There are merge conflicts

tests/devices/traits/b01/q7/conftest.py
```diff
@@ -1,20 +1,38 @@
...
+        self.published_commands.append((command, params))
+        if self.response_queue:
+            return self.response_queue.pop(0)
+        return {}
+
+    async def send_map_command(self, command: Any, params: Any = None) -> bytes:
+        self.published_commands.append((command, params))
+        if self.response_queue:
+            return self.response_queue.pop(0)
+        return b""
```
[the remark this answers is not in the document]
allenporter: I will solve this when we build new fakes for Q7/Q10 in continued work after https://github.com/Python-roborock/python-roborock/pull/860

allenporter:
> There are merge conflicts

Fixed. Ready for review.

Lash-L (approved):

PR #860, opened 2026-07-02
Add roborock.testing — stateful device simulators for integration testing

allenporter:
## Summary

Introduces a new `roborock.testing` package that provides stateful firmware simulators, fake transport channels, and cloud environment fakes. This allows downstream consumers (like the Home Assistant integration) to write high-fidelity integration tests using the real client library classes instead of fragile top-level mocks.

## Motivation

Today, testing against this library requires mocking at the Python API level (patching methods, return values, etc.), which is brittle and doesn't exercise the real protocol handling, message encoding, or trait update logic. This package fakes communication at two well-defined boundaries:

1. **HTTP API** — `aioresponses` intercepts network requests so the real `RoborockApiClient` executes fully against fake endpoints.
2. **Plaintext RPC** — Device communication is intercepted at the `RoborockMessage` level. The real `V1Channel`, subscribers, and DPS listeners all run under test.

## Architecture

```
┌────────────────────────────────────────────────────────┐
│               TESTED CLIENT (REAL CODE)                │
│                                                        │
│  RoborockDevice / Traits / V1RpcChannel / V1Channel    │
└──────────────────────────┬─────────────────────────────┘
                           │
                 ROBOROCKMESSAGE PAYLOADS
                 (Plaintext JSON commands)
                           │
┌──────────────────────────▼─────────────────────────────┐
│                 SIMULATOR (TEST FAKE)                  │
│                                                        │
│  FakeChannel (Intercepts publish/subscribe)            │
│  RoborockDeviceSimulator (Stateful firmware simulator) │
└────────────────────────────────────────────────────────┘
```

### `roborock/testing/`

| File | Description |
|------|-------------|
| `__init__.py` | Package exports and architecture documentation |
| `channel.py` | `FakeChannel` — in-memory transport with mock attributes for failure injection and call inspection |
| `simulator.py` | `RoborockDeviceSimulator` — base class with optional local/MQTT channels |
| `v1_simulator.py` | `V1VacuumSimulator` — stateful V1 firmware simulator with command handlers for `get_status`, `app_start`, `app_stop`, `set_custom_mode`, consumables, DND, etc. |
| `cloud.py` | `FakeRoborockCloud` — cloud environment orchestrator with `FakeUserState`, `FakeWebApiClient`, and `patch_device_manager()` context manager |

### `tests/testing/`

| File | Tests |
|------|-------|
| `test_channel.py` | FakeChannel subscribe/unsubscribe/publish/notify |
| `test_cloud.py` | Device discovery, unsupported protocol errors, login error overrides, dynamic device addition |
| `test_v1_simulator.py` | Consumable refresh/reset, DND refresh, fan speed changes, clean summary, state transition lifecycle, push update propagation, custom handler overrides |

### Modified

| File | Change |
|------|--------|
| `tests/fixtures/channel_fixtures.py` | Now re-exports `FakeChannel` from `roborock.testing` instead of defining it locally |

## Usage Example

```python
from roborock.testing import FakeRoborockCloud, V1VacuumSimulator

async def test_vacuum_cleaning():
    cloud = FakeRoborockCloud()
    fake_device = V1VacuumSimulator(duid="living_room_s7", battery=100)
    cloud.add_device(fake_device)

    with cloud.patch_device_manager():
        manager = await create_device_manager(
            user_params=UserParams(username="test_user", user_data=USER_DATA),
            cache=InMemoryCache(),
        )
        devices = await manager.get_devices()

    device = devices[0]
    await device.v1_properties.command.send("app_start")
    assert fake_device.state == RoborockStateCode.cleaning
```

### Migrate existing tests to use `roborock.testing`

An earlier review identified which existing test suites should migrate to use these fakes vs. remain as-is:

**High priority — should migrate:**
- `tests/devices/test_device_manager.py` — Currently mocks `create_v1_channel` and `create_mqtt_channel` with fragile `send_command` side-effect arrays. Using `patch_device_manager()` with a `V1VacuumSimulator` would eliminate the brittle handshake sequence mocking entirely.
- `tests/devices/test_v1_device.py` — Manually stubs subscribe, connection, and channel creation with `AsyncMock`. Using a real `V1Channel` bound to `FakeChannel` would test actual client routing logic.

**Medium priority — optional migration:**
- Trait-specific unit tests (e.g. `test_status.py`, `test_consumable.py`) — Direct mocks are fine for parsing edge cases, but integration-level tests using the simulator would complement them.

**Should remain unchanged:**
- E2E / session tests (`tests/e2e/`) — These verify raw protocol byte parsing, TCP framing, and cryptography at Layer 1. Our fakes operate above that layer, so these tests serve a different purpose.

### Additional enhancements
- **A01/B01 simulators** — `RoborockDeviceSimulator` supports `has_local_channel=False` for MQTT-only protocols, but no A01 (Zeo/Dyad) or B01 (Q7/Q10) simulators exist yet. `patch_device_manager()` currently raises `NotImplementedError` for these.
- **More V1 command handlers** — Only core commands are implemented. Future handlers could cover room-specific cleaning (`app_segment_clean`), map operations, volume control, etc.
- **More trait coverage** — Tests currently cover status, consumables, DND, clean summary, and command. Additional traits like rooms, maps, wash towel mode, and dust collection mode could be exercised.
- **Multi-device scenarios** — Testing multiple devices registered simultaneously, with different protocols or configurations.
- **Error simulation** — More granular transport error scenarios (timeouts, partial responses, reconnection behavior).


roborock/testing/channel.py
```diff
@@ -0,0 +1,118 @@
...
+        #   channel.connect.side_effect = RoborockException("refused")
+        self.connect = AsyncMock(side_effect=self._connect)
+
+        # MagicMock wrapping _close. Callers can assert close was called
+        # or inject errors on teardown.
+        self.close = MagicMock(side_effect=self._close)
+
+        self.protocol_version = LocalProtocolVersion.V1
+        self.restart = AsyncMock()
+        self.health_manager = HealthManager(self.restart)
```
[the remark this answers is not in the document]
allenporter: Done. Reverted `subscribe` back to an `AsyncMock` wrapping the private `_subscribe` method to preserve backward compatibility with all existing test call assertions.

roborock/testing/channel.py
```diff
@@ -0,0 +1,118 @@
...
+            response = self.response_queue.pop(0)
+            self.notify_subscribers(response)
+
+    async def subscribe(self, callback: Callable[[RoborockMessage], None]) -> Callable[[], None]:
+        """Register a callback and return an unsubscribe function.
+
+        This is a real method implementing the Channel protocol, not a mock.
+        """
+        self.subscribers.append(callback)
+        return lambda: self.subscribers.remove(callback)
```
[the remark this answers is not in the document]
allenporter: Done. Re-implemented `subscribe` as an `AsyncMock` and moved the actual logic into a private `_subscribe` method.

roborock/testing/simulator.py
```diff
@@ -0,0 +1,105 @@
...
+
+    async def _handle_local_publish(self, message: RoborockMessage) -> None:
+        assert self.local_channel is not None
+        self.local_channel.published_messages.append(message)
+        await self._handle_publish(message, self.local_channel)
+
+    async def _handle_mqtt_publish(self, message: RoborockMessage) -> None:
+        self.mqtt_channel.published_messages.append(message)
+        await self._handle_publish(message, self.mqtt_channel)
+
```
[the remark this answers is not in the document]
allenporter: Done. Updated the simulator publish handlers to check for and raise `channel.publish_side_effect` before routing messages, and added integration tests to verify.

roborock/testing/simulator.py
```diff
@@ -0,0 +1,105 @@
...
+    async def _handle_publish(self, message: RoborockMessage, channel: FakeChannel) -> None:
+        """To be overridden by subclasses to route commands."""
+        pass
```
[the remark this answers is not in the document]
allenporter: Done. The base `_handle_publish` now raises `NotImplementedError` to ensure subclasses explicitly implement it.

roborock/testing/channel.py
```diff
@@ -0,0 +1,122 @@
...
+
+from collections.abc import Callable
+from unittest.mock import AsyncMock, MagicMock
+
+from roborock.mqtt.health_manager import HealthManager
+from roborock.protocols.v1_protocol import LocalProtocolVersion
+from roborock.roborock_message import RoborockMessage
+
+
+class FakeChannel:
```
Lash-L: Have not gone all the way through yet, but does it make sense to make this a subclass of a the real object?

roborock/testing/cloud.py
```diff
@@ -0,0 +1,259 @@
...
+            "msg": "success",
+        }
+
+    def get_login_payload(self) -> dict[str, Any]:
+        """Synthesize login payload using the cloud user profile state."""
+        if self.login_payload is not None:
+            return self.login_payload
+        return {
+            "code": 200,
+            "data": {
```
Lash-L: Do any of these datas make sense to use our pythonic objects and dump them? I'm going back and forward a bit on it.

roborock/testing/v1_simulator.py
```diff
@@ -0,0 +1,354 @@
...
+DEFAULT_NETWORK_INFO = {
+    "ip": "1.1.1.1",
+    "ssid": "test_wifi",
+    "mac": "aa:bb:cc:dd:ee:ff",
+    "bssid": "aa:bb:cc:dd:ee:ff",
+    "rssi": -50,
+}
+
+# Simulated application init parameters
+DEFAULT_APP_GET_INIT_STATUS = {
```
Lash-L: Same thing here, do you think it makes more sense to use a dict or our real objects? Can't come to a decision

Our real objects give us the benefit of typing and some validation and may be a extra way we can catch some errors, but it complicates some logic and i'm not sure if it's really worth it.

Lash-L (dismissed):
Really cool stuff. minimal comments, I think the most important things will just become clear as it starts being used.

roborock/testing/cloud.py
allenporter: Yes, i also went back and forth on it. Using the dataclass objects seems more usable and less likely to make mistakes. However I leaned towards this to make these more like raw protocol texts (like the json files) to catch issues even in the parsing layer, but are slightly more annoying to use.

allenporter: Let me see what it looks like with dataclasses...

roborock/testing/channel.py
allenporter: Updated. I think the agent didnt want to do this at first for some typing reasons, but we can fix by adding the subscribe property.

roborock/testing/v1_simulator.py
allenporter: The one ceavt with using our dataclasses is that it may obscure round tripping errors, but it may catch testing errors (setting the wrong field in a test). Those layers are straight forward to test and fix independently, however, so lets go with the dataclasses.

Lash-L (approved):

PR #863, opened 2026-07-05
feat: decode Q10 (B01/ss07) dpFault codes with a YXFault enum

### LINKED ISSUE #855: Q10 (B01/ss07): fault codes aren't fully implemented — `dpFault` (DP 90) surfaces as a raw int; here's an app-sourced fault table + crosswalk to `B01Fault`
On the Q10 (`roborock.vacuum.ss07`, B01), faults aren't really decoded yet: the container surfaces `fault: int` straight from `dpFault` (DP 90), so a fault reaches the user as a bare number. There's no Q10 equivalent of the Q7 `B01Fault` map, and `B01Fault` itself isn't referenced from `b01_q10`. (Doesn't look tracked in the Q10 omnibus #767 yet.)

To help fill that in, I pulled the Q10's fault/notification strings from the device's app bundle and built a crosswalk against the existing `B01Fault`. Columns:

- **ss07 app text** — verbatim from the app's `error_<code>` strings (English).
- **Seen live** — ✅ if I observed that raw `dpFault` value on my own Q10 during normal use (otherwise it's app-sourced only, not hardware-confirmed).
- **`B01Fault` (Q7, on `main`)** — the existing upstream label for the same number, or "—" if there's no entry.
- **Notes** — where the Q7 label wouldn't fit the ss07 meaning, or where a code is really status/info, not a fault.

| code | ss07 app text (verbatim) | seen live | `B01Fault` (Q7, on `main`) | notes |
|---|---|:---:|---|---|
| 1 | LiDAR turret or laser blocked. Check for obstruction and retry. | | — | |
| 2 | Bumper stuck. Clean it and lightly tap it to release it. | | — | |
| 3 | Robot suspended. Move the robot away and restart. | | — | |
| 4 | Cliff sensor error. Clean cliff sensors, move the robot away from drops, and restart. | | — | |
| 5 | Clean the main brush and bearings. | | — | |
| 7 | Rotate the main wheels for cleaning. Move the robot away and restart. | | — | |
| 8 | Robot trapped. Move the robot away and restart. | ✅ | — | |
| 9 | Make sure the dustbin and filter are installed properly. | | — | |
| 12 | Low battery. Recharge now. | | — | |
| 14 | Temperature threshold. Wait for the temperature to return to normal. | | — | |
| 16 | Robot tilted. Move it to flat floor and restart. | | — | |
| 21 | Ensure the LiDAR sensor cover is not obstructed. | | — | |
| 24 | No-Go Zone or Invisible Wall detected. Move the robot away. | | — | |
| 27 | Mopping module stuck. Remove foreign objects if any. | | — | |
| 28 | Carpet avoidance mode. Move the robot away and restart. | | — | |
| 29 | Unable to cross the carpet. Move the robot across the carpet and restart. | | — | |
| 46 | Install the dust bag. | | — | |
| 54 | The mop mount fell off. | | — | |
| 58 | LiDAR sensor may be dirty. Wipe it. | | — | |
| 301 | The filter has reached its service life. Please replace it immediately. | | — | consumable-life notice |
| 302 | The main brush has reached its service life. Please replace it immediately. | | — | consumable-life notice |
| 303 | The side brush has reached its service life. Please replace it immediately | | — | consumable-life notice |
| 304 | The sensor is left uncleaned for a prolonged period of time. Please clean it. | | — | consumable-life notice |
| 310 | Dust bag full. Empty it. | | — | |
| 400 | Starting scheduled cleaning. | ✅ | — | status/info, not a fault |
| 407 | Cleaning in progress. Scheduled cleaning ignored. | | `cleaning_in_progress` (407) | matches; status/info |
| 500 | Empty the dustbin. It has been full for a long time. | | `lidar_blocked` (500) | Q7 label differs |
| 501 | Cleaning completed. Returning to dock. | | `robot_suspended` (501) | Q7 label differs; status/info |
| 502 | Low battery. Resume cleaning after recharging. | | `low_battery` (502) | matches |
| 503 | Docking error. Clear obstacles around the dock and clean charging contacts. | | `dustbin_not_installed` (503) | Q7 label differs |
| 556 | Positioning failed. Starting a new task. | ✅ | — | relocalization; alert |
| 569 | Too frequent emptying attempts. Try again later. | | `main_wheels_entangled` (569, = `F_568`) | Q7 label differs |
| 570 | Unable to reach the target. Cleaning incomplete. | ✅ | `main_brush_entangled` (570) | Q7 label differs — Q7's "cannot reach target" is `F_2007`/`F_2012` |
| 588 | Offline warning: The robot has been asleep for a long time and is about to shut down. Charge it. | | `fault_588` (588) | Q7 has only a generic placeholder |
| 589 | Offline warning: Battery too low. The robot is about to shut down. Charge it. | | `fault_589` (589) | Q7 has only a generic placeholder |
| 591 | In Do Not Disturb mode, Auto Top-up is disabled. | | `fault_591` (591) | Q7 generic placeholder; status/info |
| 707 | Clean carpet ultrasonic sensors, and move robot away. | | — | |
| 1002 | Robot error. Reset the system. | | — | |
| 3001 | Voice pack update available. Update? | | — | notification, not a hardware fault |

A couple of things to flag from the crosswalk: several `dpFault` values are really status/info rather than errors (400, 407, 501, 502, 3001…), and where a number also exists in `B01Fault` the Q7 label sometimes doesn't fit the ss07 meaning — most clearly **570**, which is `main_brush_entangled` in `B01Fault` but "unable to reach the target" on the Q10 (Q7's own cannot-reach-target is `F_2007`/`F_2012`). So the Q7 map isn't a safe drop-in for the Q10. (The app only exposes these display strings — no identifier names — so a Q10 enum's labels would be yours to name, the same way the existing `B01Fault` ones are.)

I'm not sure how much of this gap is a genuine per-model difference vs. worth mapping, so I didn't want to assume the fix. Happy to help however works best for you — put up a PR adding a `b01_q10` fault map seeded from this table, just make these labels available alongside the raw int, or leave it if the gap is intentional. A couple of extras I can fold in if useful: most of the hardware-fault codes (1–29) also carry a longer `error_des_<code>` remediation string in the bundle (the same short-plus-long pairing your `B01Fault` comments already use), and the app groups these into error / alert / offline / other buckets for surfacing. Thanks for all the recent B01/Q10 work.

andrewlyeats:
Following through on #855 — this types `Q10Status.fault` (DP 90, wired as a raw `int` in #792) the same way the Q7 handles it: a `YXFault | None` field plus a `fault_name` accessor, mirroring `B01Props.fault` / `fault_name`.

### Why a Q10-specific map (and not `B01Fault`)

Reusing the Q7 `B01Fault` — the obvious alternative — would actively mislabel this device: several shared codes mean different things on ss07. The divergent ones are hardware-confirmed on my unit (code fired live, in a context matching the label):

| code | Q7 `B01Fault` label | actual ss07 meaning | evidence |
|---|---|---|---|
| **501** | `robot_suspended` | cleaning completed, returning to dock | ~45 distinct captured events across 9+ days; every context-checked instance fires at the exact second clean progress hits 100% (per completed task) and self-clears in ~3 s |
| **503** | `dustbin_not_installed` | docking error | fired on a failed dock approach, with the matching "Could not find the dock" notification |
| **570** | `main_brush_entangled` | cannot reach target | 3 independent occurrences, each on an unreachable/confined target (0 m² result) |

A user seeing `570` would be sent to untangle a brush that's fine — and when this device's brush actually *was* entangled (twice, two different objects jammed physically), it reported `5` (`main_brush_stuck`), not `570`. `501` is worse in the other direction: a benign "clean finished" reads as an error.

### Provenance (single ss07 unit, so all n=1)

- **16 codes hardware-confirmed** — observed firing live with matching context: 2, 3, 4, 5, 8, 12, 46, 304, 400, 407, 501, 502, 503, 556, 569, 570. Marked `hw-confirmed` in the enum. Four of these were confirmed by *deliberately engineering the condition* and catching the code at the predicted moment: a due schedule ignored mid-clean (407), the dock bag removed (46 — fires at both dock-contact and clean-start; with the bag in, docking is clean), a third auto-empty requested too soon (569 — the dock refuses to run the cycle), and the bumper held in during commanded motion (2). (`5` additionally has physical ground truth: two separate real brush jams, each with a matching notification.)
- **We have not been able to find any third-party source for these mappings.** I searched integrations (ioBroker/openHAB/HA/Valetudo/tuya-local/localtuya and smaller projects), Tuya's docs, and user reports in several languages: nothing we found maps B01 `dpFault` numbers — the interface only ever shows users translated strings, never the raw numbers, so reports quoting them may simply never occur. Treat this table as single-source; the per-code comments say which labels are hardware-backed vs spec-only.
- **The rest are from the ss07 fault spec** (labels are mine to name; only display strings are exposed). One code (310) is contextual inference and commented as such.
- **569 has a small story:** I initially dropped it because its label couldn't be supported — then reproduced it deterministically (request a third auto-empty within ~15 minutes and the dock refuses the cycle and emits 569), which vindicated the spec label "too frequent emptying", so it's back in with hardware provenance. The Q7 table's `main_wheels_entangled` for 569 is another cross-model divergence.
- `dpFault` is **overloaded** — some values are lifecycle, not errors (400 = scheduled clean starting, 501/502 = returning/recharge). The converse also holds and is noted in the docstring: the device can sit in its error *state* (dpStatus 12) with `dpFault` still 0 (observed live with the dust-bin module removed), so a `NONE` fault doesn't imply no error. Happy to also add the spec's blocking-vs-alert buckets in a follow-up if that's useful.

### Tradeoff: unmapped codes decode to `None`

Typing the field means an *unmapped* code goes through `from_code`, fails conversion, and the field is left `None` (the library-wide enum-field behavior). The fault space is open-ended — a firmware update can mint new codes — so the raw number is lost exactly when it's most interesting. That's already true of every other enum field here and matches the Q7, so this follows the pattern; happy to reshape to keep the raw int reachable instead (e.g. an `int` field + a decoded property) if that's preferred.

### Scope

Data layer only (`roborock/data/b01_q10/` + tests) — no trait/transport files; rebased onto current main (post-#859). Tests cover the ss07-divergent decodes, the from_dict path, and unmapped-code behavior; the one existing assertion touched (`test_status.py` `fault == 0`) now reads `fault is YXFault.NONE`, consistent with its neighbors.

Closes #855.

### DIFF
--- a/roborock/data/b01_q10/b01_q10_code_mappings.py
+++ b/roborock/data/b01_q10/b01_q10_code_mappings.py
@@ -321,7 +321,7 @@ class YXFault(RoborockModeEnum):
     MOPPING_MODULE_STUCK = "mopping_module_stuck", 27
     CARPET_AVOIDANCE = "carpet_avoidance", 28
     CANNOT_CROSS_CARPET = "cannot_cross_carpet", 29
-    INSTALL_DUST_BAG = "install_dust_bag", 46
+    INSTALL_DUST_BAG = "install_dust_bag", 46  # hw-confirmed (docked with bag removed; fires at dock contact)
     MOP_MOUNT_FELL_OFF = "mop_mount_fell_off", 54
     LIDAR_DIRTY = "lidar_dirty", 58
     FILTER_SERVICE_LIFE = "filter_service_life", 301
--- a/roborock/data/b01_q10/b01_q10_code_mappings.py
+++ b/roborock/data/b01_q10/b01_q10_code_mappings.py
@@ -295,6 +295,10 @@ class YXFault(RoborockModeEnum):
     ``dpFault`` is *overloaded*: several values are lifecycle/status rather than
     errors (e.g. 400 = scheduled clean starting, 501 = returning to dock,
     502 = recharge). A non-zero fault is not necessarily a blocking error.
+    The converse also holds: the device can sit in its error state (``dpStatus``
+    12) with ``dpFault`` still 0 (observed live with the dust-bin module
+    removed), so a ``None``/``NONE`` fault does not imply the absence of an
+    error condition.
 
     These labels follow the ss07 app text, which differs from the Q7 ``B01Fault``
     for several shared numbers (500, 501, 503, 569, 570) -- so this is a
--- a/roborock/data/b01_q10/b01_q10_code_mappings.py
+++ b/roborock/data/b01_q10/b01_q10_code_mappings.py
@@ -313,7 +313,7 @@ class YXFault(RoborockModeEnum):
     MAIN_WHEELS_STUCK = "main_wheels_stuck", 7
     ROBOT_TRAPPED = "robot_trapped", 8  # hw-confirmed
     CHECK_DUSTBIN_FILTER = "check_dustbin_filter", 9
-    LOW_BATTERY = "low_battery", 12
+    LOW_BATTERY = "low_battery", 12  # hw-confirmed (fired at 14% mid clean)
     TEMPERATURE_THRESHOLD = "temperature_threshold", 14
     ROBOT_TILTED = "robot_tilted", 16
     LIDAR_COVER_OBSTRUCTED = "lidar_cover_obstructed", 21
--- a/roborock/data/b01_q10/b01_q10_code_mappings.py
+++ b/roborock/data/b01_q10/b01_q10_code_mappings.py
@@ -309,7 +309,7 @@ class YXFault(RoborockModeEnum):
     BUMPER_STUCK = "bumper_stuck", 2
     ROBOT_SUSPENDED = "robot_suspended", 3  # hw-confirmed (wheels lifted off the floor mid-task)
     CLIFF_SENSOR_ERROR = "cliff_sensor_error", 4  # hw-confirmed (one side lifted mid-task)
-    MAIN_BRUSH_STUCK = "main_brush_stuck", 5
+    MAIN_BRUSH_STUCK = "main_brush_stuck", 5  # hw-confirmed (two physical brush jams)
     MAIN_WHEELS_STUCK = "main_wheels_stuck", 7
     ROBOT_TRAPPED = "robot_trapped", 8  # hw-confirmed
     CHECK_DUSTBIN_FILTER = "check_dustbin_filter", 9
--- a/roborock/data/b01_q10/b01_q10_code_mappings.py
+++ b/roborock/data/b01_q10/b01_q10_code_mappings.py
@@ -290,7 +290,7 @@ class YXCarpetCleanType(RoborockModeEnum):
 
 
 class YXFault(RoborockModeEnum):
-    """Q10 (B01/ss07) ``dpFault`` (90) codes, seeded from the ss07 app's fault i18n.
+    """Q10 (B01/ss07) ``dpFault`` (90) codes, from the ss07 fault spec.
 
     ``dpFault`` is *overloaded*: several values are lifecycle/status rather than
     errors (e.g. 400 = scheduled clean starting, 501 = returning to dock,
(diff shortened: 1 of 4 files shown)


roborock/data/b01_q10/b01_q10_code_mappings.py
```diff
@@ -289,6 +289,63 @@ class YXCarpetCleanType(RoborockModeEnum):
     CROSS = "cross", 3
 
 
+class YXFault(RoborockModeEnum):
+    """Q10 (B01/ss07) ``dpFault`` (90) codes, seeded from the ss07 app's fault i18n.
```
allenporter: don't need to mention this came from any 'app', lets just say its the spec

allenporter (dismissed):

roborock/data/b01_q10/b01_q10_code_mappings.py
andrewlyeats: Changes made and new data added from today's experiments. Understood about references to spec despite auto-generated commit message that slipped thru. Thanks!

allenporter (approved):

allenporter:
Thank you for the contribution!

PR #866, opened 2026-07-06
fix: enable Qrevo Master wash/fill dock features (code 14 dock, Roborock QRevo S5A) 

Jason-Morcos:
## Summary

Adds `RoborockDockTypeCode.qrevo_master_dock` to `WASH_N_FILL_DOCK_TYPES` so dock type `14` is treated as a wash/fill dock.

I noticed that with the HA 2026.7 update, by Qrevo S5A dock stopped showing the tank entities and strainer entity.

## Why

HA 2026.7 diagnostics show a Qrevo Master dock reporting `dockType: 14`, which maps to `RoborockDockTypeCode.qrevo_master_dock`. The same dock reports wash-capable data, but `python-roborock` currently excludes it from `WASH_N_FILL_DOCK_TYPES`, so `is_wash_n_fill_dock()` returns `False`.

That causes the V1 `wash_towel_mode` trait to be skipped. Downstream in Home Assistant, `wash_towel_mode is not None` gates the clean-water, dirty-water, and strainer entities, so those entities can become stale/restored for this dock even though the underlying dock data is available.

Sanitized diagnostics comparison:

```text
dockType 14: dss=168, strainerWorkTimes=44, wash_towel_mode absent, smart_wash_params absent
dockType 15: dss=169, strainerWorkTimes=34, wash_towel_mode present, smart_wash_params present
```

## Changes

- `roborock/device_features.py`: add `qrevo_master_dock` to `WASH_N_FILL_DOCK_TYPES`.
- `tests/devices/traits/v1/test_device_features.py`: add a focused regression test proving `is_wash_n_fill_dock(RoborockDockTypeCode.qrevo_master_dock)` is true.

## Testing

```bash
uv run pytest tests/devices/traits/v1/test_device_features.py
uv run pytest tests/devices/traits/v1/test_wash_towel_mode.py tests/devices/traits/v1/test_smart_wash_params.py
uv run pre-commit run --all-files
uv run pytest
```

PR #865, opened 2026-07-06
feat: add QV 35A dock type code (21) to RoborockDockTypeCode

epinethrone:
## Summary

Adds dock type `21` — the wash/fill/dry dock shipped with the **Roborock QV 35A** (`roborock.vacuum.a168`) — to `RoborockDockTypeCode` and to `WASH_N_FILL_DOCK_TYPES`.

This mirrors #829 (which added the Qrevo S5V, `22`).

Fixes #864.

## Why

The QV 35A's dock reports `dock_type: 21`, which had no enum member, so `RoborockDockTypeCode(21)` fell back to `unknown`. That made `is_wash_n_fill_dock()` return `False`, which skips the `wash_towel_mode` trait at setup (`Trait 'wash_towel_mode' not supported, skipping`). Downstream in Home Assistant, the clean-water-box, dirty-water-box and strainer entities are gated on `wash_towel_mode is not None`, so they were permanently `unavailable` — even though the device reports the underlying data (`dss` for the tank statuses, `strainer_work_times` in the consumables payload).

Verified on a real QV 35A (firmware `02.36.38`): after adding the dock type, `is_wash_n_fill_dock` is `True` and the trait is set up as expected. Clean-fluid delivery is not supported on this dock (`is_clean_fluid_delivery_supported = False`), so the `clean_fluid_empty` entity correctly remains absent.

## Changes

- `roborock/data/v1/v1_code_mappings.py`: add `qv_35a_dock = 21`.
- `roborock/device_features.py`: add `RoborockDockTypeCode.qv_35a_dock` to `WASH_N_FILL_DOCK_TYPES`.
- Tests: a dedicated `test_qv_35a_dock_type`, plus the new dock type added to the `wash_towel_mode` / `dust_collection_mode` / `smart_wash_params` trait parametrizations — same coverage #829 added.

## Testing

```
pytest tests/data/ tests/devices/traits/v1/test_wash_towel_mode.py \
       tests/devices/traits/v1/test_dust_collection_mode.py \
       tests/devices/traits/v1/test_smart_wash_params.py \
       tests/test_supported_features.py tests/devices/traits/v1/test_device_features.py
```

All pass (snapshots unchanged).

PR #867, opened 2026-07-07
feat: add better dock feature control

Lash-L:
Replaces:  https://github.com/Python-roborock/python-roborock/pull/866 https://github.com/Python-roborock/python-roborock/pull/865

This adds more intelligent feature flag control than what we currently have been doing for dock related features. matches our other device features work.


roborock/data/v1/v1_code_mappings.py
```diff
@@ -650,22 +650,31 @@ class RoborockDockErrorCode(RoborockEnum):
...
+    k1_dock = 17
+    o6_dock = 18
+    k1c_dock = 19
+    k1s_dock = 20
+    shell_e_dock = 21
+    shell_2e_dock = 22
+    shell_3c_dock = 23
+    type_27_dock = 27
+    k1c_lite_dock = 28
+    shell_2e_lite_dock = 30
```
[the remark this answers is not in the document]
Lash-L: I think we skip this as no callers refer to the enum by its name outside of this repo.

roborock/data/v1/v1_code_mappings.py
```diff
@@ -650,22 +650,31 @@ class RoborockDockErrorCode(RoborockEnum):
 
 class RoborockDockTypeCode(RoborockEnum):
     unknown = -9999
-    no_dock = 0
```
allenporter, line 653: I believe we'll want to mark this as breaking given its used by callers (e.g. home assistant)

roborock/devices/traits/v1/__init__.py
```diff
@@ -262,9 +263,15 @@ async def discover_features(self) -> None:
         await self.device_features.refresh()
         # Dock type also acts like a device feature for some traits.
         dock_type = await self._dock_type()
+        if self.status.dss is None:
```
allenporter: want to stick all of this inside `_dock_type` since it's trying to handle all the fetching and caching of the doc type?

allenporter (changes requested):
Very nice. i didn't totally get what was going on here before so happy to see this.

roborock/data/v1/v1_code_mappings.py
Lash-L: will tag with breaking on merge! I believe I can just do a comment in the commit description and semantic will tag it. 

i.e. 
BREAKING CHANGE: RoborockDockTypeCode enum members were renamed, which may require downstream callers to update references

allenporter (approved):

PR #870, opened 2026-07-08
feat: use display name to fix duplicate code mappings

Lash-L:
This fixes a large portion of our enums that are duplicates. This will be a breaking change on HA, but should likely be okay


allenporter (dismissed):

Lash-L:
Holding off on this one until closer to the beta, don't want to put us in a weird situation where this breaks anything

Lash-L:
stale review due to conflicts. Approve when you can @allenporter want to make sure this gets into beta.

roborock/data/code_mappings.py
```diff
@@ -10,10 +10,22 @@
 class RoborockEnum(IntEnum):
     """Roborock Enum for codes with int values"""
 
+    _display_name_: str | None
+
+    def __new__(cls, value: int, display_name: str | None = None) -> Self:
+        member = int.__new__(cls, value)
+        member._value_ = value
+        member._display_name_ = display_name
```
allenporter, line 18: Weird syntax with trailing _

roborock/data/code_mappings.py
```diff
@@ -30,8 +42,13 @@ def _missing_(cls: type[Self], key) -> Self:
...
 
     @classmethod
-    def as_dict(cls: type[Self]):
-        return {i.name: i.value for i in cls if i.name != "missing"}
+    def as_dict(cls: type[Self]) -> dict[str, int]:
+        result: dict[str, int] = {}
+        for item in cls:
+            if item.name == "missing":
+                continue
+            result.setdefault(item.display_name, item.value)
```
allenporter, line 50: Can just be regular dict value setting? Don't think set default is needed

allenporter (approved):
Style issues can be fixed in follow up

PR #871, opened 2026-07-09
feat!: complete full Zeo device support

### LINKED ISSUE #833: [Feature/Bug] Zeo Washing Machine (M1S Ultra): Deep Sleep Wakeup Failure & Missing DP Mappings
### Device Information

* **Device Model:** Roborock M1S Ultra Washing Machine (Zeo / A01 protocol)
* **Integration Version:** Latest HA core `roborock` integration

### Overview

After extensive local reverse-engineering and testing on the M1S Ultra, I have identified main issues preventing full automation of this device via HA. I am providing the detailed findings and verified DP codes below to help improve both this integration and the underlying `python-roborock` library.

**Please note on priority:** **Issue 1 is the core blocker** that urgently breaks the local-control experience and requires investigation. **Issue 3 is low priority** (a historical summary of verified DP mappings) and can be addressed gradually in future updates whenever time permits.

---

### Issue 1: Deep Sleep Wakeup Failure & "Remote Control Authorization" Lock

**Description:**
When the washing machine is in a "Deep Sleep" state (completely powered off / black screen), the integration cannot send any commands.

* Sending the standard start command (`set_value(203, 4)`) does nothing.
* I injected a custom physical power-on command (`set_value(200, 1)`) via the `ZeoApi`. While this successfully turns on the machine **if it is already awake or just recently turned off**, it completely fails to wake the machine from Deep Sleep.

**CRITICAL HARDWARE CONTEXT:**
The M1S Ultra has a physical safety feature on its touch panel: it requires the user to manually authorize/enable "Remote Control".
If the machine enters deep sleep or loses network connection, it seems to automatically drop this local authorization. Once dropped, all local MQTT/LAN commands are completely ignored by the machine's firewall.

However, the **official Roborock App CAN still wake it up** from this locked state. This strongly suggests the official app uses a specific "Cloud Magic Packet" to bypass the hardware lock and re-enable the Wi-Fi module before sending actual DPs.

**Request:** Could we investigate the cloud wake-up packet for Zeo devices? This is the core priority. Without it, automations are broken because the local network drops the connection due to the physical authorization lock.

---

### Issue 2: Missing "Power On" Button Entity in HA

**Description:**
In `homeassistant/components/roborock/button.py`, the `ZEO_BUTTON_DESCRIPTIONS` array defines `start`, `pause`, and `shutdown`, but it completely lacks a `power_on` entity.

**Request:**
Please add a `power_on` button entity for `RoborockButtonDescriptionA01` that sends `DP 200: 1`. Even though local deep-sleep wakeup is currently blocked (as mentioned in Issue 1), having a dedicated Power On button in the HA UI is still crucial for waking the machine from lighter standby states without triggering a wash cycle immediately.

---

### Issue 3: Missing / Conflicting DP Mappings for M1S Ultra

**Description:**
This is a low-priority enhancement. The current underlying `zeo_code_mappings.py` uses legacy codes that conflict with the M1S Ultra. For instance, DP 205 `12` is mapped to `down_clean`, but on M1S Ultra, `12` is `tub_clean`.
Through historical packet capture on my machine, I have successfully verified the following missing DPs and modes:

**1. Detergent & Softener Types (DP 213 & 214):**
On M1S Ultra, `3` is Auto. It also supports precise milliliter selections. *(Note: `ZeoSoftenerType` shares the exact same mapping).*

**2. Additional Programs (DP 205):**
Verified missing modes include Standard (`39`), Tub Clean (`12`), Urine (`61`), Oxygen Bubble (`63`), and New Clothes (`17`).

**3. Temperatures (DP 207):**
Verified missing temperatures include `6` for absolute 20°C, and `7` for extreme high temp (90/95°C).(solved in the latest version)

**4. Hidden DPs (DP 233 & 267):**
Successfully mapped DP `233` for Soak Time control and DP `267` for Blue Oxygen / Hydrolysis switch.

(Codes merged below)

```python
class RoborockZeoProtocol(RoborockEnum):
    # ... existing ...
    SOAK_TIME = 233          # Soak time control
    BLUEOX_SETTING = 267     # Blue Oxygen / Hydrolysis switch

class ZeoDetergentType(RoborockEnum):
    empty = 0
    low = 1
    medium = 2
    auto = 3        # Auto intelligent mode (Overrides legacy 'high')
    ml_100 = 4      # 100ml
    ml_70 = 5       # 70ml
    ml_40 = 6       # 40ml
    off = 7         # Disabled
    ml_150 = 9      # 150ml

class ZeoProgram(RoborockEnum):
    tub_clean = 12           # Overrides legacy 'down_clean'
    new_clothes = 17         # New Clothes wash
    intensive = 22
    standard = 39            # Overrides legacy '1'
    urine = 61               # Urine stain removal (New)
    oxygen_bubble_wash = 63  # Oxygen Bubble wash (New)
    silk = 9

class ZeoTemperature(RoborockEnum):
    auto = 1
    low = 2
    medium = 3
    high = 4
    max = 5
    twenty_c = 6             # Verified: 20°C
    extreme = 7              # Verified: 90/95°C (solved in the latest version)

```

**Request:**
Please consider merging these missing DPs in a future update so the HA frontend can correctly display and select these modes without throwing errors.

Thank you to the contributors for the amazing work on this integration! Let me know if you need any packet captures or further testing on the M1S Ultra.

NOisi-x:
## Summary

This PR **completes Zeo (washing machine) device support** in `python-roborock`.

The library already had foundational A01 protocol support for Zeo devices, but several critical gaps prevented practical use: the start command didn't work, roughly half of the DP IDs were undefined, several device states and enum classes were missing, and most data points had no value converters. This PR closes those gaps, bringing the DP and enum definitions in line with the Roborock Washer app protocol.

With these changes, Zeo devices support the full range of operations defined by the protocol: start/stop/pause/shutdown, mode and program selection, temperature, rinse, spin, drying control, child lock, auto detergent/softener, UV sterilization, soak, silent mode, dry care, steam, ion deodorization, wash-dry linkage, and more.

## Motivation

The library's existing Zeo `set_value(START, 1)` did not actually start the device. Through protocol analysis and comparison with the Roborock Washer HomeAssistant integration, three root causes were identified:

1. **QoS level**: DP200 must be published at QoS 1; the default QoS 0 was rejected by the broker.
2. **Missing timestamp**: A01 command payloads require a `"t"` (Unix timestamp) field.
3. **Missing parameter bundle**: The device firmware requires MODE and PROGRAM to be sent together with START in a single MQTT message.

Beyond the start command, `RoborockZeoProtocol` covered only 31 of 67 known DP IDs, `zeo_code_mappings.py` was missing 4 enum classes and several enum values, and `ZEO_PROTOCOL_ENTRIES` provided converters for only 16 of the 52 user-facing DPs — meaning `query_values()` silently returned `None` for most data points.

### 1. MQTT QoS — `roborock/mqtt/roborock_session.py`

Changed `client.publish()` to use `qos=1` (at-least-once delivery).

### 2. A01 payload timestamp — `roborock/protocols/a01_protocol.py`

Added `"t": int(time.time())` to the JSON payload structure in `encode_mqtt_payload()`.

### 3. Combined start command — `roborock/devices/traits/a01/__init__.py`

`ZeoApi.set_value(START, 1)` now:
- Queries the device for its current MODE, PROGRAM, TEMP, RINSE_TIMES, SPIN_LEVEL, and DRYING_MODE
- Bundles the returned values with `{START: 1}` into a single MQTT message
- Falls back to `MODE=1, PROGRAM=1` if the query fails

### 4. Extended DP ID enum — `roborock/roborock_message.py`

`RoborockZeoProtocol` expanded from 31 to 67 entries, now covering all 52 user-facing DPs (200-266) plus meta DPs. Newly added: dirt detection (215-216), UV light (228), soak (233), total time (234), smart hosting (235-238), silent mode (240-242), dry care mode (244), wash-dry linkage (255), drying method (256), steam volume (257), ion deodorization (258), dryer start error (265), and more.

### 5. Enum updates — `roborock/data/zeo/zeo_code_mappings.py`

- `ZeoMode`: added `treatment = 4`
- `ZeoState`: expanded from 13 to 20 states
- `ZeoSoak`: added `very_max = 5`
- New enum classes: `ZeoDryingMethod`, `ZeoSteamVolume`, `ZeoDryAndCare`, `ZeoDryerStartError`

### 6. Value converters — `roborock/devices/traits/a01/__init__.py`

`ZEO_PROTOCOL_ENTRIES` expanded from 16 to 57 entries, providing typed converters for all user-facing DPs (200-266).

## Backward Compatibility

All changes are additive or transparent:

- `ZeoApi.query_values()` and `ZeoApi.set_value()` signatures unchanged
- New DP IDs and enums are pure additions
- QoS change applies to all MQTT publishes, matching the expected at-least-once delivery
~~- Start command query adds 1-2s latency; all other `set_value` calls unaffected~~

~~## Known Limitation~~

~~`sent_decoded_command` uses a strict "all requested DPs must respond" completion condition. Since A01 devices do not expose a capability probing mechanism, there is currently no clean way to determine which optional DPs (e.g., SOAK, STEAM_VOLUME) a given device supports before querying them.~~

~~As a temporary workaround, the start command queries and bundles only the following 6 core DPs, which has been verified to successfully start devices on physical Zeo hardware:~~

~~| DP | Name | Purpose |~~
~~|----|------|---------|~~
~~| 204 | MODE | Wash mode (e.g. wash, wash-and-dry, dry) |~~
~~| 205 | PROGRAM | Wash program (e.g. standard, quick, wool) |~~
~~| 207 | TEMP | Water temperature |~~
~~| 208 | RINSE_TIMES | Number of rinse cycles |~~
~~| 209 | SPIN_LEVEL | Spin speed |~~
~~| 210 | DRYING_MODE | Drying mode (e.g. quick, iron, store) |~~

~~These are bundled with `{START: 1}` and sent as a single MQTT message. The device applies its internally-stored values for any parameters not explicitly included.~~

~~A better long-term solution for dynamically detecting device capabilities is welcome — contributions or ideas in this area are appreciated.~~

## Testing

- `set_value(START, 1)` verified working on physical Zeo One and Zeo M1S devices
- `set_value` for non-START DPs (mode, program, temperature, etc.) continues to work as before
- Coordinator polling and DPS push updates unaffected
- All changes pass linting (ruff, mypy)

## Acknowledgments

This PR draws heavily on the excellent reverse-engineering work done in [ndwzy/roborock_washer_ha](https://github.com/ndwzy/roborock_washer_ha). The A01 protocol structure, DP definitions, enum mappings, and the combined start command behavior were all cross-referenced against that project's implementation. The DP ID list (200-266), state enum values, and device error codes were verified against the Roborock app bundle extracted and analyzed by that project.

---

Closes #833

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 <!-- version list -->
 
+## v5.34.0 (2026-07-20)
+
+### Chores
+
+- Delete slop md
+  ([`9ba41bb`](https://github.com/Python-roborock/python-roborock/commit/9ba41bb56ce7ac329c77e6ab985c98f31fd5fe51))
+
+### Features
+
+- Change license to apache 2.0 ([#890](https://github.com/Python-roborock/python-roborock/pull/890),
+  [`a304471`](https://github.com/Python-roborock/python-roborock/commit/a30447130241787cec8c6c75a63d35624e22702c))
+
+
 ## v5.33.0 (2026-07-19)
 
 ### Chores
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "python-roborock"
-version = "5.33.0"
+version = "5.34.0"
 description = "A package to control Roborock vacuums."
 authors = [{ name = "humbertogontijo", email = "humbertogontijo@users.noreply.github.com" }, {name="Lash-L"}, {name="allenporter"}]
 requires-python = ">=3.11, <4"
--- a/uv.lock
+++ b/uv.lock
@@ -1335,7 +1335,7 @@ wheels = [
 
 [[package]]
 name = "python-roborock"
-version = "5.33.0"
+version = "5.34.0"
 source = { editable = "." }
 dependencies = [
     { name = "aiohttp" },


Lash-L:
Hey @NOisi-x something about your changes seem to have broken tests. Tests are running indefinitely, can you take a look?

NOisi-x:
> Hey @NOisi-x something about your changes seem to have broken tests. Tests are running indefinitely, can you take a look?

Hi, @Lash-L ,I've updated the test files and all tests are now passing. Thanks！

## CI Failure Analysis and Fix Summary

Apologies for the broken CI — I'm a hobbyist with fairly limited programming experience and I overlooked the fact that the source changes would affect the existing test suite. Lesson learned for next time. Below is an analysis of what went wrong and what was changed to fix it.

### Root Causes

The CI test suite failed for three reasons:

1. **Payload format mismatch**: `encode_mqtt_payload()` now includes a `"t"` field. 4 unit test assertions used exact-match checks.

2. **FakeChannel signature**: `MqttChannel.publish()` gained a `qos` parameter. `FakeChannel._publish` did not accept it.

3. **Time-dependent snapshot**: The `"t"` field value comes from `time.time()`, making the A01 e2e snapshot non-deterministic across CI runs.

### Changes Made

#### `tests/devices/traits/a01/test_init.py` — 4 assertions updated

Replaced exact-match with key-check: `payload_data["dps"] == {...}` + `assert "t" in payload_data`.

#### `roborock/testing/channel.py` — 1 signature update

`FakeChannel._publish` now accepts `qos: int = 0` to match the updated `MqttChannel.publish` signature.

#### `tests/e2e/test_device_manager.py` — time frozen for A01 test

Added `from freezegun import freeze_time` and decorated `test_a01_device` with `@freeze_time("2025-01-20T12:00:00")` to ensure deterministic MQTT payloads across CI runs. Snapshot regenerated via `--snapshot-update`.

#### `roborock/devices/traits/a01/__init__.py` — 1 import fix

Removed stray blank line between `RoborockException` and `roborock.data` imports (ruff/isort ordering).

### Tests Not Affected

- `tests/protocols/test_a01_protocol.py` — `decode_rpc_response()` ignores the `"t"` field
- `tests/e2e/test_mqtt_session.py` — no client-side publish affected
- `tests/devices/test_a01_traits.py` — patches `send_decoded_command` entirely
- All V1/B01/Q7/Q10/Dyad tests — no changes needed

roborock/devices/transport/mqtt_channel.py
```diff
@@ -89,19 +89,23 @@ async def subscribe_stream(self) -> AsyncGenerator[RoborockMessage, None]:
         finally:
             unsub()
 
-    async def publish(self, message: RoborockMessage) -> None:
+    async def publish(self, message: RoborockMessage, qos: int = 0) -> None:
```
allenporter: Let's define an enum for qos level. Just having numbers 0, 1, 2 doesn't really say what these are, or why they exist.  Why would someone set a lower or higher level, etc should be explained in the code and doing it from the enum could be a good way.

roborock/data/zeo/zeo_code_mappings.py
```diff
@@ -19,8 +20,16 @@ class ZeoState(RoborockEnum):
     cooling = 8
     under_delay_start = 9
     done = 10
-    aftercare = 12
-    waiting_for_aftercare = 13
+    updating = 11
```
allenporter: let's tag as a breaking change. i think it's that commits have `feat!:` ?

roborock/devices/traits/a01/__init__.py
```diff
@@ -176,8 +228,47 @@ async def query_values(self, protocols: list[RoborockZeoProtocol]) -> dict[Robor
 
     async def set_value(self, protocol: RoborockZeoProtocol, value: Any) -> dict[RoborockZeoProtocol, Any]:
         """Set a value for a specific protocol on the device."""
-        params = {protocol: value}
-        return await send_decoded_command(self._channel, params, value_encoder=lambda x: x)
+        params: dict[RoborockZeoProtocol, Any] = {protocol: value}
+        if protocol == RoborockZeoProtocol.START and value == 1:
```
allenporter: This makes me think it may be time to define actual API methods for these specialized commands rather than a single generic query_values/set_values (e.g. a "start" and "stop" methods etc for the washing machine.

Do you expect there to be more specialized commands like this or is this the only one?

At a minimum: Split out the function that queries the current values to a separate method

roborock/devices/traits/a01/__init__.py
```diff
@@ -176,8 +228,47 @@ async def query_values(self, protocols: list[RoborockZeoProtocol]) -> dict[Robor
...
+                    (RoborockZeoProtocol.SPIN_LEVEL, None),
+                    (RoborockZeoProtocol.DRYING_MODE, None),
+                ):
+                    if dp in current:
+                        params[dp] = current[dp]
+                    elif fallback is not None:
+                        params[dp] = fallback
+            except RoborockException:
+                _LOGGER.warning(
+                    "Failed to query device state before start, using defaults",
```
allenporter: i'm not sure this fallback is a great idea e.g. if the user is expecting some values to be preserved then silently not preserving them worries me e.g. as a user  I would be concerned if my washing machine changed my water temp due to flaky wifi and i might prefer it just fail rather than shrinking cotton clothes or whatever.

Can we either remove or provide some rationale here why this is expected or ok?

tests/e2e/test_device_manager.py
```diff
@@ -521,6 +522,8 @@ async def test_a01_device(
     snapshot: syrupy.SnapshotAssertion,
 ) -> None:
     """Test the device manager end to end flow with an A01 device."""
+    # Use a fixed timestamp so the MQTT payload snapshot is deterministic.
+    monkeypatch.setattr("roborock.protocols.a01_protocol.time.time", lambda: 1755750947.0)
```
allenporter: We have `deterministic_message_fixtures`
https://github.com/Python-roborock/python-roborock/blob/89c80743e2c560f36f56e0c17c85c0d14544f406/tests/fixtures/logging_fixtures.py#L53 where these kinds of patches can live

allenporter (changes requested):
Fantastic contribution, thank you so much for the improvements here!

roborock/devices/transport/mqtt_channel.py
NOisi-x: I add an enumeration for the following three QoS levels:

AT_MOST_ONCE (0) — Fire-and-forget, no acknowledgment.
AT_LEAST_ONCE (1) — Broker sends PUBACK, guaranteed delivery.
EXACTLY_ONCE (2) — Four-step handshake, no duplicates.

These values are defined by the MQTT 3.1.1 / 5.0 protocol specification (OASIS standard).

roborock/data/zeo/zeo_code_mappings.py
NOisi-x: I change the title. Is that right?

roborock/devices/traits/a01/__init__.py
NOisi-x: The query logic has been split into `_get_current_params()` as requested. Based on testing with physical Zeo devices, I believe START is the only command that requires parameter bundling.

**START (200) will wake up the device.** The Zeo firmware requires MODE and PROGRAM to be present in the same MQTT message as the start signal because the device needs to know *what* to do when it wakes up. If the device is **already manually powered on** (via the physical power button), even START(200) can be sent as a simple `set_value` with QoS 0, and commands like PAUSE (201) and SHUTDOWN (202) work without any bundling at all.

Through testing with the actual Roborock app, I confirmed the DPs fall into two distinct categories:

- **Immediate DPs** (sound, child lock, UV light, etc) — set independently via `set_value` and take effect immediately.
- **Deferred DPs** (mode, program, temperature, rinse, spin, drying, etc) — the app collects user selections locally and only sends them bundled with the START command. The device internally caches these values — either factory defaults or the user's last selection depending on whether the device model supports parameter persistence — which is why `_get_current_params()` can query them at start time.

The current query+bundle approach in `_get_current_params()` is functional and has been verified on physical hardware. However, in the broader HomeAssistant ecosystem similar devices (thermostats, climate controls) use a **coordinator caching pattern**: entity state changes update a local value immediately but don't push to the device; all accumulated changes are sent as a single payload when the user triggers an action (e.g. "set temperature"). Implementing this pattern for Zeo devices would require changes to the HA integration layer and is probably best left for a separate PR. I would greatly appreciate it if someone could help make thses improvements.

roborock/mqtt/session.py
```diff
@@ -3,13 +3,34 @@
...
 
 DEFAULT_TIMEOUT = 30.0
 
+
+class MqttQos(IntEnum):
+    """MQTT Quality of Service levels.
+
+    Determines the delivery guarantee for published messages:
+
+    - AT_MOST_ONCE (0): Fire-and-forget. No acknowledgment required.
```
allenporter: You can move these to the actual enums like:

```
AT_MOST_ONCE = 0
"""Fire-and-forget. No acknowledgment required.""""
```

roborock/devices/traits/a01/__init__.py
allenporter: This is helpful to understand the end state you recommend.

We can do a lot of that in this library without large changes to Home Assistant. The way we do this in other device types is through traits that (a) define typed interfaces for data (b) handle commands related to that data/state and (c) hold on to the latest state with functions for refreshing. So we won't solve all of that now in this PR, but here is what we can do:

- Define a dataclass with the arguments to start that are required (mode, program, etc)
- Define a method (can be private) that can query and return those
- Define a separate method `start` that users will call that handles the now logic for querying and calling the start command

We can work on the rest in the future.

roborock/data/zeo/zeo_code_mappings.py
allenporter: I think so yeah.

For context:  This is currently used by Home Assistant for example, so it can't be merged without changing the translation strings there.  Another way to do it could be to leave the enum alone, add a comment on the enum, then later go add the translations in home assistant.

allenporter (changes requested):

tests/e2e/test_device_manager.py
NOisi-x: I move A01 timestamp patch into deterministic_message_fixtures.

roborock/mqtt/session.py
NOisi-x: done

roborock/devices/traits/a01/__init__.py
NOisi-x: Thanks for the explanation and guidance — that was really helpful for understanding this project. I've updated the code per your suggestions.

roborock/devices/traits/a01/__init__.py
NOisi-x: Your concern was completely valid — I've removed the fallback.

roborock/data/zeo/zeo_code_mappings.py
NOisi-x: The names `smart_hosting` / `smart_hosting_waiting` come from reverse-engineering the official Roborock app bundle. The previous names `aftercare` / `waiting_for_aftercare` were also from my PR — I translated them from the Chinese status labels shown in the app UI. I've reverted to the old names and added a comment noting the canonical app-bundle names.

NOisi-x:
@allenporter Thank you very much for your code review. As I spent more time analyzing the files obtained through reverse engineering, I made more exciting discoveries. I'll temporarily convert this PR to draft status and will continue to refine it. Please review it again once it's ready—thanks again!

NOisi-x:
After fully reverse-engineering the official Roborock washer app plugin bundle (extracted from [index.ios.bundle](https://github.com/ndwzy/roborock_washer_ha/blob/main/custom_components/roborock_washer/plugins/8f1d582973de4e41a73fc2ea305e17da/index.ios.bundle)), all Zeo device DP mappings, feature-bit definitions, and wire-format behaviours have been implemented with strict fidelity to the bundle's logic.

**Testing disclaimer**: I only own 2 washer devices, and my washers does not support every capability (e.g. certain voice DPs, and some series-specific behaviours).  However, all features that *are* covered by my hardware have been confirmed working.  The remaining logic mirrors the bundle exactly and should be correct, but has not been physically validated.

---

## Summary

Complete overhaul of the Zeo (washing machine / dryer) A01 protocol support, adding feature discovery, custom programme management, dryer support, delayed start, silent mode, cleanser configuration, management utilities, and a reorganised protocol enum with semantic sections.

---

## Bundle Supported Devices (56 total)

The list below reflects the models explicitly referenced in the plugin bundle as of extraction.  The bundle may be updated in the future to include additional models.  Any device not listed here should still function correctly for all **basic** operations (start/pause/stop, mode & program selection, temperature, spin, rinse, drying mode).

### 🧺 Washers — 52 models

| Series | Models |
|--------|--------|
| **H1** | a63 H1, a114 H1C, a102 H1 (Global), a90 H1Lite, a91 H1Lite (Global), a237 H1 Lite (Rev2), a242 H1C (Global), a115 H1C, a126-128 H1 (JP/TW/KR - deprecated) |
| **M1** | a92 M1, a93 M1 (Global), a133 M1Lite, a277 M1Lite (Rev3), a162 M1Lite (Global), a233 M1Lite Plus, a276 M1Lite Plus (Rev2), a234 M1 Plus, a218 M1 (Rev2), a129-131 M1 (TW/JP/KR - deprecated) |
| **Hyperion** | a141 Hyperion, a149 HyperionPro, a207 Hyperion Plus, a230 Hyperion (Global) |
| **Muse / Metis** | a142 Muse, a215 Muse (Global), a154 Metis, a214 Metis (Global) |
| **Hera** | a227 Hera, a261 (DE), a273 (NO), a268 (KR), a269 (TW) |
| **Pandora** | a239 Pandora, a262 (DE), a272 (NO), a270 (KR), a271 (TW) |
| **Halia** | a240 Halia, a241 Halia Lite |
| **Poseidon** | a180 Pro, a181, a201 Pro+, a255 (Global) |
| **Other** | a105 Couple, a191 Crius, a192 Coeus, a211 Mitty, a213 Medusa |

### 👕 Dryers — 4 models

| Model | Name | Region |
|-------|------|--------|
| a188 | Apollo Pro+ | CN |
| a265 | Apollo Pro+ | Global (US/EU) |
| a204 | Apollo Pro | CN |
| a258 | Apollo Pro | Global (US/EU) |

---

#### New methods

| Method | Description |
|--------|-------------|
| `resume()` | Sends only `START=1` — matches official app's `continue()` |
| `start_with_preset(minutes)` | Delayed start with countdown |
| `set_silent_mode(on, start_h, start_m, end_h, end_m)` | Bundled 3-DP set (240/241/242) |
| `set_cleanser_config(det_exp, sof_exp, det_type, sof_type)` | Bundled 4-DP set |
| `get_custom_mode()` | Returns `ZeoCustomMode` (washer) or `ZeoDryerCustomMode` (dryer) from DP 222 |
| `load_cloud_program()` | Sends CUSTOM_PARAM_GET=1 |
| `save_cloud_program(...)` | Saves cloud programme with optional overrides |
| `save_panel_program(...)` | Saves panel onboard programme |
| `get_auto_detergent()` / `get_auto_softener()` | Detects dispense mode (legacy vs addition-type-control) |
| `check_fcc_state()` | Queries DP 10001 (F_C) |
| `load_general_info()` | Queries DP 10005 (robotInfo) |
| `upload_log()` | RPC call with random request ID |
| `sync_privacy_to_device(agreed)` | Pushes privacy agreement to DP 10006 |
| `update_sound_package_info()` | Queries DP 10004 (SND_STATE) |
| `force_load()` / `load_feature_dps()` | Two-phase polling mirroring `WasherDpsManager` |
| `silent_mode_start_hm()` / `silent_mode_end_hm()` | Cached minute-of-day → (hour, minute) |
| `serial_number` / `network_info` / `is_fcc_state_legal` | Derived read-only properties |

#### `ZeoCustomMode` / `ZeoDryerCustomMode` dataclasses
- Washer (9 fields, bits 0–30): program (8bit), mode (2bit), temperature, rinse, spin, dry, soak, dry_care_mode, steam_volume (3bit each).
- Dryer (5 fields, bits 0–19): program (8bit), mode (3bit — absorbed temperature bit 10), dry, dry_method, steam_volume (3bit each).
- Both include `total_time_min` from DP 239.


---

## Breaking Changes

- **`RPC_RESp` → `RPC_RESPONSE`**: Old enum member name removed. No internal references existed.
- **`ZeoApi.__init__()` signature**: Added 3 keyword-only boolean parameters (`is_dryer`, `is_hyperion_halia_hera`, `is_m1_muse_metis`). All default to `False` — direct callers without keyword args are unaffected.
- **ZEO_PROTOCOL_ENTRIES reorganisation**: `FLUFF_CLEANED` moved to read-write, `APP_AUTHORIZATION` / `SMART_HOSTING_TIME` moved to read-only. Behaviour for those DPs is now correct per bundle.

allenporter (changes requested):
Looks like we have a lot of great additions here, though now the PR is huge: It's not only just changing an enum and adding a few enum methods but adding state and a ton of new traits. In my previous comments I was thinking those can be follow up improvements.

Can we please make these independent changes just to make faster to review? A huge PR like this takes awhile to set aside time to review.

Thank you for these awesome functionality improvements!

roborock/devices/traits/a01/__init__.py
```diff
@@ -151,33 +301,1121 @@ async def query_values(self, protocols: list[RoborockDyadDataProtocol]) -> dict[
...
+        """
+        # Booleans are serialised as "True"/"False" (DPBoolean) on the wire.
+        encoder = to_dp_bool if protocol in _DYAD_BOOLEAN_PROTOCOLS else None
         params = {protocol: value}
-        return await send_decoded_command(self._channel, params)
+        return await send_decoded_command(self._channel, params, value_encoder=encoder)
+
+
+@dataclass
+class ZeoStartParams:
```
Lash-L: ideally, let's use our custom objects that we use elsewhere

RoborockBase

RoborockEnum

RoborockModeEnum

depending on the usecase. These should also be in the data containers package to keep things maintainable in the traits file

roborock/devices/traits/a01/__init__.py
```diff
@@ -151,33 +301,1121 @@ async def query_values(self, protocols: list[RoborockDyadDataProtocol]) -> dict[
...
+            is_hyperion_halia_hera: ``True`` for Hyperion/Halia/Hera.
+                These series use a different detergent‑type auto‑enable
+                protocol (type=0 sends *only* AutoDetergent=False,
+                skipping DetergentType entirely).
+            is_m1_muse_metis: ``True`` for M1/Muse/Metis series.
+                These series skip softener‑related DPs in force_load
+                (they do not have a softener compartment).
+        """
         self._channel = channel
+        self._feature_trait = ZeoFeatureTrait(channel)
```
Lash-L: I'd think about seeing if it makes sense to split these up into multiple different traits. It has been a while since I have touched the Zeo logic so I don't remember everything, it might not make sense, but it would be a good point of logic separation. 

If anything uses its own command to get data, it should be its own trait.

roborock/devices/traits/a01/__init__.py
Lash-L: A good first PR to split up would be just the containers and data mapping, it's a very easy PR to review and will dramatically get your line count down. It's okay if you add them in a PR and it doesn't immediately get used.

Lash-L:
Awesome stuff! Echoing Allen, if you can split this up it will make it much easier, I left some comments on what I think will be easy separations.

NOisi-x:
@allenporter @Lash-L Thank you so much for the thorough review — I really appreciate the time both of you put into this.

You're absolutely right: I got carried away after discovering and fully parsing the plugin bundle and dumped everything at once without thinking about reviewability. That was poor PR hygiene on my part — I apologize for making this harder than it needed to be.

@Lash-L your architectural suggestions make a lot of sense. I plan to break this down as follows:

1. **Protocol fixes** (QoS + timestamp) — the atomic prerequisites
2. **DP enum & data containers** — extending `RoborockZeoProtocol`, enum classes, and moving `ZeoStartParams`/`ZeoCustomMode` into `zeo_containers.py` inheriting from `RoborockBase`
3. **Core ZeoApi** — value converters and the combined start command
4. **Individual trait PRs** — splitting `set_silent_mode`, custom mode management, dryer support, etc. into separate trait files under `a01/`, each in its own PR

I'll re-submit #1 within 24 hours, and the rest will gradually follow as standalone PRs with dependency notes in the descriptions.

Thanks again for maintaining this project — your patience is much appreciated!

PR #873, opened 2026-07-12
feat: truncate large payloads in RoborockMessage __repr__

allenporter:
This PR introduces a custom `__repr__` implementation for `RoborockMessage` that truncates payloads (bytes, bytearray, str) exceeding 1024 bytes (defined by `MAX_PAYLOAD_REPR_LEN`). This prevents large map binary messages from bloating and clogging the debug logs.

### DIFF
--- a/roborock/roborock_message.py
+++ b/roborock/roborock_message.py
@@ -274,6 +274,9 @@ ROBOROCK_DATA_CONSUMABLE_PROTOCOL = [
 ]
 
 
+MAX_PAYLOAD_REPR_LEN = 1024
+
+
 @dataclass
 class RoborockMessage:
     protocol: RoborockMessageProtocol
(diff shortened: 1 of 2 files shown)


roborock/roborock_message.py
```diff
@@ -282,3 +285,18 @@ class RoborockMessage:
     version: bytes = b"1.0"
     random: int = field(default_factory=lambda: get_next_int(10000, 99999))
     timestamp: int = field(default_factory=lambda: get_timestamp())
+
+    def __repr__(self) -> str:
+        payload_repr = "None"
```
Lash-L, line 290: can we add a little commenting here? Just to make it a bit easier to quickly see what our current form is if we need to change?

Lash-L: eh actually jk. string at the bottom is probably sufficient

Lash-L (approved):

Issue #874, opened 2026-07-12
Add status fields to StatusField enum


allenporter:
# Feature Request: Add status fields to `StatusField` enum

## Background
In the Home Assistant integration for Roborock, we currently check device capabilities for several status-derived sensors using various custom properties, traits, or check functions (e.g., checking if `wash_towel_mode` or `is_support_water_mode` is enabled).

To make these capability checks clean and uniform, we propose adding the corresponding fields from `StatusV2` to the `StatusField` enum, allowing them to be checked directly using `api.device_features.is_field_supported(StatusV2, StatusField.FIELD_NAME)`.

## Proposed `StatusField` Additions

We would like to add the following fields (defined in `StatusV2`) to the `StatusField` enum in `roborock/data/v1/v1_containers.py`:

| StatusField Candidate | `StatusV2` Dataclass Field | Current check proxy in Home Assistant |
| :--- | :--- | :--- |
| `WATER_BOX_CARRIAGE_STATUS` | `water_box_carriage_status` | `api.device_features.is_support_water_mode` |
| `WATER_BOX_STATUS` | `water_box_status` | `api.device_features.is_support_water_mode` |
| `WATER_SHORTAGE_STATUS` | `water_shortage_status` | `api.device_features.is_support_water_mode` |
| `DIRTY_WATER_BOX_STATUS` | `dirty_water_box_status` | `api.wash_towel_mode is not None` |
| `CLEAR_WATER_BOX_STATUS` | `clear_water_box_status` | `api.wash_towel_mode is not None` |
| `CLEAN_FLUID_STATUS` | `clean_fluid_status` | `api.wash_towel_mode is not None` & `api.device_features.is_clean_fluid_delivery_supported` |
| `CLEAN_PERCENT` | `clean_percent` | `api.device_features.is_support_clean_estimate` |
| `DOCK_ERROR_STATUS` | `dock_error_status` | `api.dust_collection_mode is not None` |
| `RDT` / `DRYING_TIME_REMAINING` | `rdt` | `api.device_features.is_supported_drying` |

By adding these fields to the enum and mapping them to their corresponding device features, dps protocols, IDs, or other internal checks, we can simplify entity registration and capability checking across the Home Assistant integration.

allenporter:
Note: Some of the above are incorrect: is_clean_fluid_delivery_supported vs is_clean_fluid_auto_delivery_supported (which we claim is correct)

Lash-L:
[...]

PR #875, opened 2026-07-12
feat: Implement device capability metadata for status computed property fields

allenporter:
This PR implements capability metadata on Status and StatusV2 computed properties (using a stacked @field_metadata decorator) and resolves all merge conflicts with upstream.

Issue: #874 874


roborock/data/v1/v1_containers.py
```diff
@@ -339,7 +353,8 @@ class StatusV2(RoborockBase):
     debug_mode: int | None = None
     collision_avoid_status: int | None = None
     switch_map_mode: int | None = None
-    dock_error_status: RoborockDockErrorCode | None = None
+    dock_error_status: RoborockDockErrorCode | None = field(default=None, metadata={"dock_feature": "is_collectable"})
```
Lash-L: i think this should be gated on has dock, not is collectable. non-collectable docks can error

Lash-L (changes requested):

Lash-L (approved):

PR #880, opened 2026-07-19
feat: add photo for obstacle avoidance

Lash-L:
This is kind of a fun one, it will show the photos from the vacuum when it sees obstacles (if the vac supports it).

Let me know if you want me to split it up into a few different smaller PRs there are a couple of quick and easy PRs i can break out if you'd like

HA will have to call this and cache images appropriately and gate on features.


roborock/devices/traits/v1/obstacle_photos.py
```diff
@@ -0,0 +1,112 @@
...
+        if not isinstance(response, list) or not response or not isinstance(response[0], int):
+            raise RoborockException("get_camera_status response did not contain camera status")
+        return bool((response[0] >> _MAP_OBJECT_PHOTO_ENABLED_BIT) & 1)
+
+    async def get_photo(self, photo_id: str, photo_type: int = _PHOTO_TYPE_SMALL) -> ObstaclePhoto:
+        """Fetch an obstacle photo by its map photo id."""
+        public_key = await self._standard_rpc_channel.send_command(RoborockCommand.GET_RANDOM_PKEY)
+        if not isinstance(public_key, dict) or not isinstance(public_key.get("pub_key"), dict):
+            raise RoborockException("get_random_pkey response did not contain a public key")
+        security = self._security_data.to_dict()["security"]
```
allenporter: Can we put this behind the blob rpc channel?

That is, it can handle wrapping the payload adding necessary lower level security information given a higher level command + params. Or there can be two things, a blob channel and a blob rpc channel, if you want to keep the internal layers separate.

allenporter (changes requested):
Very cool, high level i have one starting comment about protocol layering. The idea behind some of these layers was to not have to pass security data everywhere.

roborock/devices/traits/v1/obstacle_photos.py
Lash-L: How is that?

roborock/devices/traits/v1/obstacle_photos.py
```diff
@@ -0,0 +1,99 @@
...
+
+    async def get_enabled(self) -> bool:
+        """Return whether map object photo capture is enabled on the vacuum."""
+        response = await self._standard_rpc_channel.send_command(RoborockCommand.GET_CAMERA_STATUS)
+        if not isinstance(response, list) or not response or not isinstance(response[0], int):
+            raise RoborockException("get_camera_status response did not contain camera status")
+        return bool((response[0] >> _MAP_OBJECT_PHOTO_ENABLED_BIT) & 1)
+
+    async def get_photo(self, photo_id: str, photo_type: int = _PHOTO_TYPE_SMALL) -> ObstaclePhoto:
+        """Fetch an obstacle photo by its map photo id."""
```
allenporter: can you include details in the get photo pydoc about how you obtain these photo ids or photo types? (should photo type be an enum, or not exposed at all?)

roborock/protocols/v1_protocol.py
```diff
@@ -261,6 +262,48 @@ class MapResponse:
...
+            return None
+        request_id = int.from_bytes(payload[8:12], "little")
+        header_size = int.from_bytes(payload[16:18], "little")
+        payload_size = int.from_bytes(payload[20:24], "little")
+        end_offset = header_size + payload_size
+        if header_size < 24 or end_offset > len(payload):
+            raise RoborockException("Invalid V1 blob response format")
+        try:
+            data = Utils.decompress(payload[header_size:end_offset])
+        except Exception as err:
```
allenporter: i think we can probably make this more narrow to just exceptions that we're expecting here

allenporter (dismissed):

allenporter (approved):

PR #881, opened 2026-07-19
feat: decode Q10 map geometry metadata

tubededentifrice:
## Summary

Extracts the low-level, independently reviewable map decoding work from #848. The original PR remains open and unchanged as a reference.

## Changes

- Decode Q10 grid-header calibration metadata and trace heading.
- Decode erase zones and carpet masks from map packet tails.
- Correct virtual-wall framing and axis ordering, with live-capture coverage.
- Add device-agnostic fixed-origin calibration support and Q10 grid decomposition helpers.

This PR intentionally contains no trait, API orchestration, rendering-composition, or CLI changes.

## Validation

- `uv run pytest tests/map/test_b01_q10_map_parser.py tests/devices/traits/b01/q10/test_map.py` — 39 passed
- The complete split stack passes `uv run pytest` — 802 passed
- `uv run pre-commit run --show-diff-on-failure --color=always --all-files`

## Stack

1. This low-level decoding PR
2. Pure Q10 renderer
3. Low-level DPS trait + high-level map content trait

Q7 reuse is proposed independently and does not block the Q10 stack.


roborock/map/b01_q10_overlays.py
```diff
@@ -83,12 +91,67 @@ def parse_zone_blob(data: bytes | str | None) -> list[Q10Zone]:
     return zones
 
 
+_WALL_RECORD_SIZE = 8  # two (x, y) int16-BE endpoints
+
+
+def parse_virtual_wall_blob(data: bytes | str | None) -> list[Q10Zone]:
```
allenporter, line 97: Can we pick either bytes or str and not both?

roborock/map/b01_grid_layers.py
```diff
@@ -200,6 +200,62 @@ def solve_calibration(
...
+    than a dense clean. Returns ``None`` if no candidate lands a ``min_on_floor``
+    fraction of points on floor (e.g. the origin or points are bogus).
+    """
+    if not points:
+        return None
+    w, h = layers.width, layers.height
+    ox, oy = origin
+    classify = layers.classifier
+    # 1 = floor, 2 = wall/background (blocked), 0 = other. Index by cell.
+    klass = bytes(
```
allenporter, line 228: how about (1) lets go ahead and unroll this since its pretty hard to read as one line and (2) Make enums or constants to self document these magic numbers?

(When i read this line klass looks like a single byte, but then below i see it being accessed like a 2d array, so having trouble following)

allenporter: Ah I just read the other code and see this is the same. OK maybe we just do the cleanup later.

roborock/map/b01_q10_map_parser.py
```diff
@@ -23,18 +23,54 @@
...
+    if value == _BACKGROUND_VALUE:
+        return LAYER_BACKGROUND
+    if value == _UNSEGMENTED_FLOOR_VALUE:
+        return LAYER_FLOOR
+    if value >= _WALL_THRESHOLD:
+        return LAYER_WALL
+    return LAYER_FLOOR
+
+
+def decompose_layers(packet: "Q10MapPacket") -> GridLayers:
```
allenporter, line 67: Does this work as a `@property` on `Q10MapPacket`?

tests/map/test_b01_grid_layers.py
```diff
@@ -11,7 +13,30 @@
...
+        (0, "unknown"),
+        (8, LAYER_FLOOR),
+        (12, LAYER_FLOOR),
+        (240, LAYER_FLOOR),
+        (243, LAYER_BACKGROUND),
+        (249, LAYER_WALL),
+    ],
+)
+def test_classify_q10_cell(value: int, expected: str) -> None:
+    assert classify_q10_cell(value) == expected
```
allenporter, line 39: This function is in the map module, not the grid layers module so testing for it should be in the map module. Here and below with decompose_layers as well

allenporter (changes requested):

allenporter (approved):

allenporter:
I'm going to merge and we can tackle cleanups in additional follow up PRs.

PR #883, opened 2026-07-19
feat!: remove the old version of Status

Lash-L:
The old version of Status is no longer needed.

This is a breaking change for any callers, HA only calls this in one location for tests and I will do a update there before the dependency bump.


tests/data/v1/test_v1_containers.py
```diff
@@ -94,9 +90,9 @@ def test_status():
     assert s.charge_status == RoborockChargeStatus.charging
     assert s.unsave_map_reason == 0
     assert s.unsave_map_flag == 0
-    assert s.fan_power == RoborockFanSpeedS7MaxV.balanced
-    assert s.mop_mode == RoborockMopModeS7.standard
-    assert s.water_box_mode == RoborockMopIntensityS7.intense
+    assert s.fan_power == 102
+    assert s.mop_mode == 300
```
allenporter, line 94: useful to have an enum here still?

allenporter (changes requested):

tests/data/v1/test_v1_containers.py
allenporter: I see its not an enum in the V2. OK, can be improved later.

allenporter (approved):

PR #885, opened 2026-07-19
feat: compose Q10 maps in a pure renderer

tubededentifrice:
## Summary

Next PR in the #848 split, following #881 and #884. This adds the pure Q10 map-composition layer on top of the merged low-level decoding and cleanup work.

## Source lifecycle and API

`render_q10_map` accepts the three independently updated source snapshots plus static config:

- `Q10MapPacket` — latest `01 01` map-protocol response; owns grid layers, header calibration, rooms, and erase zones.
- `Q10TracePacket | None` — latest `02 01` trace-protocol response; owns path points, robot position, and heading.
- `Q10MapOverlays` — latest decoded restricted-zone and virtual-wall snapshot from DPS.
- `B01Q10MapParserConfig` — static parser/render configuration.

It returns one fully composed PNG. Calibration, layers, projected `MapData`, and drawing state remain internal; there is no separate `Q10MapRender` data model or alternate `draw_path_on_map` image path.

## Changes

- Derive calibration internally from the map and trace packets.
- Apply erase zones before rasterization.
- Project and draw trace path/position/heading and DPS overlays in one render operation.
- Keep all calibration, placement, and pixel work out of device traits.
- Cover the source boundary and observable image behavior directly in map-module tests.

This PR intentionally contains no device-trait, subscription, or CLI orchestration changes. Those remain in the next stacked PR.

## Validation

- Focused map suite — 73 passed
- `pytest -q` — 790 passed, 86 snapshots passed
- `pre-commit run --show-diff-on-failure --color=always --all-files` — all hooks passed, including ruff and mypy

PR #886, opened 2026-07-19
feat: compose Q10 map content from grouped traits

tubededentifrice:
## Summary

Completes the Q10 map integration on top of #885 with a small low-level/high-level trait boundary matching the three incoming data streams.

- `MapDpsTrait` owns the restricted-zone and virtual-wall DPS values and exposes their decoded overlays.
- `MapContentTrait` retains the latest map and trace packets and calls the pure `render_q10_map` function whenever map, trace, or overlay data changes.
- The high-level trait exposes only caller-facing map data: the composed image, rooms, path, robot position, and heading.
- Calibration, layers, projected map data, and other rendering intermediates stay inside the renderer.
- `Q10PropertiesApi` wires the DPS trait into both DPS fanout and the dependent map-content trait.

This also removes the stacked path-specific CLI command, since the normal `image_content` is already the single composed output.

## Validation

- `pre-commit run --show-diff-on-failure --color=always --all-files`
- `pytest -q` — 797 passed, 86 snapshots passed

## Stack

- #885 — merged pure Q10 renderer
- This PR — trait ownership and integration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,35 @@
 
 <!-- version list -->
 
+## v5.33.0 (2026-07-19)
+
+### Chores
+
+- Remove the old version of Status
+  ([#883](https://github.com/Python-roborock/python-roborock/pull/883),
+  [`01d0cc9`](https://github.com/Python-roborock/python-roborock/commit/01d0cc9722a66ed84390bae9e67f606d378123e4))
+
+### Features
+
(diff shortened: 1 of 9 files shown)


tubededentifrice:
Superseded by #887. This replacement uses the identical tested three-file tree on a fresh branch from merged , avoiding both the inherited stacked commit-message failure and any force-push/history rewrite.

PR #887, opened 2026-07-19
feat: compose Q10 map content from grouped traits

tubededentifrice:
## Summary

Completes the Q10 map integration on top of #885 with a small low-level/high-level trait boundary matching the three incoming data streams.

- `MapDpsTrait` owns restricted-zone and virtual-wall DPS values and exposes their decoded overlays.
- `MapContentTrait` explicitly depends on that DPS trait, retains the latest map and trace packets, and calls the pure `render_q10_map` function whenever renderable source data changes.
- The high-level trait exposes only caller-facing map data: the composed image, rooms, path, robot position, and heading.
- Calibration, layers, projected map data, and other rendering intermediates stay inside the renderer.
- A rendering failure clears the derived image instead of pairing stale image bytes with newer source state.
- DPS updates received before any map are retained by `MapDpsTrait` without emitting a no-op high-level map notification.
- `Q10PropertiesApi` wires the DPS trait into both DPS fanout and the dependent map-content trait.

The redundant path-specific CLI command from the staging stack is intentionally excluded, since normal `image_content` is already the single composed output. Trait tests cover source ownership and renderer handoff; calibration and pixel behavior remain in the map-module renderer tests.

## Validation

- `pre-commit run --show-diff-on-failure --color=always --all-files`
- focused map/trait suite — 87 passed
- `pytest -q` — 796 passed, 86 snapshots passed

## Stack

- #885 — merged pure Q10 renderer
- This PR — trait ownership and integration

### DIFF
--- a/roborock/devices/traits/b01/q10/__init__.py
+++ b/roborock/devices/traits/b01/q10/__init__.py
@@ -16,7 +16,7 @@ from .command import CommandTrait
 from .consumable import ConsumableTrait
 from .do_not_disturb import DoNotDisturbTrait
 from .dust_collection import DustCollectionTrait
-from .map import MapContentTrait
+from .map import MapContentTrait, MapDpsTrait
 from .network_info import NetworkInfoTrait
 from .remote import RemoteTrait
 from .status import StatusTrait
@@ -32,6 +32,7 @@ __all__ = [
     "DoNotDisturbTrait",
     "DustCollectionTrait",
     "MapContentTrait",
+    "MapDpsTrait",
     "NetworkInfoTrait",
     "SoundVolumeTrait",
     "StatusTrait",
@@ -79,6 +80,9 @@ class Q10PropertiesApi(Trait):
     map: MapContentTrait
     """Trait for fetching the current parsed map (image + rooms)."""
 
+    map_dps: MapDpsTrait
+    """Low-level DPS values used to compose map overlays."""
+
     clean_history: CleanHistoryTrait
     """Trait for fetching the device clean-record history (``dpCleanRecord``)."""
 
@@ -96,7 +100,8 @@ class Q10PropertiesApi(Trait):
         self.button_light = ButtonLightTrait(self.command)
         self.network_info = NetworkInfoTrait()
         self.consumable = ConsumableTrait()
-        self.map = MapContentTrait()
+        self.map_dps = MapDpsTrait()
+        self.map = MapContentTrait(self.map_dps)
         self.clean_history = CleanHistoryTrait(self.command)
         # Read-model traits updated from the device's DPS push stream.
         self._updatable_traits = [
@@ -108,6 +113,7 @@ class Q10PropertiesApi(Trait):
             self.network_info,
             self.consumable,
             self.clean_history,
+            self.map_dps,
         ]
         self._subscribe_task: asyncio.Task[None] | None = None
 
--- a/roborock/devices/traits/b01/q10/__init__.py
+++ b/roborock/devices/traits/b01/q10/__init__.py
@@ -32,7 +32,6 @@ __all__ = [
     "DoNotDisturbTrait",
     "DustCollectionTrait",
     "MapContentTrait",
-    "MapDpsTrait",
     "NetworkInfoTrait",
     "SoundVolumeTrait",
     "StatusTrait",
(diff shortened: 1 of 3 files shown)


roborock/devices/traits/b01/q10/map.py
```diff
@@ -1,110 +1,147 @@
...
-    The robot accumulates this server-side and serves the whole trajectory so
-    far in one packet, so it is complete even if we connect mid-session. Only
-    populated while a cleaning session is active."""
+    def __init__(self) -> None:
+        MapDps.__init__(self)
+        UpdatableTrait.__init__(self, command=None, logger=_LOGGER)
 
-    robot_position: Q10Point | None = None
-    """Current robot position (the most recent path point), if known."""
+    @property
```
allenporter, line 57: Do we want to do the parsing once when the dps arrives or each time the rendering isupdated?

Given this is just trying to create a Q10MapOverlays object, maybe it should just build that here some where.

roborock/devices/traits/b01/q10/map.py
```diff
@@ -1,110 +1,147 @@
...
+        """Rooms reported by the device."""
+        return self._map_packet.rooms if self._map_packet else []
+
+    @property
+    def path(self) -> list[Q10Point]:
+        """Full path from the latest trace packet."""
+        return self._trace_packet.points if self._trace_packet else []
+
+    @property
+    def robot_position(self) -> Q10Point | None:
```
allenporter, line 110: Do we have a use case for exposing path, robot position, and heading?

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -96,7 +100,8 @@ def __init__(self, channel: B01Q10Channel) -> None:
         self.button_light = ButtonLightTrait(self.command)
         self.network_info = NetworkInfoTrait()
         self.consumable = ConsumableTrait()
-        self.map = MapContentTrait()
+        self.map_dps = MapDpsTrait()
```
allenporter: How about `self._map_dps` given I think `MapDpsTrait` is effectively a private trait we don't expect callers to access

roborock/devices/traits/b01/q10/map.py
```diff
@@ -1,110 +1,147 @@
...
+    def _map_dps_updated(self) -> None:
+        """Render after the low-level DPS source changes."""
+        if self._map_packet is None:
+            return
+        self._render()
         self._notify_update()
+
+    def _render(self) -> None:
+        """Render the latest map, trace and DPS sources, if a map is available."""
+        if self._map_packet is None:
```
allenporter, line 140: so only a map packet is required, and a trace packet and map overlays are optional?

allenporter (changes requested):

roborock/devices/traits/b01/q10/map.py
tubededentifrice: Path and robot position are already public on main and used by q10-position. They also let callers build a live map UI that updates the robot marker, orientation, and cleaning path independently as trace packets arrive like in the Roborock app (maybe we should also add that image overlay, would be nice). Can also be used for telemetry (path lengths, etc.)

roborock/devices/traits/b01/q10/map.py
tubededentifrice: Moved the parsing into MapDpsTrait.update_from_dps, where it builds and caches a Q10MapOverlays. Subsequent map or trace updates reuse that snapshot instead of parsing the DPS blobs again.

roborock/devices/traits/b01/q10/__init__.py
tubededentifrice: This is now self._map_dps and MapDpsTrait is no longer included in the public exports.

roborock/devices/traits/b01/q10/map.py
tubededentifrice: Yes. The map packet is the only required source. Trace data is optional, and the DPS supplies an empty overlay snapshot until overlay data arrives. Trace and DPS updates received before the map are retained and used when a map becomes available

tests/devices/traits/b01/q10/test_map.py
```diff
@@ -37,35 +63,43 @@ def test_update_from_map_packet_populates_image_and_rooms() -> None:
...
+    assert (trait.robot_position.x, trait.robot_position.y) == (276, -1)
+    assert trait.robot_heading == -34
     assert len(updates) == 1
 
 
 def test_q10_position_is_available_as_top_level_cli_command() -> None:
     assert "q10-position" in cli.commands
 
 
+def test_q10_map_dps_trait_is_private() -> None:
```
allenporter, line 90: Lets not test internal details like this, just exercise behavior through the "api" methods.

allenporter (approved):
Looks great, just a minor change nit -- perhaps you can remove in a follow up PR? Thank you!

PR #888, opened 2026-07-19
test: use mode enums for status codes

Lash-L:
## Summary

- replace raw fan power, water mode, and mop route status codes in container tests with the corresponding dynamic mode enum `.code` values
- keep the assertions aligned with `StatusV2`'s integer representation while making their meaning explicit

## Validation

- `uv run --python 3.11 pytest tests/data/v1/test_v1_containers.py -q` (31 passed)
- `uv run --python 3.11 ruff check tests/data/v1/test_v1_containers.py`
- `uv run pytest tests/data/v1/test_v1_containers.py` on Python 3.14 (31 passed)

A repository-wide run on Windows reaches an existing Python 3.14 asyncio selector failure in `tests/e2e/test_device_manager.py`; the Linux CI matrix provides the authoritative full-suite result.

### DIFF
--- a/tests/data/v1/test_v1_containers.py
+++ b/tests/data/v1/test_v1_containers.py
@@ -5,6 +5,7 @@ import copy
 import pytest
 from syrupy.assertion import SnapshotAssertion
 
+from roborock import CleanRoutes, VacuumModes, WaterModes
 from roborock.data.v1 import (
     MultiMapsList,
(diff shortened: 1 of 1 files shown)


Lash-L:
Updating the enums to make it a bit more understandable!

allenporter (approved):

PR #894, opened 2026-07-22
feat: add support for dock type 31

NOisi-x:
While testing Zeo device functionality, I noticed that my recently purchased vacuum (P20 Ultra Water Hookup Version, `roborock.vacuum.a244`, with auto dust collection and water hookup) has an unsupported dock type, preventing Home Assistant from exposing related entities. This is a small fix.
The dock type naming conventions don't seem to follow a clear pattern, so I've tentatively named it `type_31_dock`. Suggestions for a better name are welcome.


Lash-L:
Hi @NOisi-x! Thanks for the PR, I went ahead and did a full dock type update here that should encompass that! https://github.com/Python-roborock/python-roborock/commit/f431c5dbab2730df73fdaf5b6c42319a0a77a07d

PR #892, opened 2026-07-20
feat!: audit and complete Zeo protocol DP and enum mappings

NOisi-x:
## Summary

Complete a full audit of the `RoborockZeoProtocol` DP enum against the official Roborock Washer app plugin bundle, expanding coverage from 31 to 67 entries and adding all missing enum mappings. Every known Zeo device feature is now represented in the protocol definition.

These findings build upon the excellent reverse engineering work done in [ndwzy/roborock_washer_ha](https://github.com/ndwzy/roborock_washer_ha).

## What was done

A systematic cross-reference was performed between the existing `RoborockZeoProtocol` enum and the React Native plugin bundle [index.ios.bundle](https://github.com/ndwzy/roborock_washer_ha/blob/main/custom_components/roborock_washer/plugins/8f1d582973de4e41a73fc2ea305e17da/index.ios.bundle) extracted from the official Roborock Washer app. Every DP ID, value mapping, and access mode (`ro`/`rw`/`wo`) was verified against the source of truth.

### `RoborockZeoProtocol` — 31 → 67 DP entries

Reorganized into five semantic sections matching the app's internal structure:

| Section | Count | Description |
|---------|-------|-------------|
| Control actions | 3 | START, PAUSE, SHUTDOWN |
| Read-only | 17 | STATE, ERROR, FEATURE_BITS, DRYER_ERROR, etc. |
| startWith params | 12 | MODE, PROGRAM, TEMP, SOAK, DRYING_METHOD, ION_DEODORIZATION, etc. |
| Independent (immediate) | 23 | CHILD_LOCK, DETERGENT_TYPE, SILENT_MODE_*, FLUFF_CLEANED, etc. |
| Meta / RPC / Voice | 12 | ID_QUERY, RPC_REQUEST, VOICE_VOLUME, VOICE_RECORD, etc. |

**Newly identified entries** (previously undefined):

[...]

PR #895, opened 2026-07-22
feat: add MQTT push subscription and real-time state tracking for Zeo devices

NOisi-x:
## Summary

Builds on [#892](https://github.com/Python-roborock/python-roborock/pull/892). Add MQTT push subscription and real-time state tracking for Zeo devices.

### 1. MQTT Push Subscription

`ZeoApi.start()` subscribes to the device's DPS MQTT topic after connection. All device state changes — both physical (button presses, dial rotations) and remote (App commands) — are received as real-time push messages.

```
device.connect() → zeo.start() → MQTT subscribe
                                 → _discover_features() pull full state
```

### 2. DPS Cache with Incremental Updates

Incoming push messages are decoded from RPC protocol 102, parsed as `{dps: {dp_id: value, ...}}`, and merged into `_dps_cache`. Only changed DPs are pushed — the device sends incremental updates, not full state each time.

### 3. Feature Discovery (`_discover_features`)

Mirrors V1's `discover_features()` pattern: after subscribing, queries FEATURE_BITS (DP 237) to wake the device and cache supported capabilities. Exposes `supports(ZeoFeatureBits)` for runtime feature checks.

### 4. Bug Fixes Found During Testing

- `ZeoApi.__init__`: missing `TraitUpdateListener.__init__()` call caused `_update_callbacks` AttributeError
- `device.py connect()`: `self.a01_properties` (doesn't exist) → `self.zeo`

### 5. DP 243 Placeholder

Observed on a114, increments with each push. Not found in the plugin bundle. Added as `UNKNOWN_243` with a comment.

## Testing

Tested on two Zeo devices (a234, a114) using a multi-device MQTT monitor:

| Scenario | Result |
|---|---|
| Physical operations (power on, dial rotate, wash start/pause/cancel) | Real-time push < 1s |
| Door lock/unlock events | Correctly tracked |
| Incremental DP updates | Only changed DPs pushed |
| Two devices concurrently | Independent streams, no crosstalk |
| `_discover_features()` wake | Device pushes full state |


roborock/data/zeo/zeo_code_mappings.py
```diff
@@ -1,142 +1,320 @@
+"""Zeo (washing machine) device enums.
+
+Member-level comments show the manufacturer's official identifiers
```
Lash-L: This line isn't needed. A lot of what I said on the other or stands here imo

roborock/data/zeo/zeo_code_mappings.py
```diff
@@ -1,142 +1,320 @@
...
+never sent to the device.
+"""
+
 from ..code_mappings import RoborockEnum
 
 
+class ZeoFeatureBits(RoborockEnum):
+    """Bit positions in DP 237 (FEATURE_BITS).
+
+    Extracted from the official Roborock Washer app plugin bundle
```
Lash-L: Same here

Lash-L (changes requested):

Lash-L:
Will review more when we merge the other pr as this seems to have some of the same work in it

NOisi-x:
A bit weird. I received some comments via email, but I didn't see anything here. I kept all the notes hoping to help with the review as much as possible, share my findings, and stay compatible with the existing code, leaving room for future changes for syncing and HA integration. If the reviewer thinks some comments aren't needed, I can delete them anytime. Thanks a lot.

roborock/devices/traits/a01/__init__.py
```diff
@@ -156,14 +168,74 @@ async def set_value(self, protocol: RoborockDyadDataProtocol, value: Any) -> dic
...
+
+        Sending an RPC query after subscribing triggers the device to
+        start pushing its full state — equivalent to how V1's
+        ``discover_features()`` uses ``device_features.refresh()`` to
+        initiate the push cycle.
+        """
+        try:
+            result = await self.query_values([RoborockZeoProtocol.FEATURE_BITS])
+            self._feature_bits = result.get(RoborockZeoProtocol.FEATURE_BITS, 0)
+        except Exception:
```
allenporter: Catch more narrow exceptions here, only what we actually expect to see raised here.

If there is a transient RPC here, not clear to me it is OK to fail silently with no features, is it? what do you expect the user to do here with this device?

roborock/devices/traits/a01/__init__.py
```diff
@@ -156,14 +168,74 @@ async def set_value(self, protocol: RoborockDyadDataProtocol, value: Any) -> dic
...
+        requested DPs are already up to date.
+        """
+        if message.protocol != RoborockMessageProtocol.RPC_RESPONSE:
+            return
+        try:
+            decoded = decode_rpc_response(message)
+            self._dps_cache.update(decoded)
+            self._notify_update()
+        except RoborockException:
+            _LOGGER.debug("Failed to decode push message, skipping: %s", message, exc_info=True)
```
allenporter: This seems like it should just be around the `decode_rpc_response` and not the others?

roborock/devices/traits/a01/__init__.py
```diff
@@ -156,14 +168,74 @@ async def set_value(self, protocol: RoborockDyadDataProtocol, value: Any) -> dic
...
+class ZeoApi(Trait, TraitUpdateListener):
     """API for interacting with Zeo devices."""
 
     name = "zeo"
 
     def __init__(self, channel: MqttChannel) -> None:
         """Initialize the Zeo API."""
+        TraitUpdateListener.__init__(self, _LOGGER)
         self._channel = channel
+        self._dps_cache: dict[int, Any] = {}
```
allenporter, line 215: This appears unused in this PR. Can we explain how we expect this to be used here and what the semantics are? when it is ok to use vs when do we need to refresh, etc.

allenporter (changes requested):

roborock/devices/traits/a01/__init__.py
NOisi-x: Changed `except Exception` → `except RoborockException`. 
The Bundle follows the same pattern: query FEATURE_BITS, default to `0` on null response. `featureBits = 0` means all feature-gated operations are disabled — basic start/stop/query still work.

Bundle also skips the FEATURE_BITS query entirely for older device series via `isSupportFeatureBits`. Only two models are excluded: `roborock.wm.a63` (H1) and `roborock.wm.a90` (H1 Lite). Should I add an explicit model check for those, or is the current timeout-then-default approach acceptable?

roborock/devices/traits/a01/__init__.py
NOisi-x: yes yes. fixed

roborock/devices/traits/a01/__init__.py
NOisi-x: See #897 
`ZeoCommandTrait._get_start_params()` checks the cache for `MODE`/`PROGRAM` and only issues a device query on cache miss — avoiding redundant network round-trips when values were already received via push.

roborock/data/zeo/zeo_code_mappings.py
NOisi-x: The comments have finally appeared! I’m not sure what happened before, but what I want to say is that the current annotations are all temporary, added just to make reviewing easier. If you have more suggestions, feel free to bring them up and I can delete them all at once. Thanks!

roborock/devices/traits/a01/__init__.py
allenporter: OK, see comments below.

roborock/devices/traits/a01/__init__.py
```diff
@@ -156,14 +168,80 @@ async def set_value(self, protocol: RoborockDyadDataProtocol, value: Any) -> dic
...
+
+        Only devices that support the FeatureBits DP will respond;
+        older or unsupported devices return nothing.
+        A failed query defaults to 0 — all feature-gated DPs are
+        disabled and the device operates in basic mode.
+        """
+        try:
+            result = await self.query_values([RoborockZeoProtocol.FEATURE_BITS])
+            self._feature_bits = result.get(RoborockZeoProtocol.FEATURE_BITS, 0)
+        except RoborockException:
```
allenporter: silently supporting no features on a transient error (e.g. network unreachable) is a problem. are there cases you expect this to fail that are more narrow?

roborock/devices/device.py
```diff
@@ -202,6 +202,8 @@ async def connect(self) -> None:
                 await self.v1_properties.start()
             elif self.b01_q10_properties is not None:
                 await self.b01_q10_properties.start()
+            if self.zeo is not None:
```
allenporter: the others are elif and this is if. can you make it the same unless there is a reason to be different?

roborock/devices/traits/a01/__init__.py
```diff
@@ -156,14 +168,80 @@ async def set_value(self, protocol: RoborockDyadDataProtocol, value: Any) -> dic
...
+        """Check whether the device supports a given feature bit."""
+        return bool(self._feature_bits & (1 << feature.value))
+
+    def _on_dps_message(self, message: RoborockMessage) -> None:
+        """Handle unsolicited MQTT push (protocol 102 — RPC_RESPONSE).
+
+        Zeo devices broadcast status changes as ``{"dps": {...}}`` JSON
+        payloads.  This callback decodes them and feeds the cache so
+        that ``query_values`` can skip the device round-trip when the
+        requested DPs are already up to date.
```
allenporter: how does query values know when a DPS is up to date? This implies we should be switching to a different API model like traits. (e.g. we have a `refresh` method and stateful properties) or alternatively implement two query apis (one that supports caching and one that does not, when the intent is to get live data). This currently makes the query values behavior undefined (though also it doesn't seem to be reading anyway so the comment is wrong).

allenporter (dismissed):

roborock/devices/device.py
NOisi-x: fixed

roborock/devices/traits/a01/__init__.py
NOisi-x: Fixed. `except RoborockException` → `except RoborockTimeout`. A timeout is the only expected failure mode for this query. Other exceptions (e.g. `RoborockConnectionException`) propagate so the connection layer can retry. For the two models known to lack DP 237, the query is skipped entirely via `_UNSUPPORTED_FEATURE_BITS`.

roborock/devices/traits/a01/__init__.py
NOisi-x: fixed

roborock/data/zeo/zeo_code_mappings.py
Lash-L: My account was temporarily suspended on GH mistakenly. So all of my content was removed until it was resolved, that is probably what you saw!

Lash-L:
Hey @NOisi-x can you check out the merge conflicts?

roborock/devices/traits/a01/__init__.py
NOisi-x: Finally had some spare time to add a few things. I reworked the initial query
behavior after subscribing to MQTT so that it fetches a full device state dump.
Previously it only queried feature bits, and the device would occasionally push
a full state update, which misled me into thinking it was working — turns out
that was actually triggered by the HA integration polling. A silly mistake. I
added `device_feature.py` to build the query command, making it as accurate and
efficient as possible by matching Bundle's `forceLoad()` logic per device
series. Tested on real local devices, so transient errors should now be ruled
out — any response timeout can be attributed entirely to the device being
unreachable and triggering a reconnect.

roborock/data/zeo/zeo_code_mappings.py
NOisi-x: deleted

roborock/data/zeo/zeo_code_mappings.py
NOisi-x: deleted

NOisi-x:
Conflicts solved.
While testing my devices, I noticed that the device occasionally reports
`0` for certain DPs over MQTT — just the device zeroing out values while idle.
These zeros have no real meaning, but they kept spamming "defaulting to 1"
warnings in the logs, which was annoying. As a defensive measure, I added
`null = 0` to every Zeo enum: when `0` appears, it maps to `null` to make it
explicit that the value carries no meaningful state, rather than triggering a
`_missing_` fallback warning.

roborock/devices/traits/a01/__init__.py
```diff
@@ -187,14 +199,71 @@ def on_message(message: RoborockMessage) -> None:
...
+        via incremental MQTT push.
+        """
+        await self._ensure_subscribed()
+        await self._force_load()
+
+    def close(self) -> None:
+        """Unsubscribe from MQTT push and release resources."""
+        if self._dps_unsub is not None:
+            self._dps_unsub()
+            self._dps_unsub = None
```
Lash-L, line 237: I think(?) this should be called inside RoborockDevice.close()

roborock/devices/traits/a01/device_feature.py
```diff
@@ -0,0 +1,294 @@
...
+                RoborockZeoProtocol.SOFTENER_EMPTY,  # 227
+            ]
+        )
+
+    # ── Feature-bits-gated DPs (queried immediately, not deferred) ──
+    if supports_feature_bits(model):
+        # These are always appended when FEATURE_BITS is supported.
+        # The actual bitmask is checked later in loadFeatureDps() for
+        # feature-gated sub-queries, but Bundle appends these upfront.
+        pass  # DP 237 is already in the base list.
```
Lash-L: "forceLoad() always includes DPs 235, 236, and 238 when feature bits are supported, then calls loadFeatureDps() to issue a second query for the DPs enabled by the returned bitmask. This currently does neither, so the initial state can omit smart-hosting and enabled feature state such as silent mode, dry care, smile light, voice assistant, dirt detection, and wash/dry linkage. Could we implement that two-stage load before calling this complete?"


Feedback from an agent - does this seem right?

Lash-L (changes requested):

roborock/devices/traits/a01/device_feature.py
NOisi-x: Right！I've updated the full two-stage load: 
1. **First stage** (`_force_load`): queries the base DP list plus `SmartHosting` (235), `SmartHostingTime` (236) and `SmartHostingWaitedTime` (238) when `FEATURE_BITS` is supported. 
2. **Second stage** (`_load_feature_dps`): reads `FEATURE_BITS` (237) and issues a follow-up query for the DPs gated behind each enabled feature — silent mode, dry care, smile light, voice assistant, dirt detection, wash/dry linkage, etc. UV light (228) is gated by a series whitelist (matching `isSupportUVLight`), independent of the feature bits. 
The second-stage failure is non-fatal (logged as a warning) rather than rejecting the whole load, since the base state is already populated and any missing feature DPs will be backfilled by subsequent MQTT pushes.

roborock/devices/traits/a01/__init__.py
NOisi-x: fixed

Lash-L (approved):
Thanks @NOisi-x!

Please keep an eye out for any comments Allen may leave after merging, but i'm going to go ahead and merge!

roborock/devices/traits/a01/__init__.py
allenporter: OK i think in the future we may only want to store device features in this and not have other dps values here, but instead expose them via the traits they belong to.

roborock/devices/traits/a01/__init__.py
```diff
@@ -187,14 +203,93 @@ def on_message(message: RoborockMessage) -> None:
...
+        feature_dps: list[RoborockZeoProtocol] = []
+        if self._feature_bits:
+            feature_dps.extend(build_feature_dp_list(self._feature_bits))
+        if supports_uv_light(self._model):
+            feature_dps.append(RoborockZeoProtocol.UV_LIGHT)
+        if not feature_dps:
+            return
+        try:
+            await self.query_values(feature_dps)
+        except RoborockException as exc:
```
allenporter, line 275: What I was saying before with my comment is this may still fail, it wasn't clear why its Ok to proceed here without failing startup.

I'm am a little confused by a few things happening: 
- _force_load: we make a list of initial values to query from `build_force_load_dp_list` . we then set _feature_bits
- _load_feature_dps is called  building the a feature list with `build_feature_dp_list`  plus some additional protocol values, then again run `query_values` and ignore the results. It may additionally fail optionally and we ignore it and say feature dps was not loaded (but feature bits was already set)
- then we also have an async path for setting data in a cache (which we're not yet using but may in a future PR, for unsolicited messages

PR #897, opened 2026-07-22
feat: typed commands, settings & status for Zeo devices

### LINKED ISSUE #833: [Feature/Bug] Zeo Washing Machine (M1S Ultra): Deep Sleep Wakeup Failure & Missing DP Mappings

NOisi-x:
> **Builds on [#895](https://github.com/Python-roborock/python-roborock/pull/895)**  
> **Closes [#833](https://github.com/Python-roborock/python-roborock/issues/833)**

## What this PR adds

This PR delivers the core **Zeo API** — the ability to actually start, pause,
resume, stop and schedule a wash programme on Zeo (washing machine / dryer)
devices, with fully typed parameters. It is the successor to the earlier
`ZeoCommandTrait` + `ZeoFeatureTrait` draft, restructured into a single
`ZeoApi` entry point with three lazily-built sub-traits, and hardened by
on-device verification.

### The API surface

```python
api: ZeoApi = device.zeo  # built by a01 create() from the product category

# Read-only state, typed (state, error, timers, tank levels, wash log, ...)
api.status.state                 # ZeoState.standby / washing / ...
api.status.washing_left          # minutes remaining
api.status.washing_log           # ZeoWashLog with typed ZeoProgram / ZeoMode

# Writable state + typed setters (mode, programme, temperature, spin, ...)
api.settings.mode                # ZeoMode.wash
api.settings.temperature         # ZeoTemperature.medium
await api.settings.set_temperature(ZeoTemperature.high)

# Commands
await api.command.start_with(ZeoStartParams(
    mode=ZeoMode.wash_and_dry,
    program=ZeoProgram.silk,
    temperature=ZeoTemperature.low,
    rinse=ZeoRinse.high,
    spin=ZeoSpin.mid,
    drying_mode=ZeoDryingMode.quick,
    ion_deodorization=True,       # caller-supplied, feature-gated
))
await api.command.start_with_custom_mode()   # reuse device's saved DP 222 programme
await api.command.preset_with(params, minutes=120)  # delayed start (>30 min enters countdown)
await api.command.pause() / resume() / stop() / shutdown()
```

### Key pieces

| Module | Role |
| --- | --- |
| `traits/a01/__init__.py` | `ZeoApi` — MQTT push subscription, two-stage force-load, `query_values`/`set_value`/`get_custom_mode`, lazily exposes `command`/`settings`/`status` |
| `traits/a01/command.py` | `ZeoCommandTrait` — `start_with`, `start_with_custom_mode`, `preset_with`, `pause`, `resume`, `stop`, `shutdown` |
| `traits/a01/settings.py` | `ZeoSettingTrait` — typed writable state + typed setters, `build_param_dps` (params → DPs) |
| `traits/a01/status.py` | `ZeoStatusTrait` — read-only state (state/error/timers/tanks/wash log) |
| `traits/a01/device_feature.py` | Series whitelists (`is_dryer`, `supports_uv_light`, ...) + `ZeoFeatures` parsed from DP 237 FEATURE_BITS |

### Design decisions

1. **One trait family, not per-device classes.** Washer vs dryer differ only in
   which start parameters are sent (10 DPs vs 7). Everything else is shared, so
   a single set of traits with an `is_dryer` flag — resolved once at
   construction from the model ID — mirrors how V1's `CommandTrait` handles
   vastly different dock types through one class.

2. **Two-stage state load, matching the app.** `ZeoApi.start()` queries the
   base DP list first, then issues a follow-up query for the DPs gated behind
   each enabled FEATURE_BITS (DP 237) bit — the same `forceLoad()` →
   `loadFeatureDps()` flow in the official Bundle. The second stage is
   non-fatal on failure (logged, backfilled by subsequent MQTT pushes).

3. **Caller-supplied start parameters.** `ZeoStartParams` is the *single
   source* of what gets started. Feature-gated values like
   `ion_deodorization` (DP 258) and `wash_dry_linked` (DP 255) are passed by
   the caller (`None` omits the DP), instead of being silently read back from
   the device cache. `start_with(params)` is a pure input → command function.

4. **Integer boolean protocol.** Boolean DPs (auto-dosing 211/212, feature
   gates) are sent as integers `1`/`0` — on-device testing showed string
   booleans (`"False"`) are silently ignored by the device.

5. **Typed wash log.** `ZeoWashRecord.prog_type` → `ZeoProgram` and
   `category` → `ZeoMode`, so consumers get `record.prog_type.name` directly.
   Unknown values degrade to `None` instead of breaking the whole log decode.

## Verified on real hardware

All command paths below were tested against a physical device (MQTT trace +
state transitions confirmed):

- **`start_with_custom_mode`** — DP 222 custom programme is decoded into
  `ZeoStartParams` and the device boots with exactly those parameters
  (`state: standby → washing`, `program: boiling_wash → silk`), auto-dosing
  integers accepted.
- **`preset_with`** — a valid parameter combination with `minutes > 30`
  reliably enters the delay-start countdown state (`under_delay_start`,
  `countdown` set). Constraint documented in the docstring.
- **`pause` / `resume` / `shutdown`** — single-DP commands, QoS 1.

## Follow-ups (next PR)

- Programme-config table-Solve the problem of making startup parameters valid


allenporter (commented):
Great work articulating the differences between the washer and dryer. Given the differences (e.g. params, modes, etc). This seems like a perfect use for separate traits. It seems like a "washer trait" and "dryer trait" now make sense to introduce.

Can you review the existing trait pattern for prior art?

NOisi-x:
@allenporter The difference between washer and dryer in `ZeoCommandTrait` is just the start parameter list — 10 DPs vs 7 DPs. Everything else (`start_program`, `pause`, `resume`, `shutdown`, cache checking, feature gating) is identical since they are all Zeo devices. Splitting into two classes would duplicate ~95% of the code.

V1's `CommandTrait` serves the same role for vacuums with vastly different dock types (pure collect vs collect+wash+dry+plumbing) — all through a single trait. The `is_dryer` flag here is equivalent to V1's `dock_features` check: resolved once at construction, not at runtime.

Given the minimal difference, is the current single-trait approach acceptable, or would you still prefer separate classes?

allenporter:
II'm thinking of this more like: What do solid washer and dryer APIs look like? My assumption is we'll want APIs that return the current settings to the caller in these new modes -- not just start params but also querying the values and/or holding on to the current state. That is, if we're moving to traits there are more benefit is in terms of using these new types we have defined.

The thing to prioritize is what the API looks like. We can avoid code duplicating by sharing code where it makes sense. (If its really all duplicated, then there are more solutions than just having all the code in the same file. (e.g. sharing code between separate files is possible)

NOisi-x:
@allenporter I agree with the direction — separate washer/dryer traits with type-safe APIs returning our enum types is the right end state. The individual DP getters/setters (set_program(ZeoTemperature), get_mode() → ZeoMode, etc.) are where the washer/dryer API divergence really shows up, and that's exactly where splitting makes sense.

However, those methods are out of scope for this PR. Right now the trait only does start_program (bundled command), pause, resume, shutdown — all of which are identical for washers and dryers. Adding typed per-DP methods would roughly double the size of this PR, and I have limited time for this project in the near future. I'd prefer to land this as-is (single trait, scope limited to composite commands + feature discovery) and handle the trait split together with the typed getter/setter work in a follow-up PR.

Would that be acceptable?

NOisi-x:
@allenporter Now that I have some time to continue diving into this project, I've refactored this PR. Please check out my updated PR description and review it. Thanks a lot.

roborock/devices/traits/a01/__init__.py
```diff
@@ -111,6 +132,17 @@
...
 }
 
+
+def _try_json(val: Any) -> Any:
+    """Return *val* parsed as JSON when it is a JSON string, else *val*."""
+    if isinstance(val, str):
+        try:
+            return json.loads(val)
+        except ValueError:
+            pass
```
allenporter: What are you expecting to happen, end to end, here?

roborock/devices/traits/a01/__init__.py
```diff
@@ -206,16 +294,81 @@ def on_message(message: RoborockMessage) -> None:
...
+    def features(self) -> ZeoFeatures | None:
+        """The device capabilities parsed from FEATURE_BITS (DP 237).
+
+        ``None`` until the first force-load completes.
+        """
+        return self._features
+
+    @property
+    def command(self) -> ZeoCommandTrait:
+        """Lazily-built trait for wash-programme commands."""
```
allenporter, line 318: What is your intent behind the laziy-built traits? (One side effect: Updates aren't applied if it hasn't been added yet, which is surprising to me)

roborock/devices/traits/a01/__init__.py
```diff
@@ -298,13 +449,36 @@ async def query_values(self, protocols: list[RoborockZeoProtocol]) -> dict[Robor
...
+        total_time = self._dps_cache.get(int(RoborockZeoProtocol.TOTAL_TIME))
+        try:
+            raw_int = int(raw)
+        except (TypeError, ValueError):
+            return None
+        if is_dryer(self._model):
+            return ZeoDryerCustomMode.from_raw(raw_int, total_time)
+        return ZeoCustomMode.from_raw(raw_int, total_time)
+
+    async def update_sound_package_info(self) -> Any:
```
allenporter: We shouldn't use Any but a more specific type here.

roborock/devices/traits/a01/__init__.py
```diff
@@ -298,13 +449,36 @@ async def query_values(self, protocols: list[RoborockZeoProtocol]) -> dict[Robor
...
 
     async def set_value(self, protocol: RoborockZeoProtocol, value: Any) -> dict[RoborockZeoProtocol, Any]:
         """Set a value for a specific protocol on the device."""
         params = {protocol: value}
         return await send_decoded_command(self._channel, params, value_encoder=lambda x: x)
 
+    async def get_custom_mode(self) -> ZeoCustomMode | ZeoDryerCustomMode | None:
+        """Query and decode the current custom programme (DP 222)."""
+        await self.query_values([RoborockZeoProtocol.CUSTOM_PARAM_GET, RoborockZeoProtocol.TOTAL_TIME])
+        raw = self._dps_cache.get(int(RoborockZeoProtocol.CUSTOM_PARAM_GET))
```
allenporter: I don't really understand this. query values has a return value.

Is the problem you're trying to workaround here with with dps cache is that the return values don't work reliably or something? It may be that the existing approach for this device is wrong, and we need to move to refresh + trait listeners.

roborock/devices/traits/a01/command.py
```diff
@@ -0,0 +1,219 @@
...
+        dps[RoborockZeoProtocol.COUNTDOWN] = minutes
+        await send_decoded_command(
+            self._channel,
+            dps,
+            qos=MqttQos.AT_LEAST_ONCE,
+            value_encoder=lambda x: x,
+        )
+        return dps
+
+    async def pause(self) -> dict[RoborockZeoProtocol, Any]:
```
allenporter: what is your intent intent behind including the dps return value here? I don't think these commands should have any return value at all.

allenporter (changes requested):
Hi, this PR is over 1k lines and is pretty huge, so not going to be something i can review quickly. 

We're adding features, fixing small issues/enums, and also there are also multiple fundamental architecture shifts we're making here because the existing support is pretty naive.  We're going to need to work on these more incrementally. Can we tease this apart into a multiple smaller logical PRs please?

roborock/devices/traits/a01/__init__.py
```diff
@@ -75,6 +92,10 @@
 __init__ = [
     "DyadApi",
     "ZeoApi",
+    "ZeoCommandTrait",
+    "ZeoFeatures",
+    "ZeoSettingTrait",
+    "ZeoStatusTrait",
 ]
```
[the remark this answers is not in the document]
allenporter: Sent https://github.com/Python-roborock/python-roborock/pull/920

roborock/devices/traits/a01/__init__.py
NOisi-x: These Zeo DPs (voice volume, voice switch, sound-package info, voice records) arrive over MQTT as **JSON strings**.

`_try_json` parses those JSON strings into Python dicts so other code can use them directly. Its behaviour is:

- string is valid JSON → parse into dict / list / native value
- not a string (int, already-parsed dict, `None`) → passed through untouched
- string is **malformed** (garbage from the device) → returned as-is (parsing it would just fail), with a debug log line

The reason Why not raise on malformed JSON is that `convert_zeo_value` wraps every converter in a "fail → `None`" guard:

```python
try:
    return converter(value)
except (ValueError, TypeError):
    return None
```

If `_try_json` raised on bad data, that guard would **silently turn it into `None`** — and the caller couldn't tell "device didn't return this DP" apart from "device returned it but it's corrupt". Returning the raw string keeps the data around, so anyone who cares can inspect it; the debug log makes the failure findable.

Apart from adding that debug log to `_try_json`, I also corrected the read/write semantics of the voice-related DPs:

| DP | readable? | notes |
|----|-----------|-------|
| `VOICE_VOLUME`(10009) | ✅ | device answers queries (`voiceVolume` getter, gated by `FeatureBit.VoiceAssistant`) — **read-write**, converter kept |
| `VOICE_SWITCH`(10301) | ✅ | same (`isVoiceSwitchOn` getter) — **read-write**, converter kept |
| `SET_SOUND_PACKAGE`(10003) | ❌ | fire-and-forget command, device sends no response — **write-only**, converter removed |
| `VOICE_RECORD_DELETE`(10304) | ❌ | same — **write-only**, converter removed |

Previously all four were treated as readable; the last two never hit the query path and were dead entries. Now only the ones that are actually reachable stay in the table.

roborock/devices/traits/a01/__init__.py
NOisi-x: Agreed — changed the return type from `Any` to `dict[str, Any] | None`.

roborock/devices/traits/a01/__init__.py
NOisi-x: I just read your new comments in https://github.com/Python-roborock/python-roborock/pull/895 and find I am totally wrong now.
Really sorry about that.
I want to be fully transparent here: the lazy traits + _dps_cache design was
my own choice, and I've come to see it as the root of much of this PR's
complexity — it's what forces the backfill, the pre-warm query, the null
checks, and the side-effect dependency in query_values() that you flagged in
multiple places. My lack of experience with this codebase is the honest reason
it ended up that way; I reached for a caching layer before fully working
through the data flow. I realize now that an always-present (eager) trait model
— where pushes flow straight into the traits and no cache exists — would be
simpler and match the bundle's own onDpsChange behavior directly.

Rather than restructure it inside this PR and force another review cycle, I'd
like to do this as a follow-up PR, and I'd value your confirmation on the
direction before I start. My plan:

make the traits always-present, constructed in start() after _force_load;
delete _dps_cache entirely — pushes route straight into the traits via
update_from_dps();
drop the pre-warm side-effect: _load_feature_dps keeps querying the
feature-gated DPs (the two-stage load is the bundle's design), but its
return value then gets consumed normally instead of discarded.
Scope is roughly a01/__init__.py plus the trait tests — no HA-facing API
change. Does that direction work, or would you restructure it differently?
I'm happy to adjust to your preference before opening the PR.

roborock/devices/traits/a01/__init__.py
NOisi-x: You're right — there was no reason to bypass the return value. To be honest, my
original rationale for reading `_dps_cache` was a mistake. I reasoned that MQTT
messages are parsed continuously, and since the return value is itself an MQTT
message that gets fed back into the cache, I could just read the cache back.

So the old code discarded the reliable return value and re-read the raw string
from `_dps_cache`, then converted it again by hand — pointless indirection.

Fixed: `get_custom_mode()` now uses `query_values()`'s return value directly.

roborock/devices/traits/a01/command.py
NOisi-x: Fair point that the commands are fire-and-forget — `send_decoded_command()`
only publishes the payload and never waits for a device response. But the
return value on the *start* methods is intentional, and it isn't just echoing
the caller's input:

`start_with()` / `preset_with()` accept **any** combination of start
parameters the caller provides, and not every combination is valid for the
device. There is a programme-config template (not yet implemented) that
defines the valid parameter sets; the returned DPS frame is what I plan to
validate against it. The frame also includes DPs these methods add on their
own — auto-dosing (DP 211/212) pulled from the settings trait and feature-
gated DPs (255/258) — which the caller can't derive from the params alone, so
returning it gives them the full picture of what was actually sent.

So I've kept the return value on the parameter-combination methods
(`start_with`, `start_with_custom_mode`, `preset_with`), since those are the
ones that accept arbitrary combos and are candidates for validation. The
simple one-shot commands (`pause`, `resume`, `stop`, `shutdown`) send a fixed
single DP and now return `None`, matching the rest of the codebase's command
traits.

NOisi-x:
> Hi, this PR is over 1k lines and is pretty huge, so not going to be something i can review quickly.
> 
> We're adding features, fixing small issues/enums, and also there are also multiple fundamental architecture shifts we're making here because the existing support is pretty naive. We're going to need to work on these more incrementally. Can we tease this apart into a multiple smaller logical PRs please?

@allenporter Please don't be put off by the line count — the bulk of the diff is three new
trait files (`settings.py`, `command.py`, `status.py`), and most of their
methods are trivial semantic wrappers over the raw MQTT/DPS protocol. There's almost no branching logic in them.

I wrote them for completeness of the semantic API surface. If you feel some of
them are unnecessary, they can be trimmed — I'm happy to drop the ones that
carry no real logic and keep only the wrappers that do meaningful work.

On splitting: I'd still prefer to land this as one PR, because the existing
Zeo support is a raw passthrough with almost no typed surface, and this PR
collectively gives Zeo a nearly complete typed API (commands, setters, status,
feature detection). Splitting would mean landing half-working slices. But I'm open to it — if you point at a seam
you'd like to split along, I'll follow your lead. There's no deadline here, so
please review at whatever pace is comfortable.

Finally — thank you for the careful review. Your comments have been really
valuable to me; each one made me re-examine my work and the resulting code
is much better for it. I really appreciate the time you're putting into this.

roborock/devices/traits/a01/__init__.py
NOisi-x: End to end, `_try_json` exists because these voice DPs are sent by the device
as **JSON strings**, not scalar values, and the converter registered in
`ZEO_PROTOCOL_ENTRIES` is what turns the raw string into the parsed object
callers actually want. In this PR I touched four of them:

- `VOICE_VOLUME` (10009) — read-write; the bundle has a `voiceVolume` getter
  and includes it in `loadFeatureDps` (gated by `VoiceAssistant`), so I **added**
  its `_try_json` converter so `query_values()` returns the parsed JSON object.
- `VOICE_SWITCH` (10301) — read-write; same story (`isVoiceSwitchOn` getter,
  in the feature query list) — converter **added**.
- `SET_SOUND_PACKAGE` (10003) — write-only; no getter, not in any query list —
  converter **removed** (dead code).
- `VOICE_RECORD_DELETE` (10304) — write-only; same — converter **removed**.

So the end-to-end flow for the two read-write ones is:

```
query_values([VOICE_VOLUME])
  → device returns e.g. '{"snd_volume": 15}'  (a string)
  → convert_zeo_value → _try_json → json.loads → {"snd_volume": 15}
  → caller gets the parsed dict
```

The `except ValueError: return val` fallback is deliberate: `_try_json` is
shared by several DPs, and it must not crash on a value that isn't valid JSON
(a plain number, a value already parsed, or a non-JSON string) — it returns
the value as-is, and `convert_zeo_value` catches any remaining `TypeError` and
yields `None`. So the contract is: **parse when possible, leave it alone
otherwise.**

Lash-L:
@NOisi-x can you fix the merge conflicts and then mark as ready for review?

roborock/devices/traits/a01/__init__.py
NOisi-x: After rebasing onto `main`, #941 landed the `A01Api` base class with its own
`values` state tracking — which is the shape I was describing — so I adopted it
directly:

- the traits (`command`, `settings`, `status`) are now built eagerly and are
  always present;
- `_dps_cache` and the first-access backfill are gone — a single
  `_merge_datapoints()` hook projects the raw datapoints into the traits and
  merges them into `values`, covering unsolicited pushes, the initial cloud
  status snapshot, and query responses;
- `_load_feature_dps` flows through that same path now, so its return value is
  no longer discarded and there is no hidden side-effect dependency.

Tests cover the push → trait projection and the initial-status seeding. Thanks
again for pointing this out.

NOisi-x:
> @NOisi-x can you fix the merge conflicts and then mark as ready for review?

@Lash-L Done. Based on #941 I also made some small improvements and removed dead code. Thanks!

PR #898, opened 2026-07-24
Recover A01 devices after sleep and support push updates

piitaya:
## Proposed change

A01 devices (Dyad wet/dry vacuums, Zeo washers) could stop responding after going to sleep and never recover on their own. They now reconnect automatically, and Dyad vacuums can report state changes as they happen instead of only when polled.

Both changes are tested on a Roborock F25 XT (`roborock.wetdryvac.a238`). The listener change will need a HA core change to take effect (I'll open a PR for that)


Lash-L:
Going to leave this one for @allenporter before I merge, looks good to me, but RPC channels are definitely more his domain!

Happy to see someone contributing logic for A01 devices though 👀

allenporter (dismissed):

allenporter:
Please address CI Lint errors

piitaya:
Done ! Should be good now.

Lash-L (approved):

Lash-L:
TY @piitaya!

PR #900, opened 2026-07-25
feat: align Q10 map rendering with V1

tubededentifrice:
This draft intentionally remains the aggregate implementation and full real-device/Home Assistant before-and-after reference. Please review and merge the smaller PRs:

- #901 — independent `dpMultiMap` refresh transport
- #902 — V1 palette and deterministic adjacent-room colors (rendering 1/3)
- #903 — shared V1 overlay renderer (rendering 2/3, stacked on #902)
- #904 — docked-state composition (rendering 3/3, stacked on #903)

The stacked PRs include an explicit incremental compare link in their descriptions. Once each parent merges, its child can be rebased onto `main` without changing the review boundary.

## What changed

- fetch Q10 maps through the asynchronous `dpMultiMap` list/get flow
- render the Q10 raster with the V1 palette and transparent outside area
- assign room colors through one deterministic V1/Q10 adjacency policy, changing a room's canonical color only when a neighboring room would otherwise look identical
- apply erase zones and restriction geometry with their 5 mm header calibration, independently from 2.5 mm trace calibration
- reuse the V1 image generator for path, robot, dock, no-go/no-mop areas, and virtual walls to prevent visual drift
- place the saved dock from its absolute header pixel coordinates and show the robot directly in front of it while charging, without inventing an idle path
- place restrictions from header metadata even while the robot is idle

## Why

Q10 maps did not consistently refresh through `dpRequestDps`, erase polygons were projected with the trace scale, and Q10 maintained custom marker/path/wall drawing that differed from V1. The V1 palette also intentionally repeats some colors (for example, room slots 2 and 12), so adjacent rooms using those slots could visually merge. The saved dock coordinates were also receiving the world origin a second time, which moved the dock off its real position, while an idle trace contains no robot coordinates. The Q10 `charger_phi` is already the V1-compatible outward heading and must be preserved directly: R1 reports about `3°` and points right, while RDC reports about `91°` and points up. Together these issues made through-window phantom areas remain visible, hid or misplaced the charging robot/dock, merged some adjacent rooms, and allowed the Q10 and V1 renderers to drift.

## Before / after

Captured from the same R1 map in the configured local Home Assistant test instance. The before image is the original preview saved before any change in this PR. The after image is the final R1 Home Assistant render with corrected erase zones, restrictions, transparency, adjacent-room colors, and charging robot/dock placement.

| Before: original Q10 renderer | After: final V1-aligned renderer |
| --- | --- |
| Original Q10 rendering before this PR | Final R1 Q10 rendering in Home Assistant |

## Validation

- `pytest -q` — 901 passed, 86 snapshots passed
- focused Q10 map/status/protocol and shared V1 parser tests — 113 passed
- `ruff check .`
- `ruff format --check .`
- repository-pinned mypy hook on all changed non-CLI source modules
- `codespell` on all changed files
- `git diff --check`
- current R1 grid/header render (`charger_phi=3°`) places the dock left of the robot, pointing right, with no idle path
- RDC's current `charger_phi=91°` remains upward-facing; both cardinal directions are covered by regression tests
- GitHub CI on the preceding visual head — all required checks passed; the exact final head is covered by the local validation above, while its Actions run failed before GitHub created any jobs or logs

## Impact

The map/protocol changes remain scoped to Q10. The shared room-color policy is also applied to V1 base-map parsing: it preserves every canonical room color unless an edge-adjacent room already uses the same RGB value, and resets from the canonical palette on every parse to prevent color drift.


tubededentifrice:
While cleaning
image

PR #901, opened 2026-07-25
fix: request Q10 maps through dpMultiMap

tubededentifrice:
## Summary

This PR separates the Q10 saved-map list from the Q10 map content.

`MapsTrait` requests and stores the saved-map list. `MapContentTrait` requests content for a stored map ID. `Q10PropertiesApi` routes both map-list data and other DPS data through its normal update list.

## Why

The map list and map content can change at different times. A general status refresh does not reliably publish the map. The Q10 map request has this flow:

1. Send `dpMultiMap` with `op: list`.
2. Receive the saved-map list later in the DPS stream.
3. Send `dpMultiMap` with `op: get` and a string map ID.
4. Receive the content later in a protocol-301 `MAP_RESPONSE`.

The map list is an initial dependency only when no map ID is available. After that, a caller can refresh map content without a new list request.

## Changes

- Add `MapsTrait` for the map list, its typed converter, and the list request.
- Keep content requests, map packets, trace packets, and rendered content in `MapContentTrait`.
- Require `CommandTrait` in both traits that send commands.
- Add `MapsTrait` to `_updatable_traits`.
- Ignore `op: get` acknowledgements in `MapsTrait` so they cannot replace a valid list.
- Keep general status, map-list, and map-content refresh schedules separate.
- Make the CLI request the list only when it has no stored map ID.
- Use the standard `TraitUpdateListener` interface for CLI waiting.
- Use the standard DPS converter pattern for the typed string map ID.

## Validation

- `pytest -q`: 918 tests passed and 86 snapshots passed.
- All pre-commit checks passed, including Ruff and mypy.
- The branch includes the latest `main` commit.

## Relationship to #900

This PR is independent and targets `main`. The rendering work is split into #902, #903, and #904. PR #900 remains the full real-device and Home Assistant reference.


roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -132,22 +156,55 @@ async def close(self) -> None:
...
         """Refresh all traits."""
-        # Sending the REQUEST_DPS will cause the device to send all DPS values
-        # to the device. Updates will be received by the subscribe loop below.
+        # Status and map retrieval use separate Q10 requests. A bare REQUEST_DPS
+        # reliably refreshes status but does not reliably make every firmware
+        # publish its map.
         await self.command.send(B01_Q10_DP.REQUEST_DPS, params={})
+        await self.request_map()
+
+    async def request_map(self) -> None:
```
allenporter: This should be on the map trait?

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -132,22 +156,55 @@ async def close(self) -> None:
...
+        """
+        async with self._map_request_lock:
+            now = asyncio.get_running_loop().time()
+            if (
+                self._map_list_request_token is not None
+                and self._map_list_requested_at is not None
+                and now - self._map_list_requested_at < _MAP_LIST_REQUEST_TIMEOUT
+            ):
+                return
+            token = object()
```
allenporter: What's going on here? We're trying to implement some kind of rate limiting?

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -132,22 +156,55 @@ async def close(self) -> None:
 
     async def refresh(self) -> None:
         """Refresh all traits."""
-        # Sending the REQUEST_DPS will cause the device to send all DPS values
-        # to the device. Updates will be received by the subscribe loop below.
+        # Status and map retrieval use separate Q10 requests. A bare REQUEST_DPS
+        # reliably refreshes status but does not reliably make every firmware
+        # publish its map.
         await self.command.send(B01_Q10_DP.REQUEST_DPS, params={})
+        await self.request_map()
```
allenporter: (1) Can you move all the map related request code into the map trait?
(2) I think we should refresh the map on its own and not do it here, tied to the overall map refresh interval. We may want to request the map at a different interval and also change it when cleaning, etc.

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -38,6 +40,25 @@
...
+    """Return the first usable map ID from a ``dpMultiMap`` list response."""
+    if not isinstance(response, dict) or response.get("op") != "list":
+        return None
+    data = response.get("data")
+    if not isinstance(data, list):
+        return None
+    for map_info in data:
+        if not isinstance(map_info, dict):
+            continue
+        map_id = map_info.get("id")
```
allenporter: What is the format of "map_id"?

Can we use the standard pattern in in the traits of using a converter etc?

allenporter (changes requested):

allenporter (changes requested):
Please see requested changes to where the code lives

roborock/devices/traits/b01/q10/__init__.py
tubededentifrice: Preventing overlapping requests...

allenporter: The way we typically do this is with asyncio.Lock.

allenporter (changes requested):

roborock/devices/traits/b01/q10/__init__.py
tubededentifrice: device reports map_id as a string, which is sent back unchanged

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -159,6 +158,7 @@ def _handle_message(self, message: Q10Message) -> None:
             # only updates the fields that it is responsible for.
             for trait in self._updatable_traits:
                 trait.update_from_dps(message.dps)
+            await self.map.update_from_dps(message.dps)
```
allenporter: does it make sense for map to be at the end of the list of _updatable_traits?

roborock/devices/traits/b01/q10/map.py
```diff
@@ -70,27 +74,73 @@ def update_from_dps(self, decoded_dps: dict[B01_Q10_DP, Any]) -> None:
...
         self._map_packet: Q10MapPacket | None = None
         self._trace_packet: Q10TracePacket | None = None
         self._image_content: bytes | None = None
+        self._map_packet_callbacks: CallbackList[None] = CallbackList(_LOGGER)
+        self._trace_packet_callbacks: CallbackList[None] = CallbackList(_LOGGER)
         self._map_dps.add_update_listener(self._map_dps_updated)
 
+    async def refresh(self) -> None:
+        """Request the current saved map independently of general status."""
+        if self._command is None:
```
allenporter: This seems like a case that would never happen give it is always passed into the constructor. Update the typing to be `CommandTrait`?

roborock/devices/traits/b01/q10/map.py
```diff
@@ -70,27 +74,73 @@ def update_from_dps(self, decoded_dps: dict[B01_Q10_DP, Any]) -> None:
...
 
-class MapContentTrait(TraitUpdateListener):
+@dataclass
+class MapListDps(RoborockBase):
+    """Typed ``dpMultiMap`` state delivered through the Q10 DPS stream."""
+
+    multi_map: dpMultiMap | None = field(default=None, metadata={"dps": B01_Q10_DP.MULTI_MAP})
+
+
+class MapContentTrait(MapListDps, TraitUpdateListener):
```
allenporter: My previous comment to move things here may have been based on a misunderstanding of what is happening. Perhaps you can explain the overview of whats happening here. I see multiple requests happening... One is to refresh the list of maps and one is to get the map content?

In v1 we use separate traits for these separate concepts:
- maps: gets the list of maps
- map_content: gets the image byte content of for the current map
- home: combines all the data from rooms/maps/etc to hold state e.g.in case different map flags are loaded etc

It seems like we're moving to do something more like that here were we need muliple things, but i'm not sure so I don't want to give you bad advice.

allenporter (changes requested):

roborock/devices/traits/b01/q10/map.py
tubededentifrice: The current code combines two separate operations:
dpMultiMap with op: list requests the map list. The list arrives later in the DPS stream.
The code reads the first string map ID and sends dpMultiMap with op: get.
The get command does not contain the map data. The map data arrives later as a protocol-301 MAP_RESPONSE.
I combined these operations in MapContentTrait because the first response is asynchronous. However, this mixes the map-list state with the map-content state.
I propose this split:
MapsTrait owns the map list, the typed converter, and the list request.
MapContentTrait owns the get request, map packets, and rendered content.
Q10PropertiesApi connects the asynchronous map-list response to the map-content request. This has the same role as the V1 home-level coordination.
Does this match the design that you expect?

roborock/devices/traits/b01/q10/__init__.py
tubededentifrice: Its DPS update is synchronous. Q10PropertiesApi can then request the map content after the list update.

roborock/devices/traits/b01/q10/map.py
allenporter: Having split responsibilities like that seems good to me.

However on implementation details: I don't think we necessarily need to coordinate a map list with async updating map content. My take is that map lists may or may not change over time independent of map content. There may be a dependency, but only initially? I do understand that loading a new map can invalidate the map image which can be handled when accessing the map content.

Thank you for the thoughtful consideration on approach.

tubededentifrice: Agreed. The map list is only a dependency when no map ID is available or when the selected map changes. Map content can otherwise refresh independently with the stored map ID.

So let's split the responsibilities:
- MapsTrait owns the map list
- MapContentTrait gets content for the stored map ID
- Q10PropertiesApi does not request content after each list update
- I will also require CommandTrait and add MapsTrait to _updatable_traits

roborock/data/b01_q10/b01_q10_containers.py
```diff
@@ -83,11 +83,32 @@ def start_datetime(self) -> datetime.datetime | None:
...
+
     op: str
     result: int
-    data: list
+    data: list[Q10MapInfo] = field(default_factory=list)
+
+    @property
+    def current_map_id(self) -> str | None:
+        """Return the first saved-map identifier, if one was reported."""
+        first = next((map_info for map_info in self.data if isinstance(map_info, Q10MapInfo) and map_info.id), None)
```
allenporter: Please remove the isinstance checks.  (In what cases would `data` not be `Q10MapInfo`?). The primary reason for python type checking is to avoid stuff like this -- I assume AI just added it since it tends to do that rather than you just adding this check, right? If there is a reason, then lets discuss the root cause.

roborock/devices/traits/b01/q10/map.py
```diff
@@ -128,12 +156,22 @@ def update_from_map_packet(self, packet: Q10MapPacket) -> None:
...
+        self._map_packet_callbacks(None)
 
     def update_from_trace_packet(self, packet: Q10TracePacket) -> None:
         """Store a trace-protocol update and render the latest sources."""
         self._trace_packet = packet
         self._render()
         self._notify_update()
+        self._trace_packet_callbacks(None)
+
+    def _add_map_packet_listener(self, callback: Callable[[], None]) -> Callable[[], None]:
```
allenporter: This trait is already implementing the `TraitUpdateListener`. I don't think we should add more hooks on internal details for the CLI to use that are not using that interface. 

What is the limitation you're trying to workaround here that requires this back door?

allenporter (changes requested):

roborock/data/b01_q10/b01_q10_containers.py
tubededentifrice: You’re right. Added defensively because the data comes from the device, but the converter already guarantees Q10MapInfo. Removed.

roborock/devices/traits/b01/q10/map.py
tubededentifrice: The idea was trying to stop an overlay or status update from waking the CLI while it waited for a new map packet. But yes, wrong abstraction. Removed; the CLI now uses add_update_listener.

allenporter (approved):

allenporter:
Thank you, great improvement!

roborock/devices/traits/b01/q10/map.py
allenporter: If you want to improve this, then maybe don't send the update signal if the data we're looking for isn't set yet. Or don't fire the update event if the trait is not "ready"?

PR #902, opened 2026-07-25
feat: align Q10 map colors with V1

tubededentifrice:
## Summary

First rendering slice extracted from #900. The Q10 base raster now follows the V1 map palette, including transparent out-of-map pixels, while both map formats share the same deterministic adjacent-room color policy.

## Changes

- render floor, wall, room, and outside pixels with the V1 palette
- make pixels outside the map transparent
- assign room colors deterministically while avoiding the same color for adjacent rooms where possible
- reset V1 palette state per render so repeated and concurrent renders stay independent
- keep this slice limited to map parsing/rasterization; it has no trait or transport changes

## Before / after

The same synthetic occupancy grid is rendered on `main` and on this branch.
The checkerboard in the after image denotes transparency.

| Before | After |
| --- | --- |
| Before: Q10-specific colors and opaque outside pixels | After: V1 colors and transparent outside pixels |

## Validation

- `pytest -q` — 882 passed, 86 snapshots passed
- `pre-commit run --files ...` — all hooks passed

## Stack

1. **This PR:** base raster and room palette
2. #903: V1 overlay renderer
3. #904: docked-state composition

The transport correction is independent in #901. #900 remains the aggregate real-device/Home Assistant before-and-after reference.

### DIFF
--- a/roborock/map/map_parser.py
+++ b/roborock/map/map_parser.py
@@ -128,8 +128,10 @@ class _AdjacencyAwareRoborockImageParser(RoborockImageParser):
     ):
         """Assign non-conflicting room colors before the V1 image pass."""
         with self._palette_lock:
-            self._room_palette.cached_room_colors.clear()
-            self._room_palette.cached_room_colors.update(self._base_room_colors)
+            # cached_room_colors is a read-only property, so reset its dict in place.
+            cached_room_colors = self._room_palette.cached_room_colors
+            cached_room_colors.clear()
+            cached_room_colors.update(self._base_room_colors)
 
             if self._recolor_rooms:
 
@@ -140,8 +142,9 @@ class _AdjacencyAwareRoborockImageParser(RoborockImageParser):
 
                 room_colors = adjacency_aware_room_colors(raw_data, width, self._room_palette, room_id)
                 for number, color in room_colors.items():
-                    self._room_palette.cached_room_colors[number] = color
-                    self._room_palette.cached_room_colors[str(number)] = color
+                    # ColorsPalette caches both forms for get_room_color(str | int).
+                    cached_room_colors[number] = color
+                    cached_room_colors[str(number)] = color
             return super().parse(raw_data, width, height, carpet_map, removed_map)
 
 
--- a/roborock/map/map_parser.py
+++ b/roborock/map/map_parser.py
@@ -2,6 +2,7 @@
 
 import io
 import logging
+import threading
 from dataclasses import dataclass, field
 
 from vacuum_map_parser_base.config.color import ColorsPalette, SupportedColor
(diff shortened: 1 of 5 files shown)


roborock/map/map_parser.py
```diff
@@ -98,6 +102,49 @@ def parse(self, map_bytes: bytes) -> ParsedMapData | None:
...
+
+                def room_id(value: int) -> int | None:
+                    if value in (self.MAP_OUTSIDE, self.MAP_WALL, self.MAP_INSIDE, self.MAP_SCAN):
+                        return None
+                    return self._get_room_number(value) if value & 0x07 == 0x07 else None
+
+                room_colors = adjacency_aware_room_colors(raw_data, width, self._room_palette, room_id)
+                for number, color in room_colors.items():
+                    self._room_palette.cached_room_colors[number] = color
+                    self._room_palette.cached_room_colors[str(number)] = color
```
allenporter: I dont see any code that uses strings for this. Can you explain what's going on here?

otherwise i think tis can be:

```
self._room_palette.cached_room_colors.update(room_colors)
```

roborock/map/map_parser.py
```diff
@@ -98,6 +102,49 @@ def parse(self, map_bytes: bytes) -> ParsedMapData | None:
...
+        raw_data: bytes,
+        width: int,
+        height: int,
+        carpet_map: set[int] | None,
+        removed_map: set[int] | None = None,
+    ):
+        """Assign non-conflicting room colors before the V1 image pass."""
+        with self._palette_lock:
+            self._room_palette.cached_room_colors.clear()
+            self._room_palette.cached_room_colors.update(self._base_room_colors)
```
allenporter: Does this work as a one-liner?

```suggestion
            self._room_palette.cached_room_colors = {**self._base_room_colors}
```

allenporter (changes requested):
Looks good to me, just some small nits

roborock/map/map_parser.py
tubededentifrice: ColorsPalette.get_room_color accepts str | int and intentionally caches both key forms. Since the cache is pre-populated, updating only the integer entries would leave the corresponding string entries at their original colors, so get_room_color("12") could disagree with get_room_color(12) after recoloring. I kept both assignments, made the cache contract explicit in a comment, and added a regression assertion for the two key forms in e44a7ae.

roborock/map/map_parser.py
tubededentifrice: That one-liner does not work because cached_room_colors is a getter-only property; assigning to it raises AttributeError. The clear/update pair intentionally resets the existing dictionary in place. I introduced a local variable and added a comment in e44a7ae to make that constraint clearer.

allenporter: Ah ok. So we're mutating a property meant to be readonly. This seems to me like it is abusing this API, and not really whats intended by returning a readonly property.

I want to get this improvement in, but I don't think this is sustainable. We can discuss further in discord how we want to fix this.

allenporter (approved):

PR #904, opened 2026-07-25
fix: show docked Q10 robot on map

tubededentifrice:
## Summary

The Q10 map trait now uses map-specific DPS data when it renders a map. The map shows the robot at the dock when the robot has no visible live trace.

## Changes

- keep dock status and map overlays in one private map DPS model
- use named dock-state values for charging and bin emptying
- keep all received trace packets unchanged
- omit the trace from the rendered image when the robot is at the dock
- render the robot at the saved dock position
- report a DPS update only when a model value changes
- keep `Q10PropertiesApi` independent of map rendering details

## Before and after

These images are from the Home Assistant map card in #900.

| Before | After |
| --- | --- |
| Before: Q10 map without the docked robot state | After: docked Q10 robot rendered at the saved dock position |

## Validation

- `pytest -q` — 895 passed, 86 snapshots passed
- `pre-commit run --files ...` — all hooks passed

## Review scope

#902 and #903 are merged. This PR is based directly on `main` and contains one dock-state commit. The transport correction is in #901. PR #900 is the full before-and-after reference.

### DIFF
--- a/roborock/devices/traits/b01/q10/map.py
+++ b/roborock/devices/traits/b01/q10/map.py
@@ -4,12 +4,12 @@ Map-related state arrives on three independent streams:
 
 * map packets are decoded from map-protocol responses;
 * trace packets are decoded from trace-protocol responses;
-* restricted zones and virtual walls arrive as ordinary DPS values.
+* restricted zones, virtual walls and dock state arrive as ordinary DPS values.
 
-``MapDpsTrait`` owns the low-level DPS read model. ``MapContentTrait`` depends
-on it and combines that state with the latest map/trace packets through the pure
-functions in :mod:`roborock.map.b01_q10_render`. The high-level trait keeps only
-the latest value from each source and one replace-whole rendered image;
+``MapDpsTrait`` owns the low-level map-specific DPS read model.
+``MapContentTrait`` combines that state with the latest map/trace packets
+through the pure functions in :mod:`roborock.map.b01_q10_render`. The high-level
+trait keeps only the latest value from each source and one replace-whole image;
 calibration, path placement and overlay placement remain inside the renderer.
 """
 
@@ -18,7 +18,7 @@ from dataclasses import dataclass, field
 from typing import Any
 
 from roborock.data import RoborockBase
-from roborock.data.b01_q10.b01_q10_code_mappings import B01_Q10_DP
+from roborock.data.b01_q10.b01_q10_code_mappings import B01_Q10_DP, YXDeviceState
 from roborock.devices.traits.common import DpsDataConverter, TraitUpdateListener
 from roborock.exceptions import RoborockException
 from roborock.map.b01_q10_map_parser import (
@@ -34,12 +34,14 @@ from roborock.map.b01_q10_render import Q10MapOverlays, render_q10_map
 from .common import UpdatableTrait
 
 _LOGGER = logging.getLogger(__name__)
+_DOCKED_STATES = {YXDeviceState.CHARGING, YXDeviceState.EMPTYING_THE_BIN}
 
 
 @dataclass
 class MapDps(RoborockBase):
     """Low-level map values delivered in the Q10 DPS stream."""
 
+    status: YXDeviceState | None = field(default=None, metadata={"dps": B01_Q10_DP.STATUS})
     restricted_zone_up: str | None = field(default=None, metadata={"dps": B01_Q10_DP.RESTRICTED_ZONE_UP})
     virtual_wall_up: str | None = field(default=None, metadata={"dps": B01_Q10_DP.VIRTUAL_WALL_UP})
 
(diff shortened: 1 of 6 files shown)


roborock/devices/traits/b01/q10/map.py
```diff
@@ -124,17 +134,52 @@ def update_from_map_packet(self, packet: Q10MapPacket) -> None:
...
+        # A late packet from the completed clean cannot move a robot that the
+        # status stream already confirmed is docked.
+        self._trace_packet = None if self._robot_at_dock else packet
         self._render()
         self._notify_update()
 
-    def _map_dps_updated(self) -> None:
-        """Render after the low-level DPS source changes."""
-        if self._map_packet is None:
+    def _begin_source_update(self) -> None:
```
allenporter: I don't really love this interaction between the main properties api and this trait. At a high level what are we trying to accomplish here? Naively it seems like the AI is worried about a theoretical race or something but i can't tell. But basically, i don't think this is the solution we want to use.

allenporter (changes requested):

roborock/devices/traits/b01/q10/map.py
tubededentifrice: was trying to avoid two renders when one DPS payload updates both dock status and map overlays, changed

allenporter: So: Do we always want to wait for both to change? Do both arrive in a single message? Schedule a timer to perform the update after a "quiet period"?

roborock/devices/traits/b01/q10/map.py
```diff
@@ -86,6 +102,7 @@ def __init__(
         TraitUpdateListener.__init__(self, logger=_LOGGER)
         self._config = map_parser_config or B01Q10MapParserConfig()
         self._map_dps = map_dps
+        self._robot_at_dock = self._map_dps.robot_at_dock
```
allenporter: I don't think we want to copy this since it will certainly be stale, so i think we need to always refer to the trait if we want to check it.

tests/devices/traits/b01/q10/test_map.py
```diff
@@ -304,11 +304,110 @@ def test_map_dps_update_without_map_does_not_notify_map_content() -> None:
...
     notified = []
     trait.add_update_listener(lambda: notified.append(True))
 
     map_dps.update_from_dps({B01_Q10_DP.BATTERY: 50})
 
     assert map_dps.overlays == Q10MapOverlays()
     assert not notified
+
+
+# --- Dock state --------------------------------------------------------------
```
allenporter: Not sure this is needed

tests/devices/traits/b01/q10/test_map.py
```diff
@@ -304,11 +304,110 @@ def test_map_dps_update_without_map_does_not_notify_map_content() -> None:
...
+
+
+def test_combined_status_and_overlay_update_renders_once() -> None:
+    """One map DPS update publishes the complete new rendering state."""
+    map_dps = MapDpsTrait()
+    trait = _map_trait(map_dps)
+    notified: list[None] = []
+    trait.add_update_listener(lambda: notified.append(None))
+
+    with patch(
```
allenporter: Seems like this can be an always available fixture

tests/devices/traits/b01/q10/test_map.py
```diff
@@ -304,11 +304,110 @@ def test_map_dps_update_without_map_does_not_notify_map_content() -> None:
...
+
+# --- Dock state --------------------------------------------------------------
+
+
+def test_charging_status_renders_robot_at_dock() -> None:
+    """Charging status adds the idle robot marker without inventing a path."""
+    map_dps = MapDpsTrait()
+    trait = _map_trait(map_dps)
+    packet = parse_map_packet(FIXTURE.read_bytes())
+    notified: list[None] = []
```
allenporter: Use an `asyncio.Event`?

roborock/devices/traits/b01/q10/map.py
```diff
@@ -124,13 +141,19 @@ def update_from_map_packet(self, packet: Q10MapPacket) -> None:
 
     def update_from_trace_packet(self, packet: Q10TracePacket) -> None:
         """Store a trace-protocol update and render the latest sources."""
-        self._trace_packet = packet
+        # A late packet from the completed clean cannot move a robot that the
```
allenporter: The intent of this code is to adjust what is rendered. So instead of mutating the state to clear trace packets and worrying about out of order stuff, we can just change what we pass into the render function. That is, the trace packet received is still completely valid, so we shouldn't clear it just because we want to change how its rendered on the map.  (What we render can be different than the raw data values).

Instead:
- Revert the __init__ change
- Revert the update_from_trace_packet change
- Revert the  _map_dps_updated change
- In _render add: `self._trace_packet if not self._robot_at_dock else None`

I think that is equivalent?

roborock/devices/traits/b01/q10/map.py
```diff
@@ -59,22 +61,36 @@ def overlays(self) -> Q10MapOverlays:
...
+        """Update one coherent snapshot of the DPS inputs used by the map."""
+        previous_robot_at_dock = self.robot_at_dock
+        previous_restricted_zones = self.restricted_zone_up
+        previous_virtual_walls = self.virtual_wall_up
         if not self._CONVERTER.update_from_dps(self, decoded_dps):
             return
-        self._overlays = Q10MapOverlays(
-            zones=tuple(parse_zone_blob(self.restricted_zone_up)),
-            virtual_walls=tuple(parse_virtual_wall_blob(self.virtual_wall_up)),
+        overlays_changed = (
```
allenporter: I think we could remove a lot of the logic added in this class if `update_from_dps` was modified to return `False` is there was no actual change in values converted. What do you think about that? I think nearly everything likely wants this behavior, but this feels like too much code to get it so looking for something more succint.

allenporter (changes requested):

allenporter (approved):

PR #908, opened 2026-07-31
feat: add Q10 zone, position, and goto support

hmmbob:
## Summary

- expose Q10 `robot_position` in the common Roborock millimetre coordinate space
- add native zone cleaning with typed points and protocol-layer encoding
- emulate goto with the app's minimum 40 x 40 cm zone
- isolate goto ownership and decisions in a push-driven `GotoAction`

## Why

Q10 firmware has no V1 `app_goto_target` command. Goto therefore starts a small zone, pauses within 20 cm of the target, and stops after five minutes only if it still owns that cleaning session. It uses streamed map and status updates without extra polling.

The public API uses `Q10RoborockPoint`; Q10 trace and vector coordinates remain implementation details. The point model matches the one prepared in #934 so that its follow-up work can build on this PR.

## Hardware validation

Validated on `roborock.vacuum.ss07` firmware 03.11.24. Target `(29900, 28650)` paused at `(30020, 28705)`, 13.2 cm away, and the robot remained at the destination.

## Tests

- 89 focused Q10 tests passed
- all pre-commit hooks passed, including Ruff, mypy and codespell
- 980 additional non-E2E/non-MQTT tests passed locally; one unrelated Windows-only URL-normalization assertion failed

## Downstream

- Home Assistant: https://github.com/home-assistant/core/pull/177780


roborock/devices/traits/b01/q10/map.py
```diff
@@ -111,6 +112,21 @@ def robot_position(self) -> Q10Point | None:
         """Current position for live status and caller-rendered map overlays."""
         return self._trace_packet.robot_position if self._trace_packet else None
 
+    @property
+    def roborock_position(self) -> Q10Point | None:
+        """Current position in the common Roborock millimetre coordinate space."""
```
allenporter, line 117: You have information written in the PR, but it should also be here (e.g. Whats the difference between `roborock_positiion` and `roborock_position` from an end user perspective? how should I be using either of these?)

But really, I would advocate for taking a more opinionated stance: Should we make a breaking change here and always use the common coordinate system?  We can consider the Q10 specific thing an implementation.

roborock/devices/traits/b01/q10/vacuum.py
```diff
@@ -55,25 +227,87 @@ async def clean_segments(self, segment_ids: list[int]) -> None:
             # "parameters" -- the firmware only accepts that exact key.
             params={"cmd": YXDeviceCleanTask.ELECTORAL.code, "clean_paramters": segment_ids},
         )
+        self.cancel_goto()
+
+    async def clean_zone(
+        self,
+        x1: int,
```
allenporter, line 234: I think we should:

- Define a new variation of Q10Point that is in roborock coordinate space (and we say Q10 is in trace space) e.g. Q10RoborockPoint (or some better name). This can be used for input here and output (e.g. exposed current position)
- Accept Q10RoborockPoint (or new name here).
- Q10Point and Q10RoborockPoint can have functions for converting between each other (e.g. handling the roborock to vector coordinate functions)
- Then for our internal start command params add a `CleanParams` dataclass object that holds the roborock points or q10 points, and handles sorting them.
- Move the struct packing in roborock/protocols/b01_q10_protocol.py e.g. `def encode_clean_params(params: CleanParams) -> str`

roborock/devices/traits/b01/q10/vacuum.py
```diff
@@ -55,25 +227,87 @@ async def clean_segments(self, segment_ids: list[int]) -> None:
...
+                "cmd": YXDeviceCleanTask.DIVIDE_AREAS.code,
+                # "clean_paramters" is the spelling required by the firmware.
+                "clean_paramters": encoded_zone,
+            },
+        )
+        self.cancel_goto()
+
+    async def goto_position(self, x: int, y: int) -> None:
+        """Move to a coordinate using an owned 40 cm zone-clean task."""
+        if (position := self._map.roborock_position) is not None and hypot(
```
allenporter, line 255: Will _async_monitor_goto_target effectively do the same thing?

roborock/devices/traits/b01/q10/vacuum.py
```diff
@@ -17,9 +69,128 @@ class VacuumTrait:
...
+        self._goto_trace_sequence = None
+
+    def cancel_goto(self) -> None:
+        """Cancel monitoring for an emulated goto replaced by another command."""
+        if self._goto_monitor_task is not None:
+            self._goto_monitor_task.cancel()
+            self._goto_monitor_task = None
+            self._goto_trace_sequence = None
+
+    async def _async_monitor_goto_target(
```
allenporter, line 104: A lot of the logic in this method is related to understanding if the current task is allowed to change state -- and it's intermixed with the other business logic for trying to coordinate the vacuum move. I think we can simplify this by having a separate standalone object that manages a single goto lifecycle (e.g. `GotoAction`.  The action can handle the logic for the next step to take given the information it has (e.g. checking current status of the vaccum, and deciding when to stop)  but keep the actions handled by vacuum.py which manages the lifecycle of `GotoAction` -- it would register listeners/callbacks on the GotoAction then do the work to call the "pause" or "stop" command.

roborock/devices/traits/b01/q10/vacuum.py
```diff
@@ -17,9 +69,128 @@ class VacuumTrait:
...
+        self._goto_monitor_task = None
+        self._goto_trace_sequence = None
+
+    def cancel_goto(self) -> None:
+        """Cancel monitoring for an emulated goto replaced by another command."""
+        if self._goto_monitor_task is not None:
+            self._goto_monitor_task.cancel()
+            self._goto_monitor_task = None
+            self._goto_trace_sequence = None
+
```
allenporter, line 103: Overall I like this approach of listening for map and status events. Are you finding that results are streamed back reliably such that it works without additional polling?

allenporter (changes requested):
Thank you for this great contribution, see my comments below.

hmmbob:
@allenporter Thanks for the detailed review. I used Codex to help build this, so I want to be transparent about that. I do understand the functional side of the implementation (it actually developed this based on my questions regarding robot positioning and if we could use the zoned cleaning as alternative for the `goto`) and have tested it on my own Q10, but I’m happy to have Codex rework the internals along the lines you suggested. It told me already it understood your ideas 😉 

Codex:
> My understanding is that you’d like the public API to use one common Roborock coordinate space, with typed points and the Q10-specific conversions treated as implementation details. The clean parameters and binary encoding should move to the protocol layer, while the goto lifecycle should be separated into something like a GotoAction, leaving VacuumTrait responsible for listeners and sending the actual commands. Please correct me if I’ve misunderstood any part of that.

Regarding the push updates: on my ss07 running firmware 03.11.24, the map/trace and status updates were streamed reliably during the live goto tests. The robot reached the target and was paused based solely on those updates; we did not use any additional polling or periodic refresh() calls. I was actually quite happy we got this approach to work 😃 

There is also a downstream Home Assistant PR using this API: [home-assistant/core#177780](https://github.com/home-assistant/core/pull/177780). You may want to review that as well, particularly to make sure the public API introduced here fits the way Home Assistant consumes it.

If the above matches what you have in mind, I’ll have Codex work through the individual comments and update the PR accordingly. Also, if you prefer to take over and implement it yourself, I'd be ok with that as well. As said, I am not a dev, but a Q10 (ss07) owner happy to have my tokens do some useful work so don't hold back on commenting. I'm actually having quite some fun guiding Codex doing this.

roborock/devices/traits/b01/q10/vacuum.py
hmmbob: Not entirely, as I understand it. The monitor pauses the goto it already owns when the robot reaches that goto’s original target. This early check also avoids starting a new zone when the robot is already at the requested position. It additionally handles a new goto request for the robot’s current position while another owned goto is still active, even if the previous target was different.

I think the comment below addresses this as well, right?

roborock/devices/traits/b01/q10/vacuum.py
hmmbob: Yes, at least on my ss07 running firmware 03.11.24. During the live goto tests, the trace and status updates arrived reliably enough to detect arrival and pause the robot. We did not use periodic polling or call refresh() from the goto monitor.

allenporter:
I'm fine with you continuing. there is a merge conflict.

roborock/devices/traits/b01/q10/map.py
hmmbob: Done. robot_position now directly returns Q10RoborockPoint in the common millimetre coordinate space, and the separate roborock_position property is gone. The Q10 trace representation is now only used as an implementation detail.

roborock/devices/traits/b01/q10/vacuum.py
hmmbob: Done. clean_zone and goto_position now accept Q10RoborockPoint values. CleanParams owns the corner sorting, and encode_clean_params plus the struct packing now live in b01_q10_protocol.py. I also used the same point model prepared in #934 so that work can rebase onto this.

roborock/devices/traits/b01/q10/vacuum.py
hmmbob: Kept this as a separate check. It avoids starting a new zone when the robot is already at the requested target; if the current session is an owned goto, it also pauses that session before clearing it.

roborock/devices/traits/b01/q10/vacuum.py
hmmbob: Done. GotoAction is now a standalone state machine with no tasks or device commands. VacuumTrait feeds it push-derived snapshots, listens for pause/stop/complete decisions, and owns the timeout, retry, and actual command calls.

hmmbob:
@allenporter This is rebased onto current main and I've worked through and resolved all five review threads. The point API is aligned with #934, and the final safety pass adds explicit confirmed zone ownership before goto may pause. 89 focused tests and all pre-commit hooks pass. GitHub won't let me re-request your review on this fork; the workflow is also waiting for maintainer approval. Ready for another look when you have time.

allenporter (commented):
Thank you for thoroughly working through and addressing all the previous review feedback!

We ran an experimental agentic code review pass on the updated diff:
- The separation of `GotoAction` into a pure state machine cleanly isolates lifecycle decisions from command execution and task management.
- Adopting `Q10RoborockPoint` in common millimetre coordinates provides a consistent, clean consumer-facing API across device families.
- Struct packing and zone payload encoding are properly isolated in `b01_q10_protocol.py`.
- Teardown and cancellation in `VacuumTrait.close()` are cleanly hooked into `Q10PropertiesApi.close()`.
- Full test suite and CI matrix are completely green.

I'll do another human review pass shortly to give it a final look.

allenporter (approved):

allenporter:
Thank you @hmmbob, great work!

PR #909, opened 2026-07-31
Add support for Q10 diagnostics data

allenporter:
This PR exposes state-holding properties for Q10 devices in `diagnostic_data()`. Maps are included but large binary fields (`image_content`, `map_data`) are filtered out.


roborock/devices/traits/b01/q10/map.py
```diff
@@ -149,3 +149,18 @@ def _render(self) -> None:
         except RoborockException as ex:
             _LOGGER.debug("Failed to render Q10 map packet: %s", ex)
             self._image_content = None
+
+    def as_dict(self, exclude: set[str] | None = None) -> dict[str, Any]:
+        """Return the trait data as a dictionary, excluding large binary data."""
+        import dataclasses
```
Lash-L: Should probably just be top of the file right?

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -160,6 +162,16 @@ def _handle_message(self, message: Q10Message) -> None:
             for trait in self._updatable_traits:
                 trait.update_from_dps(message.dps)
 
+    def as_dict(self) -> dict[str, Any]:
+        """Return the trait data as a dictionary."""
+        result: dict[str, Any] = {}
+        for name, value in self.__dict__.items():
+            if isinstance(value, RoborockBase):
+                result[name] = value.as_dict()
```
Lash-L: This is going to grab all of the traits right? This will duplicate map I think? As MapDpsTrait(MapDps, UpdatableTrait) -> MapDps(RoborockBase)

Any of these traits have info we don't want in diagnostics? Not as familiar with them but there might be ssid. I don't recall if diagnostics auto filters out a lot of this

Lash-L (changes requested):

Lash-L:
Previously missed this one - sorry!

Lash-L (approved):

PR #917, opened 2026-08-11
fix: fix Q7 code mapping issue for workmode

Lash-L:
Should handle: https://github.com/home-assistant/core/issues/177387

adds Unknown similar to how we do elsewhere. will require a HA mapping update too


allenporter (approved):
Co-pilot is not wrong with its comments about this overriding the default from code behavior and interaction with optional -- but  i think this makes sense to support, allowing some enums to map to an unknown value. This case also seems to make sense (allowing the warning but still working).  Approving.

As we look to other cases: Do you have a sense of what enums should support this kind of fallback vs cases that should stay strict?

PR #919, opened 2026-08-15
feat: add dust collection settings to the B01 Q7 api

akj:
## What

Adds two setters to `Q7PropertiesApi` for the auto-empty dock, following the existing `prop.set` setter conventions (`set_child_lock`, `set_volume`):

- `set_dust_collection(enabled)` — writes `dust_auto_state`
- `set_dust_collection_frequency(cleans)` — writes `dust_frequency` (1 = empty after every clean); validates >= 1

Plus parametrized tests in `tests/devices/traits/b01/q7/test_init.py` matching the existing setter tests.

## Verified on real hardware

Tested against a Roborock Q7 M5+ (`roborock.vacuum.sc05`, fw 03.01.80, US region):

- `dust_auto_state` round-trips 1 → 0 → 1 via `set_dust_collection`, confirmed by `GET_PROP` read-back after each write.
- `dust_frequency` round-trips 1 → 2 → 1 via `set_dust_collection_frequency`, same method.

## Deliberately not included: manual "empty now"

Writing `dust_action` via `prop.set` is **rejected by the device** with code 1 (`B01 command failed with code 1`), so `dust_action` appears to be read-only state. The product schema for sc05 exposes write-only DPS `201 start_clean_task`, `202 start_back_dock_task`, `203 start_dock_task` — a manual empty trigger most likely goes through `203 start_dock_task`, which the Q7 channel doesn't support sending yet. Happy to test candidate encodings against my device in a follow-up if someone knows the expected payload shape.

Side observation from the same session, can file separately: this healthy, docked device steadily reports `fault: "fault_2103"` (unmapped placeholder in `B01Fault`, adjacent to 2102 `cleaning_complete`) — the 2103 reading appeared right after a completed clean while charging, so it looks like another benign 2100-series notification rather than a real fault.


allenporter (commented):
This passes our initial agentic review pass:
- Follows the established `Q7PropertiesApi` property-setter pattern using `RoborockB01Props` and includes input validation.
- Mirroring unit tests in `tests/devices/traits/b01/q7/test_init.py` with clean parametrization and zero-RPC assertions on invalid input.
- All CI checks and test suites are passing.

I will do a follow-up human review pass shortly.

allenporter (approved):

PR #921, opened 2026-08-16
fix: name B01 Q7 code 2103 and document dock dust-collection state fields

akj:
Small follow-up to #919, based on a live end-to-end capture of a completed clean → docking → auto-empty on a Q7 M5+ (`roborock.vacuum.sc05`, fw 03.01.80). Full evidence timeline is in [this #919 comment](https://github.com/Python-roborock/python-roborock/pull/919#issuecomment-5308270383).

**Changes:**
- `B01Fault.F_2103`: placeholder `fault_2103` → `cleaning_complete_docked`. Observed appearing the moment a completed clean's dock + dust collection finished, persisting while docked-idle; it follows 2102 (`cleaning_complete`, seen exactly at completion/return) and is a benign task notification, not an error.
- `B01Props.station_act` / `dust_action`: documented hw-observed dock-emptying semantics — `station_act=3` + `dust_action=1` during the ~30s auto-empty cycle, both 0 otherwise; `dust_action` is read-only (`prop.set` rejected with code 1).

These two fields are enough for a "dock emptying" state/sensor downstream; the clean record's `record_dust_num` additionally tracks empties per clean.

Left 2105 as a placeholder — I've observed it while long-idle on the dock (possibly a later idle/charged notice) but don't have a clean state transition for it yet; happy to capture more if useful.

### DIFF
--- a/roborock/data/b01_q7/b01_q7_code_mappings.py
+++ b/roborock/data/b01_q7/b01_q7_code_mappings.py
@@ -290,7 +290,12 @@ class B01Fault(RoborockModeEnum):
     # charging.
     F_2101 = ("fault_2101", 2101)
     F_2102 = ("cleaning_complete", 2102)  # Cleaning completed. Returning to the dock.
-    F_2103 = ("fault_2103", 2103)
+    F_2103 = (
+        "cleaning_complete_docked",
+        2103,
+    )  # Task finished; robot docked (follows 2102 once docking and any auto-empty complete). hw-confirmed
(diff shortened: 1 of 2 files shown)


allenporter (approved):
LGTM! Thank you for capturing the hardware trace and properly naming code 2103 as a `RoborockModeEnum` tuple along with documenting the dock dust collection fields.

As a quick thought for a follow-up PR: we could formalize those integer states (`station_act`, `dust_action`) into proper `RoborockEnum` or `RoborockModeEnum` mappings rather than leaving them as bare `int` fields.

PR #922, opened 2026-08-21
fix: fix charging for off peak

Lash-L:
This should handle: https://github.com/home-assistant/core/issues/169151

Need a matching HA side change to use the new property.


roborock/data/v1/v1_containers.py
```diff
@@ -224,6 +224,10 @@ def dock_state(self) -> RoborockDockState:
         newer off-peak charging logic seamlessly while maintaining backwards compatibility
         with older devices.
         """
+        return self.get_dock_state()
```
allenporter, line 227: why is it ok to have two versions, one with an extra input and one not? I think it implies that this method be determined by the container, since it needs device features as input.

I would say: (1) the current `dock_state` property needs to be deprecated and (2) the method below should be a function that takes as input `state`, `battery`, and `is_supported_valley_electricity` and is called from the trait.

(What is in this PR works in practice because the trait overrides the `dock_state` method, so I can approve, but i think this should be cleaned up)

allenporter (approved):

PR #927, opened 2026-08-26
docs(q7): document live map pushes and cover them in tests

ximex:
Follow-up on the Q7 (B01) live map pushes: the runtime behaviour is unchanged, this only makes the push path discoverable and pins it down in tests.

### Docs

`MapContentTrait`'s module docstring describes a pure pull contract — `refresh()` does I/O, `parse_map_content()` reparses, fields "are then readable". That no longer holds: `update_from_push()` writes the same cached fields from the MQTT receive callback and notifies listeners, so `image_content` / `map_data` / `raw_api_response` can change without the reader calling anything. The docstring now says so.

`docs/DEVICES.md` listed only `device.b01_q7_properties` under B01 Q7, with no hint that maps arrive on their own. A consumer reading it finds just the pull API and builds a polling loop for something the device streams by itself — so the entry point (`map_content.add_update_listener(cb)`) and the fields to read are now named, the way the Q10 section already does for its commands.

### Tests

Two gaps in `test_q7_map_content_updates_from_push`:

- `map_data` was not asserted, even though it carries the live robot pose and cleaning path. `_parse_and_store()` sets three fields; dropping the `map_data` assignment left the test green. Asserted by identity, since two empty `MapData` instances compare equal and equality would also accept a stale or copied object.
- Nothing asserted that the push path publishes no command. That is the actual claim of the feature: `refresh()` needs `map_trait.current_map_id` and costs two round trips (the refresh tests assert `len(published_commands) == 2`), while a push needs neither. Without the assertion the test would still pass if `update_from_push()` or `start()` started issuing RPCs.

`uv run pytest tests/devices/traits/b01/q7 tests/devices/rpc/test_b01_q7_channel.py` → 55 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/DEVICES.md
+++ b/docs/DEVICES.md
@@ -19,6 +19,7 @@ Cloud and Network.
         *   Use `device.b01_q10_properties.vacuum` to access vacuum commands (start, pause, stop, dock, empty dustbin, set clean mode, set fan level).
         *   Use `device.b01_q10_properties.command.send()` for raw DP commands.
     *   **Vacuums (B01 Q7)**: Use `device.b01_q7_properties` for Q7 series devices.
+        *   The device streams full map frames on its own while cleaning, so the rendered map stays current without polling or a heartbeat. Register `device.b01_q7_properties.map_content.add_update_listener(cb)` and read `image_content` / `map_data` when notified.
 
 ## Background: Understanding Device Protocols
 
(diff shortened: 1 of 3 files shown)


allenporter (approved):
LGTM! Thank you for clarifying the live map push semantics in the docs and pinning down the zero-RPC / identity assertions in tests.

PR #928, opened 2026-08-28
refactor: Migrate to zmqtt

borisalekseev:
## Summary

Replace the `aiomqtt`/Paho-based MQTT transport with `zmqtt` and move MQTT integration tests from packet-level monkeypatching to a real EMQX broker.

## Why

The previous adapter owned a substantial amount of MQTT client lifecycle code: reconnect loops, exponential backoff, connection recovery, subscription restoration, and delayed unsubscribe bookkeeping. These are routine MQTT concerns, but they are difficult to reproduce faithfully in application code and easy to get wrong around cancellation and shutdown.

`zmqtt` handles those concerns inside the client library. It provides a native asyncio API, automatic reconnect with configurable backoff, transparent restoration of active subscriptions, bounded subscription queues, and cancellation-aware protocol cleanup. This leaves python-roborock responsible for its actual domain behavior: Roborock authentication, topic construction, payload encoding/decoding, callback fan-out, health management, and device-level request/response correlation.

This is also a proven integration path rather than a project-specific experiment: [`zmqtt`](https://github.com/faststream-community/zMQTT) is used as the MQTT client underneath [FastStream](https://github.com/ag2ai/faststream).

## What changed

- Replaced the direct `aiomqtt` and Paho dependencies with `zmqtt`.
- Delegated unexpected connection recovery, reconnect backoff, and subscription restoration to `zmqtt`.
- Preserved the public `MqttQos` API and the Roborock-specific request/response behavior.
- Kept one broker subscription per topic for the lifetime of the shared MQTT session and fan out messages to active callbacks locally.
- Simplified shutdown while preserving the required MQTT ordering: finish `UNSUBSCRIBE`/`UNSUBACK` before disconnecting the client.
- Removed the old reconnect loop, idle-unsubscribe timers, packet generators, and protocol-level client fixtures.
- Redacted MQTT topics and stopped logging complete payloads/messages.

## Real-broker tests

MQTT session and device-manager E2E tests now run against EMQX over TCP. The test suite exercises the actual `zmqtt` codec and lifecycle instead of monkeypatching Paho callbacks or generating synthetic MQTT packets in process.

The repository now includes a Compose service for local development, and the GitHub Actions test matrix starts the same EMQX version as a service container. This gives the tests real `CONNECT`, `SUBSCRIBE`, `PUBLISH`, `UNSUBSCRIBE`, acknowledgement, and disconnect behavior while keeping focused mocks only for error injection that is impractical to trigger through a broker.

## Validation

- Full test suite on Python 3.11 and Python 3.14 against EMQX
- Snapshot tests against real broker traffic
- Pre-commit checks, mypy, Ruff, and lock-file validation
- Package build


allenporter (changes requested):
Hello! Very appreciative of this contribution. Definitely interested in a better MQTT implementation that natively supports asyncio. I have not yet reviewed zmqtt but it sounds very appealing.

Can we slow this down a little bit? We have `MqttSession` which abstracts the details of the implementation of the MqttSession so I think what could make sense here is a  smaller change that adds a different implementation.  We can swap in at a call site to experiment with.  Also there seems to be a lot of, i'm assuming, ai generated text changes that just make the review a lot larger with smaller nits or rewordings.

PR #933, opened 2026-08-29
fix: request Q10 maps without starting cleaning

hCoureau:
Related to #767. Built on the stable Q10 vector calibration merged in #932 and released in `7.2.1`.

## Summary

- request the current Q10 map with the read-only `REQUEST_DPS` command
- remove the saved-map-list dependency from current-map refreshes
- keep the one-shot CLI wait bounded around the asynchronous map push

## Why

On observed `roborock.vacuum.ss07` firmware, `dpMultiMap op:get` is not a safe read: it can initiate cleaning or relocation. `REQUEST_DPS` requests the current status/map push without moving or starting the robot and does not require a saved-map ID.

## Validation

- `uv run pytest -q`: **965 passed**, **92 snapshots passed**
- `pre-commit run --all-files`: passed, including Ruff, mypy, codespell, structured-file checks, and private-key detection
- `uv build`: passed for both sdist and wheel
- validated repeatedly on a physical Q10 S5+; current-map pushes arrived without cleaning or relocation and the robot remained docked

No credentials, identifiers, room labels, floorplans, packet captures, or private fixtures are included.

The archive parser, archive traits, obstacles, and current-room work from the earlier revision have been moved to separately validated follow-up branches and will be submitted sequentially as focused PRs.

## Reviewer guide

1. `MapContentTrait.refresh`: the command change
2. `_await_q10_map_push`: removal of the map-list prerequisite
3. map-trait tests: exact command and map-list independence

### DIFF
--- a/roborock/cli.py
+++ b/roborock/cli.py
@@ -594,9 +594,9 @@ async def maps(ctx, device_id: str):
     await _display_v1_trait(context, device_id, lambda v1: v1.maps)
 
 
-# The Q10 publishes its map asynchronously after a dpMultiMap list/get request.
-# Firmware throttles pushes to ~once per 60-70s, so rapid re-requests may not be
-# answered immediately. This bounds how long a one-shot CLI command waits.
+# The Q10 publishes its current map asynchronously after a REQUEST_DPS. Firmware
+# throttles pushes to ~once per 60-70s, so rapid re-requests may not be answered
+# immediately. This bounds how long a one-shot CLI command waits.
 _Q10_MAP_PUSH_TIMEOUT = 30.0
 
 
@@ -609,10 +609,9 @@ async def _await_q10_map_push(
 ) -> bool:
     """Request Q10 map content and wait for usable map-trait state.
 
-    A Q10 needs a saved-map ID before it can request content. The map list and
-    content have independent refresh schedules, so the list is requested only
-    when no ID is stored. The content then arrives as a later ``MAP_RESPONSE``
-    and is published through the standard trait update interface.
+    The read-only ``REQUEST_DPS`` request returns immediately; current map
+    content arrives as a later ``MAP_RESPONSE`` and is published through the
+    standard trait update interface.
     """
     loop = asyncio.get_running_loop()
     updated: asyncio.Future[None] = loop.create_future()
@@ -624,20 +623,6 @@ async def _await_q10_map_push(
     unsub = properties.map.add_update_listener(on_update)
     try:
         async with asyncio.timeout(timeout):
-            if properties.maps.current_map_id is None:
-                map_list_updated: asyncio.Future[None] = loop.create_future()
-
-                def on_map_list_update() -> None:
-                    if properties.maps.current_map_id is not None and not map_list_updated.done():
-                        map_list_updated.set_result(None)
-
-                unsub_maps = properties.maps.add_update_listener(on_map_list_update)
-                try:
-                    await properties.maps.refresh()
-                    if properties.maps.current_map_id is None:
-                        await map_list_updated
-                finally:
-                    unsub_maps()
             await properties.map.refresh()
             await updated
         return True
--- a/roborock/devices/traits/b01/q10/map.py
+++ b/roborock/devices/traits/b01/q10/map.py
@@ -7,10 +7,10 @@ Map-related state arrives on three independent streams:
 * restricted zones, virtual walls and dock state arrive as ordinary DPS values.
 
 ``MapDpsTrait`` owns the low-level map-specific DPS read model.
-``MapContentTrait`` uses a stored ID from ``MapsTrait`` only when it requests
-content. It combines the latest map and trace packets with the map DPS state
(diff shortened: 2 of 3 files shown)


roborock/devices/traits/b01/q10/clean_history.py
```diff
@@ -141,6 +164,49 @@ async def refresh(self) -> None:
...
         )
 
+    async def refresh_detail(self, record: Q10CleanRecord) -> None:
+        """Request the saved map and path for one clean record.
+
+        The complete 12-field raw record is the firmware's detail identifier;
+        the shorter human-facing record ID is not accepted. Only one request
+        may be outstanding because ``03 01`` responses carry no correlation ID.
+        """
+        if self._command is None:
```
allenporter: i don't think this is a state we need to worry about

roborock/devices/traits/b01/q10/clean_history.py
```diff
@@ -151,6 +217,25 @@ def update_from_dps(self, decoded_dps: dict[B01_Q10_DP, Any]) -> None:
...
+        self._pending_detail_record = None
+        self.detail_packet = packet
+        try:
+            self.detail_image_content = render_q10_map(
+                packet,
+                packet.historical_trace,
+                Q10MapOverlays(),
+                config=self._map_parser_config,
+            )
+        except Exception:
```
allenporter: lets not catch such a broad exception, but instead catch a more narrow type actually raised. If its a lot of exceptions then we may need to push some exception handling into this module.

roborock/devices/traits/b01/q10/__init__.py
```diff
@@ -92,7 +97,12 @@ class Q10PropertiesApi(Trait):
     clean_history: CleanHistoryTrait
     """Trait for fetching the device clean-record history (``dpCleanRecord``)."""
 
-    def __init__(self, channel: B01Q10Channel) -> None:
+    def __init__(
+        self,
+        channel: B01Q10Channel,
+        *,
+        map_parser_config: B01Q10MapParserConfig | None = None,
```
allenporter: This doesn't seem like an optional argument, is there a reason not to make required?

roborock/map/b01_q10_map_parser.py
```diff
@@ -194,9 +204,17 @@ def charger_pixels(self) -> tuple[float, float] | None:
         )
 
 
+class Q10MapPacketKind(Enum):
```
allenporter: My understanding is that these values correspond with:
```
MAP_PACKET_MARKER = b"\x01\x01"
TRACE_PACKET_MARKER = b"\x02\x01"
CLEAN_RECORD_MAP_PACKET_MARKER = b"\x03\x01"
SAVED_MAP_PACKET_MARKER = b"\x04\x01"
```
So the string enum values used in this enum look arbitrary. What I would expect instead is:

```
class Q10MapPacketKind(RoborockEnum):
    MAP_PACKET_MARKER = 1
    TRACE_PACKET_MARKER = 2
    CLEAN_RECORD_MAP_PACKET_MARKER = 3
    SAVED_MAP_PACKET_MARKER = 4
````

And we replace individual constants and use the enum directly. This means we don't need to define these twice with an additional function to do the mapping. The enum string values look arbitrary and they can be dropped i think.

File: roborock/map/b01_q10_map_parser.py [hunk 5]
```diff
@@ -211,6 +229,15 @@ class Q10MapPacket:
     """Carpet mask decoded from the packet tail: a full ``width*height`` grid in
     the same (top-down) pixel space as :attr:`grid`, where a non-zero cell is
     carpet (the value is the carpet kind). ``None`` if the packet carried none."""
+    kind: Q10MapPacketKind = Q10MapPacketKind.CURRENT
```
allenporter: Drop default value and explicit set one? This field seems like it should be first.

roborock/map/b01_q10_map_parser.py
```diff
@@ -211,6 +229,15 @@ class Q10MapPacket:
...
+    historical_trace: "Q10HistoricalTracePacket | None" = None
+    """Cleaning path embedded in a clean-record detail packet, if present."""
+    trailing_bytes: bytes = b""
```
allenporter: where is this used

allenporter (changes requested):
Can we split this ~2k diff into smaller logical chunks to facilitate a faster review? The sheer size of the PR is making it take longer for me set aside a large chunk of time to review (e.g. 30+ minutes) where as having smaller changes we can review quickly could help this get into a release sooner.

Just to be transparent, the primary questions I am trying to answer are:
- 1st priority: How is the end user API/Traits changing (and how can we keep it readable for users etc)
- 2nd priority: Data lifecycle within the traits (and how do we keep state management simple and easy to follow)
- 3rd priority: How is the Map Parser API consumed by the traits changing (how do the map parser API boundaries impact complexity within the Trait)
- Lowest/4th priority: Code internally within the map parser

I think there are a lot of good improvements here, its just a lot being fixed at once. Can you consider how to logically partition some of this?

allenporter (approved):

PR #936, opened 2026-08-31
feat: parse Q10 archived map packets

hCoureau:
Q10 clean-record and saved-map responses use different packet markers from live maps. This adds bounded archive parsing and a `Q10CleanRecordDetail(map, trace)` response so callers receive the map and historical path together without extending the map raster model.

## Implementation

- Classify current, live-trace, clean-record and saved-map markers with `Q10MapPacketKind`. Unknown values resolve to `unknown` and are ignored rather than defaulting to a current map.
- Decode shared raster fields once; clean-record responses compose that map with an optional historical path.
- Keep `Q10HistoricalTracePacket` limited to points and heading. Validate wire version/reserved fields internally and do not expose the undecoded wire value.
- Apply the same conservative stray-leading-point filter used for live paths. Short paths and ordinary origin points are preserved.
- Render the explicitly supplied map and trace without type-dependent branching or silently ignored arguments.
- Remove unused archive-marker and carpet-mask helpers; require a decompression output bound and return only the consumed trace model.

#937 consumes the new detail object in its clean-history trait. Saved and live map state remain independent. No private map captures or account data are included.

## Dependencies and validation

#933 is merged. Current upstream `main` (`eac001c`, including #908) is integrated. This branch also includes the isolated conformance fix in #965, which addresses the independently reproduced upstream failure tracked in #964; please review/merge that small fix first.

- Full suite: **1349 passed, 53 existing xfailed**, **92 snapshots passed**.
- All pre-commit checks, including Ruff and mypy: passed.
- Wheel and sdist builds: passed.
- Regression coverage includes unknown/truncated markers, unsupported archive framing, malformed paths, stray-point filtering, and explicit historical-path rendering.
- No new physical-device run was performed for this revision.

Related: #767.

### DIFF
--- a/roborock/map/b01_q10_map_parser.py
+++ b/roborock/map/b01_q10_map_parser.py
@@ -1,8 +1,9 @@
 """Parser for Roborock Q10 (B01/ss07) map packets.
 
-Q10 devices deliver map data as a protocol-301 ``MAP_RESPONSE`` message after a
-``dpMultiMap`` list/get request. Unlike the Q7 ``SCMap`` protobuf
-format, the Q10 uses a custom, unencrypted binary packet:
+Q10 devices deliver map data as protocol-301 ``MAP_RESPONSE`` pushes. Current
+maps follow a read-only status request, while saved-map and clean-record detail
+packets follow their respective ``select`` requests. Unlike the Q7 ``SCMap``
+protobuf format, the Q10 uses a custom, unencrypted binary packet:
 
 - ``01 01`` marker, then a ``u32be`` map id (bytes 2-5) and two consecutive
   ``u16be`` dimensions: grid width (bytes 7-8) and grid height (bytes 9-10).
@@ -22,6 +23,7 @@ https://github.com/v1b3c0d3x3r/roborock-qseries-map-bridge
 import io
 import math
 import statistics
+import struct
 from dataclasses import dataclass, field, replace
 
 from PIL import Image
@@ -29,6 +31,7 @@ from vacuum_map_parser_base.config.color import ColorsPalette, SupportedColor
 from vacuum_map_parser_base.config.image_config import ImageConfig
 from vacuum_map_parser_base.map_data import ImageData, MapData, Point
 
+from roborock.data.code_mappings import RoborockEnum
 from roborock.data.containers import RoborockBase
 from roborock.exceptions import RoborockException
 
@@ -66,9 +69,6 @@ def classify_q10_cell(value: int) -> str:
     return LAYER_FLOOR
 
 
-MAP_PACKET_MARKER = b"\x01\x01"
-TRACE_PACKET_MARKER = b"\x02\x01"
-
 _MAP_ID_OFFSET = 2
 # Width and height are two consecutive u16be fields. An earlier revision read the
 # width as u16le at offset 8; that high byte is actually the height's high byte,
@@ -83,6 +83,7 @@ _LAYOUT_COMPRESSED_OFFSET = 29
 _ROOM_RECORD_LENGTH = 47
 _ROOM_NAME_LENGTH_OFFSET = 26
 _MAX_ROOMS = 32
+_MAX_GRID_CELLS = 16_000_000
 # Sanity bound for the erase-zone vector section's vertices-per-polygon field.
 _MAX_ERASE_ZONE_VERTICES = 16
 
@@ -194,10 +195,30 @@ class Q10HeaderCalibration:
         )
 
 
+class Q10MapPacketKind(RoborockEnum):
+    """Semantic kind identified by a Q10 map packet's two-byte marker."""
+
+    CURRENT = 1
+    TRACE = 2
+    CLEAN_RECORD_DETAIL = 3
+    SAVED_MAP_DETAIL = 4
+
+    @property
+    def marker(self) -> bytes:
+        """Return the two-byte wire marker for this packet kind."""
+        return bytes((self.value, 1))
+
+    @classmethod
+    def from_payload(cls, payload: bytes) -> "Q10MapPacketKind | None":
+        """Return the recognized kind for a payload marker."""
+        return next((kind for kind in cls if payload[:2] == kind.marker), None)
+
+
 @dataclass
 class Q10MapPacket:
-    """Decoded contents of a Q10 ``01 01`` map packet."""
+    """Decoded contents of a Q10 current or archived map packet."""
 
+    kind: Q10MapPacketKind
     map_id: int
     width: int
     height: int
@@ -211,6 +232,8 @@ class Q10MapPacket:
     """Carpet mask decoded from the packet tail: a full ``width*height`` grid in
     the same (top-down) pixel space as :attr:`grid`, where a non-zero cell is
     carpet (the value is the carpet kind). ``None`` if the packet carried none."""
+    historical_trace: "Q10HistoricalTracePacket | None" = None
+    """Cleaning path embedded in a clean-record detail packet, if present."""
 
     @property
     def layers(self) -> GridLayers:
@@ -266,6 +289,27 @@ class Q10TracePacket:
         return self.points[-1] if self.points else None
 
 
+@dataclass
+class Q10HistoricalTracePacket:
+    """Cleaning path embedded in a Q10 ``03 01`` clean-record detail packet.
+
+    This is a different wire layout from the live ``02 01`` trace. Its header
+    carries a 16-bit format version, a 32-bit opaque value, a 32-bit
+    point count, a signed heading, and a zero reserved word. Points use the same
+    signed big-endian ``(x, y)`` coordinate pairs as the live trace.
+    """
+
+    points: list[Q10Point] = field(default_factory=list)
+    version: int = 0
+    opaque_value: int = 0
+    heading: int = 0
+
+    @property
+    def robot_position(self) -> Q10Point | None:
+        """The final recorded position, if the historical path is non-empty."""
+        return self.points[-1] if self.points else None
+
+
 # Trace packet (``02 01``): a 14-byte header followed by big-endian int16 (x, y)
 # point pairs forming the accumulated session path. Header layout confirmed
 # against live ss07 captures and cross-checked by @andrewlyeats:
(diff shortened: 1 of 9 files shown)


roborock/map/b01_q10_map_parser.py
```diff
@@ -486,9 +564,14 @@ def parse_map_packet(payload: bytes) -> Q10MapPacket:
     rooms = _parse_rooms(room_data, grid)
     tail = payload[layout_end:]
     erase_zones = _parse_erase_zones(tail)
-    carpet_mask = _parse_carpet_mask(tail, width, height)
+    carpet_mask, carpet_end = _parse_carpet_block(tail, width, height)
+    if kind is Q10MapPacketKind.CLEAN_RECORD_DETAIL and carpet_end is not None:
```
allenporter, line 568: when this kind of map packet is received, how many of these other fields are also included?

What i'm wondering is if a historical trace kind has a lot of overlap with other fields in Q10MapPacket or if it needs to be a separate type.  basically as more fields are added the map packet seems like a sparse object. You could imagine each kind Q10MapPacketKind has a separate dataclass for example, if the overlap is low.

roborock/protocols/b01_q10_protocol.py
```diff
@@ -113,18 +112,18 @@ class Q10DpsUpdate:
...
     """
     if message.protocol == RoborockMessageProtocol.MAP_RESPONSE:
         payload = message.payload or b""
-        if is_map_packet(payload):
-            return parse_map_packet(payload)
-        if is_trace_packet(payload):
+        if Q10MapPacketKind.from_payload(payload) is Q10MapPacketKind.TRACE:
             return parse_trace_packet(payload)
+        if Q10MapPacketKind.from_payload(payload) is not None:
+            return parse_map_packet(payload)
```
allenporter, line 127: How about we get the kind once:
```suggestion
        kind = Q10MapPacketKind.from_payload(payload)
        if kind is Q10MapPacketKind.TRACE:
            return parse_trace_packet(payload)
        if kind is not None:
            return parse_map_packet(payload)
```

roborock/map/b01_q10_map_parser.py
```diff
@@ -348,11 +414,13 @@ def _drop_stray_leading_point(points: list[Q10Point]) -> list[Q10Point]:
     return points
 
 
-def lz4_block_decompress(data: bytes) -> bytes:
+def lz4_block_decompress(data: bytes, max_output_size: int | None = None) -> bytes:
```
allenporter, line 417: No need to make the second arg optional given we're updating all the callers in this change i believe.

roborock/map/b01_q10_render.py
```diff
@@ -84,7 +84,7 @@ class Q10MapOverlays:
 
 def render_q10_map(
```
allenporter, line 85: Whats the relationship between a Q10MapPacket and a trace? especially Q10HistoricalTracePacket. 

My impression is Q10HistoricalTracePacket is a field that was added to Q10MapPacket, so why would it also be passed here as `trace`?

roborock/map/b01_q10_map_parser.py
```diff
@@ -486,9 +564,14 @@ def parse_map_packet(payload: bytes) -> Q10MapPacket:
     rooms = _parse_rooms(room_data, grid)
     tail = payload[layout_end:]
     erase_zones = _parse_erase_zones(tail)
-    carpet_mask = _parse_carpet_mask(tail, width, height)
+    carpet_mask, carpet_end = _parse_carpet_block(tail, width, height)
+    if kind is Q10MapPacketKind.CLEAN_RECORD_DETAIL and carpet_end is not None:
+        historical_trace, _ = _parse_clean_record_trace(tail, carpet_end)
```
allenporter, line 569: we just ignore the second return value, so can it be removed?

roborock/map/b01_q10_map_parser.py
allenporter: re-reading the code I don't actually see the historical trace even being used anywhere yet, except in the tests.

Naively, it seems to me like we shouldn't be sticking this on to a map package and instead just using the kind to parse a clean record?

allenporter (changes requested):
I am very happy to keep momentum going on this.  I am adding additional feedback since this hasn't been moving since my last comment, in case it was missed. I have a lot of time to review during the us holidy weekend.

roborock/map/b01_q10_map_parser.py
```diff
@@ -194,10 +195,30 @@ def charger_pixels(self) -> tuple[float, float] | None:
...
+
+    @property
+    def marker(self) -> bytes:
+        """Return the two-byte wire marker for this packet kind."""
+        return bytes((self.value, 1))
+
+    @classmethod
+    def from_payload(cls, payload: bytes) -> "Q10MapPacketKind | None":
+        """Return the recognized kind for a payload marker."""
+        return next((kind for kind in cls if payload[:2] == kind.marker), None)
```
allenporter, line 214: Because this inherits from `RoborockEnum`, `RoborockEnum._missing_` expects an `unknown` member (e.g. `unknown = -1`). Without it, `_missing_` defaults to `next(item for item in cls)`, which silently resolves unknown future wire markers to `CURRENT = 1`.

Also, rather than iterating all members with `next((kind for kind in cls ...))` on every push, we can do an O(1) lookup:

```python
class Q10MapPacketKind(RoborockEnum):
    """Semantic kind identified by a Q10 map packet's two-byte marker."""

    unknown = -1
    CURRENT = 1
    TRACE = 2
    CLEAN_RECORD_DETAIL = 3
    SAVED_MAP_DETAIL = 4

    @property
    def marker(self) -> bytes:
        """Return the two-byte wire marker for this packet kind."""
        if self is self.unknown:
            return b""
        return bytes((self.value, 1))

    @classmethod
    def from_payload(cls, payload: bytes) -> "Q10MapPacketKind | None":
        """Return the recognized kind for a payload marker."""
        if len(payload) < 2 or payload[1] != 1:
            return None
        kind = cls(payload[0])
        return None if kind is cls.unknown else kind
```
*(Or if this enum is strictly an internal framing marker classifier, a standard `IntEnum` without `RoborockEnum` would avoid the fallback behavior altogether).*

roborock/map/b01_q10_map_parser.py
```diff
@@ -220,6 +241,18 @@ def layers(self) -> GridLayers:
         return decompose_grid(self.width, self.height, self.grid, rooms, classify_q10_cell, flip=False)
 
 
+@dataclass
+class Q10CleanRecordMapPacket(Q10MapPacket):
```
allenporter, line 245: Seeing `**vars(packet)` in `parse_map_packet` and `test_b01_q10_render.py`, along with `isinstance(packet, Q10CleanRecordMapPacket)` in `render_q10_map`, makes me wonder if subclassing `Q10MapPacket` is the right abstraction.

A clean-record response isn't really a specialized variant of a map raster; it's a compound payload containing both a map and a historical path. If we model it via composition instead:

```python
@dataclass
class Q10CleanRecordDetail:
    map: Q10MapPacket
    trace: Q10HistoricalTracePacket | None = None
```

Then:
1. `Q10MapPacket` remains pure and uniform across all map kinds (avoiding `**vars(packet)` or duplicating 9 constructor arguments).
2. `render_q10_map(packet, trace=None)` remains pure and consumer-agnostic—callers just do `render_q10_map(record.map, trace=record.trace)` with zero `isinstance` branching or ignored arguments.

allenporter (commented):
Thanks for iterating on this and splitting out the clean-record path ownership.

Taking a close look at the resulting `Q10CleanRecordMapPacket` boundary, the `**vars(packet)` unpacking in `parse_map_packet` and the `isinstance` check in `render_q10_map` feel like an awkward fit for dataclass inheritance. Rather than subclassing `Q10MapPacket`, modeling this as a composite object `Q10CleanRecordDetail(map=Q10MapPacket, trace=Q10HistoricalTracePacket | None)` would be significantly cleaner and avoid both constructor duplication and `isinstance` branches in rendering.

I've also left a few inline comments on `RoborockEnum` fallback resilience, stripping stray origin points, and trimming internal wire fields from public models.

roborock/map/b01_q10_map_parser.py
```diff
@@ -266,6 +299,27 @@ def robot_position(self) -> Q10Point | None:
...
+
+    This is a different wire layout from the live ``02 01`` trace. Its header
+    carries a 16-bit format version, a 32-bit opaque value, a 32-bit
+    point count, a signed heading, and a zero reserved word. Points use the same
+    signed big-endian ``(x, y)`` coordinate pairs as the live trace.
+    """
+
+    points: list[Q10Point] = field(default_factory=list)
+    version: int = 0
+    opaque_value: int = 0
```
allenporter, line 314: Do we need to expose `version` and `opaque_value` on the public dataclass?

`version` is already validated to equal 1 by the parser (returning `None` otherwise), and `opaque_value` is undecoded wire framing. Removing them keeps the public domain model clean and aligned with `Q10TracePacket`.

roborock/map/b01_q10_map_parser.py
```diff
@@ -301,12 +364,22 @@ def robot_position(self) -> Q10Point | None:
 
 def is_map_packet(payload: bytes) -> bool:
     """Return True if the payload is a Q10 full-map (``01 01``) packet."""
-    return payload[:2] == MAP_PACKET_MARKER
+    return Q10MapPacketKind.from_payload(payload) is Q10MapPacketKind.CURRENT
+
+
+def is_clean_record_map_packet(payload: bytes) -> bool:
```
allenporter, line 370: Can we prune `is_clean_record_map_packet` and `is_saved_map_packet`? Since `decode_message` now checks `Q10MapPacketKind.from_payload(payload)` directly, these single-expression helpers aren't used in production code.

Similarly, `_parse_carpet_mask` at line 711 appears to be an unused private compatibility helper that can be cleaned up.

roborock/map/b01_q10_map_parser.py
```diff
@@ -578,23 +685,83 @@ def _parse_carpet_mask(tail: bytes, width: int, height: int) -> bytes | None:
...
+    point_count = int.from_bytes(
+        tail[offset + _HISTORICAL_TRACE_POINT_COUNT_OFFSET : offset + _HISTORICAL_TRACE_POINT_COUNT_OFFSET + 4],
+        "big",
+    )
+    points_end = header_end + point_count * 4
+    if points_end > len(tail):
+        return None
+    coordinates = struct.iter_unpack(">hh", memoryview(tail)[header_end:points_end])
+    return Q10HistoricalTracePacket(
+        points=[Q10Point(x=x, y=y) for x, y in coordinates],
```
allenporter, line 753: Should we run `points = _drop_stray_leading_point(points)` here as well?

If clean-record paths share the same firmware sentinel near `(0, 0)` that live traces have, leaving it in will cause visual lines to the origin and skew `solve_q10_calibration`.

roborock/map/b01_q10_render.py
```diff
@@ -97,8 +98,10 @@ def render_q10_map(
     available trace and DPS overlays are projected and drawn in pixel space.
     Raises :class:`RoborockException` if map rendering fails.
     """
+    # An archived map owns its path; a live trace must never replace it.
+    render_trace = packet.historical_trace if isinstance(packet, Q10CleanRecordMapPacket) else trace
```
allenporter, line 102: This line illustrates the friction with subclassing: `render_q10_map` takes `trace: Q10TracePacket | None = None`, but if `packet` is a `Q10CleanRecordMapPacket`, the passed `trace` argument is silently ignored.

If we use composition (`Q10CleanRecordDetail(map=..., trace=...)`), `render_q10_map` doesn't need this branch at all—the caller simply passes `render_q10_map(record.map, trace=record.trace)`.

hCoureau:
@allenporter Implemented the requested composition in 4eb0b9b and updated the consumers in #937–#939:

- `Q10CleanRecordDetail(map, trace)` replaces map inheritance and `**vars(packet)` construction. Rendering uses the supplied map/path directly, without type-dependent branches or ignored arguments.
- `Q10MapPacketKind` now has an explicit `unknown` member and constant-time lookup; unknown firmware markers are ignored.
- Historical paths expose only points and heading; version validation stays internal and the opaque wire value is no longer public.
- Removed the unused archive-marker and carpet-mask helpers.
- Historical paths use the existing conservative stray-leading-point filter, with tests for the outlier, ordinary origin points, and short paths.

Merged current main and preserved the new position/zone/goto support. Full local validation on #936: 1349 passed, 53 existing xfailed, 92 snapshots passed; pre-commit and package builds passed.

Separately, current main's new model conformance test fails on the existing frozen `Q10RoborockPoint`. I reproduced that on untouched `eac001c`, opened #964, and submitted the isolated fix in #965. This branch includes that commit so validation can pass; #965 should land first.

Could you review the revised boundary and parser changes when you have time?

lboue:
@hCoureau @allenporter This PR is really valuable for Q10 map support, but it's been waiting on the Sep 12 architecture feedback.

Do you want help picking up the composition refactor? The changes are clear:
- `Q10CleanRecordDetail(map, trace)` composition model
- RoborockEnum + enum resilience fixes
- Cleanup (unused helpers, internal fields trimming)

@allenporter, should hCoureau tackle these all at once, or is there a staged approach you'd prefer?

Happy to pair on implementation or review if that unblocks things.

hCoureau:
Thanks — I’ve already implemented the full set of Sep 12 changes in 4eb0b9b, including the Q10CleanRecordDetail(map, trace) composition model, enum fallback handling, removal of unused helpers, internal-field cleanup, and the historical-trace filtering. I also updated the dependent consumers in #937–#939 and resolved the inline threads.
CI is green and the branch is ready for Allen’s follow-up review. If you have time to review the updated implementation too, that would be very helpful.

Issue #262, opened 2024-12-26
app_set_dryer_setting, intended behaviour


Nerih:
Hi when calling:-
command: app_set_dryer_setting
      params:
        status: 1
        on:
          dry_time: 60  # Hardcoded 60 seconds for testing

correctly sets the mop dry time, however does not turn on the drying. 

Is this intended, as there seems to be no way to programmatically start the drying process?

Best,
Nerih

Nerih:
Hi, solved with app_set_dryer_status with param status:1

Lash-L:
Hi sorry i missed this, glad you figured it out!

PR #952, opened 2026-09-11
feat: add MopDryerTrait for controlling the dock mop dryer

### LINKED ISSUE #905: Add a trait for controlling the dock mop dryer
## Summary

There's currently no way to start or stop a mop drying cycle through the trait API. `dry_status` is readable, it was annotated with dps metadata in #740 and is exposed to Home Assistant via `is_field_supported(StatusV2, StatusField.DRY_STATUS)`. But the write side has no trait behind it, so consumers have to fall back to `RoborockCommand.APP_SET_DRYER_STATUS`.

I want to contribute a function to Home Assistant to switch the dryer on and off , but this is currently only possible via this workaround that I do not think fits into the design. Other switches in the Home Assistant integration are now entirely trait-driven (`child_lock`, `flow_led_status`, `dnd`, `valley_electricity_timer`). 

## Prior Art

#262 asked for exactly this in Dec 2024:

> correctly sets the mop dry time, however does not turn on the drying. Is this intended, as there seems to be no way to programmatically start the drying process?

It was closed by the reporter with:

> Hi, solved with `app_set_dryer_status` with param `status:1`

So the command is confirmed working by another user, but the finding only ever lived in a closed issue and it was never carried into the library. `APP_SET_DRYER_SETTING` configures the dry duration; `APP_SET_DRYER_STATUS` starts and stops the cycle. I've since confirmed the latter against my own dock.

That issue predates the trait system entirely, so as far as I can tell this was never considered and rejected.

## Proposed fix

A switch trait in the shape of the existing ones (`ChildLockTrait`, `ValleyElectricityTimerTrait`), gated on the dock being able to dry:

```python
class MopDryerTrait(common.V1TraitMixin, common.RoborockSwitchBase):
    """Trait for controlling the dock mop dryer."""

    requires_dock_features = lambda f: f.is_dryable

    def __init__(self, status_trait: StatusTrait) -> None:
        super().__init__()
        self._status_trait = status_trait

        await self._status_trait.refresh()

    @property
    def is_on(self) -> bool:
        return bool(self._status_trait.dry_status)

    async def enable(self) -> None:
        await self.rpc_channel.send_command(
            RoborockCommand.APP_SET_DRYER_STATUS, params={"status": 1}
        )
        self._status_trait.dry_status = 1  # optimistic, as in the other switch traits

    async def disable(self) -> None:
        await self.rpc_channel.send_command(
            RoborockCommand.APP_SET_DRYER_STATUS, params={"status": 0}
        )
        self._status_trait.dry_status = 0
```

Because it takes a constructor argument it would be created in the special-case block in `discover_features()`, alongside `WashTowelModeTrait` and `ObstaclePhotoTrait`.

This follows the same pattern as the action methods being added to V1 traits elsewhere (#856, and the q7 series in #841–#844).

## TBC before opening a PR

**1. Where should the state come from?**

Unlike every other switch trait, this one has no query command of its own. If a cycle is running is reported as `dry_status` on `StatusV2`. The sketch above therefore holds the status trait and refreshes through it, which keeps `V1TraitMixin`'s `command`/`converter` unused.

The alternative is a self-contained trait backed by `APP_GET_DRYER_SETTING` with a new `DryerSetting` dataclass. I didn't go that way because I have no sample of that response and didn't want to guess at the format, But I'm happy to capture it from my dock if you'd prefer that shape, it would also make the dry-duration setting available.

**2. Which capability should be used to check?**

The code uses `dock_features.is_dryable`. The alternatives are `is_supported_drying` (device feature bit) or `is_field_supported(StatusV2, StatusField.DRY_STATUS)`, which is what the binary sensor on the HA side already uses.

dschuld:
## Summary
A switch trait for starting and stopping a mop drying cycle on docks that support
drying, via `APP_SET_DRYER_STATUS`.

`dry_status` is already readable. It was annotated with dps metadata in #740 and is
consumed by Home Assistant through `is_field_supported(StatusV2, StatusField.DRY_STATUS)`. But there is no trait behind the write side, so callers have to fall back to a raw command.

```python
mop_dryer = device.v1_properties.mop_dryer
if mop_dryer is not None:
    await mop_dryer.enable()   # start drying
    mop_dryer.is_on            # True
    await mop_dryer.disable()  # stop
```

The trait is created in the special-case block in `discover_features()` alongside
`WashTowelModeTrait` and `ObstaclePhotoTrait`, since it takes a constructor argument.
It is gated on `dock_features.is_dryable`.

## Background

#262 asked for this in Dec 2024 and was closed by the reporter with:

> solved with `app_set_dryer_status` with param `status:1`

`APP_SET_DRYER_SETTING` configures the dry *duration*; `APP_SET_DRYER_STATUS` starts and
stops the *cycle*. That finding never made it into the library. I've confirmed the
command works against my own dock.

## Open Issues / Design Choices

**1. Where the state comes from.** Unlike the other switch traits, the dryer has no query
command of its own. If a cycle is running is reported as `dry_status` on `StatusV2`.
So this trait holds the status trait, reads `is_on` from it, and overrides `refresh()` to
delegate, leaving `V1TraitMixin`'s `command`/`converter` unused.

That is a deliberate deviation from the pattern documented in the module docstring
("subclass `V1TraitMixin` and a `RoborockBase` dataclass… you must define a `command`
class variable"). I would like to point this out here rather than have it found in review.
The alternative is a self-contained trait backed by `APP_GET_DRYER_SETTING` with a new
`DryerSetting` dataclass, which would also expose the dry-duration setting. I didn't go
that way because I have no sample of that response and didn't want to guess at the wire
format. I am happy to capture it from my dock and rework this if this approach is preferred.

## Testing

50 new tests in `tests/devices/traits/v1/test_mop_dryer.py`:

- trait is created for all dock types where `is_dryable` is true, and absent for all where
  it is false (parametrized over the full `RoborockDockTypeCode` enum)
- `is_on` mapping for `dry_status` of `None` / `0` / `1`
- `enable()` / `disable()` send `APP_SET_DRYER_STATUS` with the right params and apply the
  optimistic update
- `refresh()` delegates to the status trait

100% line and branch coverage on the new module.

## AI assistance disclosure
This contribution was prepared with Claude Code assistance. I reviewed the submitted changes and test results and take responsibility for the contribution.

Closes #905


roborock/devices/traits/v1/__init__.py
```diff
@@ -285,6 +289,11 @@ async def discover_features(self) -> None:
             obstacle_photos._rpc_channel = self._get_rpc_channel(obstacle_photos)
             self.obstacle_photos = obstacle_photos
 
+        if self.mop_dryer is None and self._is_supported(MopDryerTrait, "mop_dryer", dock_features):
```
allenporter, line 292: Because `MopDryerTrait` takes constructor arguments (`status_trait: StatusTrait`), please also add it to the exclusion filter in the dynamic reflection loop below (around line 299):

```python
if item_type in (ObstaclePhotoTrait, MopDryerTrait):
    continue
```

Otherwise, if `_is_supported` evaluates to `True` when `self.mop_dryer` is `None`, the loop attempts `item_type()` without arguments and raises a `TypeError`.

roborock/devices/traits/v1/mop_dryer.py
```diff
@@ -0,0 +1,48 @@
...
+from roborock.roborock_typing import RoborockCommand
+
+_STATUS_PARAM = "status"
+
+
+def _supports_mop_dryer(dock_features: RoborockDockFeatures) -> bool:
+    return dock_features.is_dryable
+
+
+class MopDryerTrait(common.V1TraitMixin, common.RoborockSwitchBase):
```
allenporter, line 15: Subclassing `V1TraitMixin` without declaring the class variables `command: ClassVar[RoborockCommand]` and `converter: V1TraitDataConverter` (and without `MopDryerTrait` subclassing a `RoborockBase` dataclass) leaves the mixin contract incomplete. `V1TraitMixin.refresh()` expects to run `merge_trait_values(self, ...)` on dataclass fields.

As you noted in the PR description, the V1 protocol actually supports `APP_GET_DRYER_SETTING` and `APP_SET_DRYER_SETTING` (documented in `docs/V1_API_COMMANDS.md#app_get_dryer_setting`).

Two possible paths here:
1. **Option A (Preferred / Complete Trait)**: Model a self-contained `MopDryerTrait` backed by `APP_GET_DRYER_SETTING` with a `DryerSetting(RoborockBase)` dataclass. This exposes both the dryer switch (`enable`/`disable`) and duration settings while completely avoiding coupling with `StatusTrait`.
2. **Option B (Lightweight Switch)**: If keeping this strictly to a start/stop cycle switch without settings, avoid inheriting from `V1TraitMixin` since its query/converter contract isn't used. Instead, accept the RPC channel explicitly.

roborock/devices/traits/v1/mop_dryer.py
```diff
@@ -0,0 +1,48 @@
...
+    @property
+    def is_on(self) -> bool:
+        """Return whether a drying cycle is currently running."""
+        return bool(self._status_trait.dry_status)
+
+    async def enable(self) -> None:
+        """Start drying the mop."""
+        await self.rpc_channel.send_command(RoborockCommand.APP_SET_DRYER_STATUS, params={_STATUS_PARAM: 1})
+        # Optimistic update to avoid an extra refresh
+        self._status_trait.dry_status = 1
```
allenporter, line 42: Mutating `_status_trait.dry_status` directly across trait boundaries leaks encapsulation and bypasses any listeners/subscribers registered on `StatusTrait` (like `TraitUpdateListener`).

Per repository guidelines, traits should manage their own internal state (e.g., maintaining an internal `_is_on` attribute on `MopDryerTrait` initialized or refreshed as needed) rather than mutating an external trait's cached telemetry.

tests/devices/traits/v1/test_mop_dryer.py
```diff
@@ -0,0 +1,128 @@
...
+    assert device.v1_properties
+
+    await getattr(mop_dryer, method_name)()
+
+    mock_rpc_channel.send_command.assert_called_with(
+        RoborockCommand.APP_SET_DRYER_STATUS, params={"status": expected_status}
+    )
+    # The command result is applied optimistically to avoid an extra refresh
+    assert device.v1_properties.status.dry_status == expected_status
+    assert mop_dryer.is_on is bool(expected_status)
```
allenporter, line 103: Tests should verify the public contract and behavior of the trait under test (`mop_dryer.is_on`, emitted RPC commands) rather than asserting on cross-trait side-effects in `device.v1_properties.status.dry_status`.

allenporter (commented):
Thank you for putting this together @dschuld! Exposing mop drying controls via a trait is a very welcome addition and resolves a longstanding request (#905, #262). (Note: I am currently experimenting with agentic code review to help review PRs, so please let me know if anything in the feedback looks off).

I have left a few inline comments on the architecture, trait contract, and state encapsulation.

roborock/devices/traits/v1/mop_dryer.py
dschuld: Thanks for the detailed feedback and directions! I went with option A and implemented a proper trait. I captured `APP_GET_DRYER_SETTING` from my dock. There are now MopDryerSetting and MopDryerProfile dataclasses in v1_containers.py, and the trait declares command and converter and inherits the generic refresh().

roborock/devices/traits/v1/mop_dryer.py
dschuld: Fixed. The trait is now a `MopDryerSetting` dataclass and updates its own `self.status`, the same way `ChildLockTrait` does with `lock_status`. It no longer holds a reference to `StatusTrait` at all, so nothing bypasses the update listeners.

tests/devices/traits/v1/test_mop_dryer.py
dschuld: Fixed. The tests now assert only on commands emitted and the trait's own state, no references to `device.v1_properties.status`. The refresh test uses the `APP_GET_DRYER_SETTING` payload captured from my dock and checks the nested `on`/`off` profiles parse correctly.

roborock/devices/traits/v1/__init__.py
dschuld: After the refactoring, the constructor does not take arguments anymore, so this is obsolete.

allenporter (approved):

Issue #930, opened 2026-08-28
Saros 10R: set_clean_motor_mode generates invalid Mop payload (-10005 Param error)


BlackRockCity:
### Description

`python-roborock` appears to generate an incorrect `set_clean_motor_mode` payload when selecting Mop mode on a Roborock Saros 10R.

The generated command is rejected by the robot with:

```text
code: -10005
message: Param error
```

I captured the command generated through the Home Assistant Roborock integration, compared it with the command sent by the official Roborock iOS app, and then independently tested the official-app parameters through `vacuum.send_command`.

The official-app parameters work successfully.

### Environment

- Robot: Roborock Saros 10R
- Model: `roborock.vacuum.a144`
- Firmware: `02.50.56`
- Home Assistant: `2026.8.3`
- Integration: Home Assistant Roborock integration / `python-roborock`

### Reproduction

Using the Home Assistant Roborock integration, the cleaning mode entity exposes:

```text
vacuum
vac_and_mop
mop
```

Selecting `mop` results in `SET_CLEAN_MOTOR_MODE` being sent with:

```python
[
    {
        "fan_power": 109,
        "water_box_mode": 202,
        "mop_mode": 300
    }
]
```

The Saros 10R rejects this payload:

```text
{'code': -10005, 'message': 'Param error'}
```

This occurs both through the local connection and MQTT.

### Official Roborock app behavior

I captured the RPC sent by the official Roborock iOS app when selecting Mop on the same robot.

The official app sends:

```json
{
  "method": "set_clean_motor_mode",
  "params": [
    {
      "fan_power": 105,
      "mop_mode": 300,
      "water_box_mode": 225
    }
  ]
}
```

The robot accepts this command.

### Independent verification

I then sent the official-app parameters directly through Home Assistant:

```yaml
action: vacuum.send_command
target:
  entity_id: vacuum.saros_10r
data:
  command: set_clean_motor_mode
  params:
    - fan_power: 105
      mop_mode: 300
      water_box_mode: 225
```

This succeeds.

The robot changes to Mop mode, and the official Roborock iOS app subsequently reflects Mop as the active cleaning mode.

This appears to rule out the robot, MQTT/local transport, and Home Assistant service call itself as the cause. The Saros 10R accepts `set_clean_motor_mode`; the issue appears to be the model-specific values being generated for that command.

### Additional mode captures

For comparison, I captured the following commands from the official Roborock app on this Saros 10R:

[...]

Issue #801, opened 2026-03-26
Missing Water Flow settings on Saros 10R


thibaultmol:
I have a saros 10R (roborock.vacuum.a144) and normally put my bot on the 'gentle' water flow setting but it seems like this library goes from from low to medium but doesn't have the gentle option.

I'm not familiar with mappings on other bots, but on the Saros 10R it seems:
1-3: Slight
4-7: Low
8-12: Gentle
13-17: Medium
18-22: Moderate
23-27: High
28-30: Extreme

BlackRockCity:
I have some additional testing from a Saros 10R (`roborock.vacuum.a144`) that may help diagnose this.

### Environment

- Roborock Saros 10R
- Model: `roborock.vacuum.a144`
- Firmware: `02.50.56`
- Home Assistant: `2026.8.3`
- Official Roborock iOS app used to verify the resulting water-flow slider position

I can confirm the ranges originally reported in this issue:

```text
1-3    Slight
4-7    Low
8-12   Gentle
13-17  Medium
18-22  Moderate
23-27  High
28-30  Extreme
```

I also captured commands generated by the official Roborock app.

The app uses:

```text
set_water_box_custom_mode
```

The parameter appears to correspond directly to the 1-30 slider position with:

```text
protocol_value = 220 + slider_position
```

Examples captured from the official app:

```text
Position 1  (Slight):  params [221]
Position 5  (Low):     params [225]
Position 16 (Medium):  params [236]
Position 25 (High):    params [245]
Position 30 (Extreme): params [250]
```

The robot also reports DPS 124, but this does not appear to contain the actual 1-30 slider position. For example, I observed:

```text
Low / position 5 / command 225       -> DPS 124 = 201
Medium / command 236                 -> DPS 124 = 202
High / position 25 / command 245     -> DPS 124 = 203
Extreme / position 30 / command 250  -> DPS 124 = 208
```

I then tested the native Home Assistant `select` entity:

```text
select.saros_10r_mop_intensity
```

using `select.select_option`.

The resulting settings shown by the official Roborock iOS app were:

```text
HA option "low"      -> Low, slider position 5
HA option "moderate" -> High, slider position 25
HA option "high"     -> Extreme, slider position 28
```

So this appears to be more than the missing `gentle` option. The existing discrete enum mapping is also selecting incorrect physical water-flow settings on the Saros 10R.

In particular:

```text
"moderate" -> actual High
"high"     -> actual Extreme
```

[...]

PR #961, opened 2026-09-16
fix: preserve water-slide codes in cleaning-mode commands and status

### LINKED ISSUE #931: WaterModes pure-water members collapse into StrEnum aliases, sending water code 202 to water-slide devices (-10005 Param error)
## Summary

Four `WaterModes` members share a value string with an earlier member. `StrEnum` de-duplicates by value, so they silently become aliases and carry the earlier member's `code`.

The visible consequence: `_get_default_mopping_water_mode()` returns `PURE_WATER_FLOW_MIDDLE` specifically to avoid code 202 (per its own comment, water-slide devices "don't accept" it), but the alias resolves to `MEDIUM`, so `set_cleaning_mode()` sends 202 to exactly those devices and they reject the command.

## Version

`python-roborock==7.1.1` (current release, and the version Home Assistant pins). Reproduced on a clean install, CPython 3.13.

## Reproduction

```python
from roborock.data.v1.v1_clean_modes import WaterModes, WATER_SLIDE_MODE_MAPPING

print(len(list(WaterModes)))                                   # 15, though 19 names are declared
print(WaterModes.PURE_WATER_FLOW_MIDDLE is WaterModes.MEDIUM)  # True
print(WaterModes.PURE_WATER_FLOW_MIDDLE.code)                  # 202, expected 235
print(WaterModes.from_code_optional(235))                      # None

for code, mode in WATER_SLIDE_MODE_MAPPING.items():
    if code != mode.code:
        print(code, "->", mode.name, mode.code)
# 225 -> LOW 201
# 235 -> MEDIUM 202
# 248 -> HIGH 203
# 250 -> EXTREME 208
```

The four collapsing pairs are `("low", 225)` into `LOW`, `("medium", 235)` into `MEDIUM`, `("high", 248)` into `HIGH`, and `("extreme", 250)` into `EXTREME`.

## Impact

On a device with `is_water_slide_mode_supported`, `get_cleaning_mode_parameters()` emits `{'fan_power': 105, 'water_box_mode': 202, 'mop_mode': 300}` for `CleaningMode.MOP` (and the same 202 for `VAC_AND_MOP`). Observed on a Saros 20 Sonic (`roborock.vacuum.a288`) through the Home Assistant integration:

```
File ".../roborock/devices/traits/v1/status.py", line 131, in set_cleaning_mode
    await self.rpc_channel.send_command(
roborock.exceptions.RoborockException: {'code': -10005, 'message': 'Param error'}
```

Sending the same command with 235 instead of 202 succeeds on that device.

Secondary: `from_code_optional()` returns `None` for 225, 235, 248 and 250, so those device codes cannot be decoded through the enum.

The display path is unaffected, because `get_water_mode_mapping()` keys off the dict rather than the member. This only surfaces on writes.

## Suggested fix

Unique values plus `display_name`, matching the existing `DEEP_PLUS_CN = ("deep_plus_cn", 305, "deep_plus")` precedent:

```diff
-    PURE_WATER_FLOW_SMALL = ("low", 225)
-    PURE_WATER_FLOW_MIDDLE = ("medium", 235)
-    PURE_WATER_SUPER_BEGIN = ("high", 248)
-    PURE_WATER_FLOW_END = ("extreme", 250)
+    PURE_WATER_FLOW_SMALL = ("pure_water_flow_small", 225, "low")
+    PURE_WATER_FLOW_MIDDLE = ("pure_water_flow_middle", 235, "medium")
+    PURE_WATER_SUPER_BEGIN = ("pure_water_super_begin", 248, "high")
+    PURE_WATER_FLOW_END = ("pure_water_flow_end", 250, "extreme")
```

and in `get_water_mode_mapping()`, read `mode.display_name` instead of `mode.value` in both branches.

Verified with that patch applied to 7.1.1: 19 members, all four codes correct, `from_code_optional()` resolves every slide code, `WATER_SLIDE_MODE_MAPPING` round-trips, and `get_water_mode_mapping()` still returns `{200: 'off', 221: 'slight', 225: 'low', 235: 'medium', 245: 'moderate', 248: 'high', 250: 'extreme'}`, so the user-facing strings do not change.

One follow-up outside this repo: Home Assistant's `select.py` builds the water options from `mode.value` and would need `mode.display_name`, as its mop-route select already does.

Happy to open a PR.

xiaoyaner0201:
## Summary

Refs #931 and #930. Thanks to @liborplisek for identifying the enum-alias root cause and @BlackRockCity for the command/status captures.

Bring the combined cleaning-mode paths into line with the existing feature-aware water-mode mapping:

- Encode water-slide cleaning-mode commands using the device mapping's wire code, so `vac_and_mop` and `mop` emit `235` instead of the aliased member's legacy `202`.
- Resolve reported slide water codes through `WATER_SLIDE_MODE_MAPPING` when classifying cleaning mode. Keep the existing legacy-code fallback and direct enum-input behavior.
- Correct existing tests that compared against `PURE_WATER_FLOW_MIDDLE.code` on both sides: those tests were passing with the same incorrect `202`. Expected wire values now use explicit protocol integers.
- Add public `StatusTrait` read/write regressions for all currently mapped slide intensities, vacuum/mop selection, optional mop-route fields, legacy codes, enum inputs, and incomplete/unknown status.

## Compatibility and scope

This deliberately keeps the public enum values and user-facing option strings unchanged, so existing consumers using `mode.value` do not need a coordinated migration. It fixes the demonstrated combined-mode command/classification paths, **not the standalone enum-alias representation**: `WaterModes.PURE_WATER_FLOW_MIDDLE.code` and generic `WaterModes.from_code_optional(235)` are still affected. A broader unique-value/display-name migration remains separate, so this PR references rather than closes #931.

The water-flow range/label changes discussed in #801, other protocol families, Home Assistant automation policies, and login fixes are outside this PR.

## Reproduction and hardware evidence

On a G30S Ultra with Home Assistant 2026.9.1 and python-roborock 7.2.3, a real status payload contained `fan_power=102, water_box_mode=235, mop_mode=303`. Individual settings decoded as balanced/medium/deep_plus, while the combined mode was `None` (HA `unknown`). An automation waiting for confirmed `vac_and_mop` stopped before sending its start command.

After applying the same feature-mapping read/write fix locally, the combined mode reported `vac_and_mop`, the outgoing combined-mode payload used `235`, and the automation passed its setting check and started real cleaning. Direct device readback showed cleaning active, no error, and the expected motor settings; cleaning counters increased. The upstream change is tested independently against current main, not just the deployed library version.

## Validation

Base: `eac001c9d0a5eb45cd20e49033c37ff00054617f`.

- Regression RED on unchanged production behavior: **19 assertion failures / 79 tests**, no collection errors. Failures include `202 != 235` and mode `None` versus `mop`/`vac_and_mop`.
- GREEN: `uv run pytest tests/devices/traits/v1/test_status.py -q` — **79 passed**.
- `uv run pre-commit run --all-files` — passed, including Ruff, Mypy and Codespell.
- `uv build` — source distribution and wheel built successfully.
- Full candidate suite: **1355 passed, 53 xfailed, 1 failed**; **92 snapshots passed**.
- The full suite also fails on the exact clean base with identical dependencies: **1317 passed, 53 xfailed, 1 failed**. The failing test name and assertion message match exactly: `tests/conformance/test_model_conformance.py::test_data_model_subclasses_roborock_base[roborock.data.b01_q10.b01_q10_containers.Q10RoborockPoint]`. This pre-existing Q10 conformance failure is not changed, suppressed, or marked xfail by this PR.

No device/account credentials, raw diagnostic dumps, or deployment scripts are included.


Lash-L:
Hi thanks for the fix! Can we take it a step further? Right now, this will still fail for say 223. Can we resolve the value via the range? When the user selects a value - we can set it to an exact value. But if the user changes it in the app to one of the inbetween slider values, we could still display the closest category.

e.g.

Add a small feature-gated range resolver

Preserve the raw value 223.

Resolve its range (221–223) to "slight".

PR #965, opened 2026-09-17
fix: preserve immutable Q10 points in model conformance checks

### LINKED ISSUE #964: Model conformance test fails on existing frozen Q10RoborockPoint
The model conformance test added in #955 fails on current `main` (`eac001c`) for the existing `Q10RoborockPoint` introduced by #908.

Reproduction from an untouched checkout:

```sh
uv run pytest -q tests/conformance/test_model_conformance.py
```

Result: **1 failed, 83 passed**. The failure is `test_data_model_subclasses_roborock_base[roborock.data.b01_q10.b01_q10_containers.Q10RoborockPoint]`.

`Q10RoborockPoint` is intentionally `@dataclass(frozen=True)` and supports hashable coordinate values. `RoborockBase` is a non-frozen dataclass; Python rejects inheriting a frozen dataclass from it. Removing `frozen=True` to satisfy the check would change the point's existing immutability and hashability contracts.

Proposed fix: document and check an explicit exception for this existing coordinate value, retain the inheritance requirement for every other domain model, and add tests that coordinate assignment/deletion remain prohibited and hashing is stable. This is independent of the pending Q10 archive PRs.

hCoureau:
The model conformance check fails on `main` for the existing frozen `Q10RoborockPoint`. A frozen dataclass cannot inherit from the non-frozen `RoborockBase`, so enforcing that inheritance would require changing the coordinate value's public behavior.

Document and check an explicit exception for this one existing type. All other domain models retain the inheritance requirement. Add regression tests for rejected coordinate assignment/deletion, equality, and stable hashing; production code is unchanged.

Fixes #964.

Validation: reproduced the failure on untouched `eac001c` (1 failed, 83 passed); with this change, the full suite reports **1320 passed, 53 existing xfailed**, with **92 snapshots passed**. All pre-commit checks and wheel/sdist builds pass.


allenporter (approved):
Thank you for the fix.

PR #966, opened 2026-09-20
fix: offload map parsing to worker thread to prevent event loop blocking

allenporter:
## Description
Offload CPU-intensive map parsing and rendering to worker threads via `asyncio.to_thread` to prevent blocking the asyncio event loop.

When devices are actively cleaning and returning map updates, parsing the raw map payload and rendering the map into PNG format runs synchronously on the asyncio event loop. In multi-device setups or frequent update cycles, this locks the event loop and leads to high latency spikes in downstream consumers like Home Assistant (see home-assistant/core#182656).

### Changes
- In `MapContentTrait.refresh()`: offload `self.converter.convert(response)` to a worker thread using `await asyncio.to_thread(self.converter.convert, response)`.
- In `HomeTrait`:
  - Factor cached map parsing into a helper function `_parse_cached_home_map_content(...)` and run it via `await asyncio.to_thread(...)` once, rather than running `asyncio.to_thread` per item in a loop.
  - Avoid redundant reparsing in `_parse_cached_home_map_content` when the cached bytes match the previously parsed `raw_api_response`.

Realted to https://github.com/home-assistant/core/issues/182656


roborock/devices/traits/v1/map_content.py
```diff
@@ -90,3 +92,17 @@ def __init__(self, map_parser_config: MapParserConfig | None = None) -> None:
         """Initialize MapContentTrait."""
         super().__init__()
         self.converter = MapContentConverter(MapParser(map_parser_config or MapParserConfig()))
+
+    async def refresh(self) -> None:
+        """Refresh the contents of this trait without blocking the event loop."""
+        response = await self.rpc_channel.send_command(self.command)
+        try:
```
allenporter, line 117: can we also make this avoid re-parsing if the content is unchanged?

allenporter: Updated! Added a check in `MapContentTrait.refresh()` to skip conversion and re-parsing if `self.raw_api_response` matches `response`, along with a test verifying repeated refreshes with unchanged data do not re-parse.

roborock/devices/traits/v1/home.py
```diff
@@ -95,10 +114,12 @@ async def discover_home(self) -> None:
             self._home_map_info = device_cache_data.home_map_info
             self._discovery_completed = True
             try:
-                self._home_map_content = {
-                    k: self._map_content.converter.parse_map_content(base64.b64decode(v))
-                    for k, v in (device_cache_data.home_map_content_base64 or {}).items()
-                }
+                self._home_map_content = await asyncio.to_thread(
+                    _parse_cached_home_map_content,
+                    self._map_content.converter,
```
allenporter: i think i may want to make an async version on self._map_content.converter that is also shared by the refresh

allenporter: Updated! Added `async_convert` and `async_parse_map_content` to `MapContentConverter`, used directly by both `MapContentTrait.refresh()` and `HomeTrait.discover_home()`.

Lash-L (approved):

Lash-L:
We also could probably get significant improvements in parsing time - may be something we want to look into in the future! Good change nonetheless!

